PR #5, opened 2015-08-17
Home page template

noahmanger:
Just wanted to get this out of the way. It needs the actual hero image eventually, but now it actually looks like a thing!


fec/fec/templates/base.html
```diff
@@ -13,9 +13,12 @@
...
 
     {# Global stylesheets #}
-    {% compress css %}
+<!--     {% compress css %}
     <link rel="stylesheet" type="text/x-scss" href="{% static 'scss/fec.scss' %}">
-    {% endcompress %}
+    {% endcompress %} -->
+
+     <!-- Uncomment while developing to use css served from fec-style -->
+    <link rel="stylesheet" href="http://localhost:8080/css/styles.css" />
```
jmcarp: What do you think about keying this on an environment variable or adding to `settings/local.py`? For example:
```
FEC_CMS_STYLE=http://localhost:8080/css/styles.css ./fec/manage.py runserver
```
or
```python
# local.py
FEC_CMS_STYLE = 'http://localhost:8080/css/styles.css'
```
(settings/local.py is a configuration file that isn't checked into git, and is used, appropriately enough, for settings specific to a specific local environment)

noahmanger: Oops, yep. I was going to ask about something like this. I like that idea.

noahmanger: We should also do this in the web-app too.

noahmanger:
I removed the link to the local stylesheet for now. Wasn't sure how to set up the local variable.

PR #13, opened 2015-08-20
Bundle scripts from fec-style.

jmcarp:
* Use glossary and accordion widgets from fec-cms
* Bundle assets with gulp and browserify
* Compress bundled scripts with django-compressor


bin/run.sh
```diff
@@ -1,4 +1,5 @@
 cd fec
+gulp build-js
```
noahmanger: The gulpfile is in /fec so I had to run this from that directoyr.

PR #40, opened 2015-09-09
Content updates

noahmanger:
- Various content and layout tweaks. 
- Updates the menu
- Updates the various "learn more" links on the checklist
- Adds images for the campaign guides

### DIFF
--- a/fec/fec/templates/base.html
+++ b/fec/fec/templates/base.html
@@ -52,7 +52,7 @@
             <input class="nav-toggle__input" id="dropdown-toggle-1" type="checkbox">
             <label for="dropdown-toggle-1" class="site-nav__link nav-toggle__label">Campaign finance data</label>
             <ul class="site-nav__dropdown">
-              <li class="site-nav__item">
+              <li class="site-nav__item u-under-lg-only">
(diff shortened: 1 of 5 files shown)


jenniferthibault:
Nice! Everything in here looks spot on to me, but I noticed that if you want to you do a few things to kill two :bird:s with one pull request. Based on this issue https://github.com/18F/fec-alpha/issues/67 , you could:
 - Remove the » symbols from the nav
 - Make "Browse" a category in the nav, and use only "Candidates", "Committees", "Filings", "Receipts", and "Disbursements" as line items. Consider using the tables icon once paired with "Browse" to see if you like it
 - Use "and" instead of "&" in the "Registration and reporting" navigation text
 
Last (and this isn't in an issue, nor is it crucial), to address spacing with the check list, we could use the regular-size check box instead of the larger one. This may help readability of the list items as well.

noahmanger:
Great! Thanks. I updated the nav (and also in the web app) and also made the checkboxes smaller in https://github.com/18F/fec-style/pull/76

noahmanger:
This should be ready to merge and deploy. Also fixes the broken link to the checklist.

Issue #26, opened 2015-08-31
Figure out how to do HTML tables in CMS


noahmanger:
Some of the registration content pages need simple HTML tables and we need to figure out how to do this.

noahmanger:
@jmcarp any thoughts on this? This would be a really good thing to get going for launch as it really helps the formatting of certain pages.

jmcarp:
The simplest thing here would be to add raw HTML as a content type in the streamfield sections. WYSIWYG editing for HTML tables is notoriously lousy, and it seems like many editors just skip it. Should we give that a try?

noahmanger:
Sounds good to me. I agree, it always gets nasty.

noahmanger:
This has been figured out!

PR #53, opened 2015-09-15
Allow raw HTML blocks in StreamField fields.

jmcarp:
Proposed resolution to #26. @noahmanger, want to check this out and give it a try before merging? I'm hoping this will be enough to get by for now.


noahmanger:
Great.

PR #62, opened 2015-09-18
[WIP] Allow editing raw HTML in rich text fields.

jmcarp:
This is still basically a proof of concept, but far enough along for someone to take a look. @noahmanger, want to see if this runs locally? You'll need to re-run `bower install` and `npm install` for stuff to work.


jmcarp:
@noahmanger AFAICT, this deploys fine. I think we can merge.

PR #75, opened 2015-09-25
Styling error pages

noahmanger:
Resolves https://github.com/18F/fec-cms/issues/37


fec/fec/templates/500.html
```diff
@@ -1,17 +1,21 @@
...
-        <meta name="viewport" content="width=device-width, initial-scale=1" />
-    </head>
-    <body>
-        <h1>Internal server error</h1>
+{% extends "base.html" %}
 
-        <h2>Sorry, there seems to be an error. Please try again soon.</h2>
-    </body>
-</html>
+{% block body_class %}template-404{% endblock %}
```
jmcarp: Are we setting `body_class` to `template-404` for the 500 page but not for the 404 page?

noahmanger: Oops, copy and paste mistake.

PR #80, opened 2015-09-27
Adds meta tags to cms

noahmanger:
Adds the meta tags from the app to the cms.


jmcarp:
Actually, now that this is merged, I see something that needs an update--we need to use absolute URLs for images in <meta> tags. For now, the simplest thing might be to hard-code the absolute URL, but eventually we'll want to figure out how to auto-generate the correct URL, even when we're behind the proxy app.

attn @noahmanger

PR #84, opened 2015-10-02
Restore admin user bar.

### LINKED ISSUE #49: Restore Wagtail admin user bar
We disabled the Wagtail user bar in 754fcaef73a4fc40689859f562a0a7e39e6e89ad due to a configuration issue such that the user bar widget cannot be included in an iframe. At the time, we discussed a possible resolution with @dlapiduz: modifying the cloud foundry nginx config to not override `x-frame-options` if that header is already included in the response. Now that we're getting closer to a public launch, it would be great to revisit this issue.

@dlapiduz: is this something we can get onto the devops queue? Would we need formal justification of the request? I'd like to set `x-frame-options` to `sameorigin` instead of `deny`, which AFAICT should be safe for our application.

jmcarp:
* Restore user bar template tag.
* Respect x-forwarded-proto header.

Thanks @dlapiduz!

[Resolves #49]

### DIFF
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -150,6 +150,7 @@ COMPRESS_PRECOMPILERS = (
 # Proxy settings
 
 USE_X_FORWARDED_HOST = True
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 
 
 # Wagtail settings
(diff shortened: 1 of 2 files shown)


jmcarp:
Ping @LindsayYoung and/or @noahmanger for review. This restores a pretty handy built-in feature: if you're already logged in as an administrator, you get a little wagtail icon in the upper right corner of the screen that lets you edit the current page.

PR #86, opened 2015-10-03
Keyboard exit ethnio

noahmanger:
Adds a very crude hack to allow the Ethnio popup to be closed by keyboard. Because both the pre-screener modal and the screener modal are closed by `<div>`s masquerading as buttons, it took some ugly code. 

First, it adds a tabindex to the initial close button on the pre-screener modal and fires a click event when hitting enter on it.

Second, if the actual screener is opened, it adds a listener to the body to listen for `escape` and then simply removes the iframe. 

I just heard back from Ethnio that they're looking into this accessibility issue, but in the mean time this can work.

Resolves https://github.com/18F/openFEC-web-app/issues/762

### DIFF
--- a/fec/home/templates/home/home_page.html
+++ b/fec/home/templates/home/home_page.html
@@ -102,5 +102,32 @@
 
 {% block extra_js %}
 <!-- Ethnio Activation Code -->
-<script type="text/javascript" language="javascript" src="//ethn.io/70862.js" async="true" charset="utf-8"></script>
+<script type="text/javascript" language="javascript" src="https://ethn.io/70862.js" async="true" charset="utf-8"></script>
+<script type="text/javascript">
+  // Hack to add a tabindex to the ethnio popup close button
+  var target = document.querySelector('body');
+  var observer = new MutationObserver(function(mutations){
+    mutations.forEach(function(mutation) {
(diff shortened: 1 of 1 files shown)


fec/home/templates/home/home_page.html
```diff
@@ -102,5 +102,34 @@ <h2 class="feature__title">The FEC API: Use our data in your project</h2>
...
+            if (e.keyCode === 13) {
+              $(this).click();
+              observer.disconnect();
+            }
+          });
+        }
+        if ( mutation.addedNodes[0].id === 'ethnio-screener-id-70862' ) {
+          $('body').on('keyup', function(e){
+            if (e.keyCode === 27) {
+              $('iframe').remove();
```
jmcarp: There can be other iframes on the page--for example, the wagtail admin widget. This isn't a big deal, but if it's simple, could we remove only the ethnio iframe here?

noahmanger: I ended up reaching out to Ethnio to see if they could just fix this themselves and it seems like they may have. I'll double check, but this is likely unnecessary n ow.

jmcarp: :tada:

noahmanger:
Ok, Ethnio updated their script to close the popups on escape. Still not tabbable, but they said they're working on that. Just going to close this.

PR #90, opened 2015-10-08
Adding contact page

noahmanger:
Adds a new contact page template, though there's no map yet:

screen shot 2015-10-08 at 4 31 41 pm

I'm thinking we may wait on the map while we figure out our long term map solution. This is the kind of thing we could use our custom map tiles for (if we get them) rather than introducing a third map style.

Thoughts?

Depends on https://github.com/18F/fec-style/pull/135

### DIFF
--- a/fec/home/templates/home/contact_page.html
+++ b/fec/home/templates/home/contact_page.html
@@ -1,5 +1,6 @@
 {% extends "base.html" %}
 {% load wagtailcore_tags %}
+{% load staticfiles %}
 {% block body_class %}template-{{ self.get_verbose_name | slugify }}{% endblock %}
 
 {% block content %}
@@ -13,7 +14,7 @@
       <ul>
         <li class="contact-item">
           <div class="contact-item__icon">
-            <img src="/static/img/i-phone--primary.svg" alt="Icon of a phone">
+            <img src="{% static "img/i-phone--primary.svg" %}" alt="Icon of a phone">
           </div>     
           <div class="contact-item__content">
             <h5 class="contact-item__title">Toll free</h5>
@@ -37,7 +38,7 @@
       <ul>
         <li class="contact-item">
           <div class="contact-item__icon">
-            <img src="/static/img/i-mail--primary.svg" alt="Icon of a stamp">
+            <img src="{% static "img/i-mail--primary.svg" %}" alt="Icon of a stamp">
           </div>     
           <div class="contact-item__content">
             <span class="t-block">Federal Election Commission</span>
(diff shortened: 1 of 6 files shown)


noahmanger:
Other note: @jenniferthibault it doesn't quite re-flow the way you had it in the mockups for mobile. Because the left column is all one column, that all gets stacked on top of the "Special request" items (rather than having mailing address below). Is that ok? Otherwise we need to get kind of hacky re-arranging things.

jenniferthibault:
@noahmanger leaving the map out for now makes sense to me. 

From a content flow perspective on mobile, it is a little weird to have the mailing address between two inquiry sections, but maybe better than having it at the very end without a map, and risk getting missed. When I made the mobile mockup I remember thinking the column re-arranging was going to be a problem, then did a "well, I'll just see what Noah says when it comes to it...." All going according to plan then!

One little thing—I think we need a more visible active/"you are on this page" state for the Contact/About nav. Any thoughts? I was originally thinking this to match the others, thoughts?

screen shot 2015-10-09 at 10 45 48 am

noahmanger:
Cool. Well the other option for the mailing address would be to change the desktop view so it's in it's own row below the inquiry stuff. Then it would be easy to flow it below.

And good call on the active state. I started an issue a while ago to get this going throughout the app, so I'll add it there https://github.com/18F/openFEC-web-app/issues/741

fec/home/templates/home/contact_page.html
```diff
@@ -0,0 +1,85 @@
...
+<article class="main">
+  <div class="container">
+    <h1 class="t-ruled--bottom t-ruled--bold t-display">{{ self.title }}</h1>
+    <div class="main__content--full u-padding--top">
+     <div class="usa-width-one-third">
+      <h2>General inquiries</h2>
+      <ul>
+        <li class="contact-item">
+          <div class="contact-item__icon">
+            <img src="/static/img/i-phone--primary.svg" alt="Icon of a phone">
```
jmcarp: Prefer `{% static %}` tag over hard-coded URLs.

fec/home/templates/home/contact_page.html
```diff
@@ -0,0 +1,85 @@
...
+            <h5 class="contact-item__title">Teletypwriter (TTY) for the hearing impaired</h5>
+            <span class="t-block">202-219-3336</span>
+          </div>                    
+        </li>
+      </ul>
+      <h2>Mailing address</h2>
+      <ul>
+        <li class="contact-item">
+          <div class="contact-item__icon">
+            <img src="/static/img/i-mail--primary.svg" alt="Icon of a stamp">
```
jmcarp: As above.

noahmanger:
@jmcarp good catch on the hardcoded URLs. Apparently never updated those anywhere in the app, so I just did that.

jmcarp:
Rad. Are this patch and its complement in fec-style ready to go?

noahmanger:
Yep.

PR #92, opened 2015-10-14
Hmac

jmcarp:
Verify HMAC signatures from proxy app in production. Should be ready for merge once hmac_authentication_py is published.

cc @mbland


jmcarp:
Ready for review @LindsayYoung.

noahmanger:
What's the deal with this? Still something we need to do?

noahmanger:
Is this still relevant? cc @ccostino

jmcarp:
I'm feeling ambivalent about this now. The goal is to prevent users from accessing `fec-prod-cms.18f.gov`, and require them to access the site through the proxy at `beta.fec.gov`. But there's really no harm in letting people see the site without the proxy, and this patch does introduce the possibility that we screw up the configuration and block everyone from seeing the site.

¯\_(ツ)_/¯

PR #94, opened 2015-10-15
Adding staging banner to cms

noahmanger:
@jmcarp I added a separate variable for `FEC_CMS_STAGING` for when we have different staging and develop environments, and then added logic into base.py to set `FEC_CMS_ENVIRONMENT` to either `STAGING` or `DEVELOP`. Is there a better way to do this?


fec/fec/templates/base.html
```diff
@@ -56,7 +64,7 @@
                 <span class="t-bold site-nav__subtitle">Browse:</span>
               </li>
               <li class="site-nav__item">
-                <a class="site-nav__link site-nav__sublink" href="{{ settings.FEC_APP_URL }}/candidates/?cycle=2012&cycle=2014&cycle=2016&candidate_status=C&candidate_status=P&candidate_status=F">Candidates</a>
+                <a class="site-nav__link site-nav__sublink" href="{{ settings.FEC_APP_URL }}/candidates">Candidates</a>
```
jmcarp: Was this supposed to be reverted?

noahmanger: Oh weird. No. Not sure how that happened.

jmcarp:
We could set `FEC_CMS_ENVIRONMENT` to "dev", "stage", or "prod" in the environment directly. The nice thing about using boolean-ish variables is that there's no way to ask for environments that don't exist (`FEC_CMS_ENVIRONMENT=foo`), but it's also possible to do things that don't make sense like `FMC_CMS_PRODUCTION=true FEC_CMS_STAGING=true`. So not a big deal, but now that we have distinct behaviors for dev, stage, and prod, we can switch to a single env var. I can make that change when I merge.

noahmanger:
Ok cool. That makes sense. Can you do it on the web app too?

PR #95, opened 2015-10-16
Add monitoring with new relic.

### LINKED ISSUE #30: In order to detect and fix errors as they occur, set up New Relic monitoring
So we notice if the site goes down.

jmcarp:
[Resolves #30]


jmcarp:
cc @LindsayYoung

PR #96, opened 2015-10-20
Removing the statutory filtering, it did not work well.

LindsayYoung:
This links to all candidates active in the last 6 years.


jmcarp:
@LindsayYoung: For record-keeping, was there an issue where we talked about why the candidate status filter was problematic? And if not, could you say more here? Sorry if I missed this!

PR #101, opened 2015-10-27
Adds missing closing tags to fix display issue on IE11

noahmanger:
Minor fix that fixes a weird display issue I noticed on IE 11 which the other browsers apparently ignored.


noahmanger:
Would be nice to merge for launch.

PR #128, opened 2015-11-18
Add .about.yml.

jmcarp:
Note: Much of this file is copied from the openFEC repo. Once
inheritance behavior via the `parent` attribute is finalized, we should
be able to delete everything but the unique fields from this file.

cc @mbland

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -1,0 +1,60 @@
+---
+# .about.yml project metadata
+#
+# Short name that acts as the project identifier (required)
+name: fec-cms
+
+# Full proper name of the project (required)
+full_name: fec-cms
+
+# The type of content in the repo
+# values: app, docs, policy
+type: app
+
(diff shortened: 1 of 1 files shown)


.about.yml
```diff
@@ -0,0 +1,60 @@
...
+  fec-cms:
+    name: CC0-1.0
+    url: https://github.com/18F/fec-cms/blob/master/LICENSE.md
+
+# URIs for points-of-contact
+# Items:
+# - url: URL for the link
+#   text: Anchor text for the link
+contact:
+- url: leah.bannon@gsa.gov
```
mbland: Might want to add the `text:`: and add the `mailto:` prefix:

```yaml
contact:
- text: Leah Bannon
  url: mailto:leah.bannon@gsa.gov
```

mbland:
LGTM, but I suppose you want one or more of your `team:` folks to look it over before merging.

PR #132, opened 2015-11-19
Pin fec-style to last commit before beta release.

jmcarp:


noahmanger:
So this should be pinned to 1.0.0 right?

jmcarp:
:+1:

jmcarp:
Replaced by #134.

PR #133, opened 2015-11-20
Log errors using Sentry.

jmcarp:
EOM


jmcarp:
Paging @LindsayYoung or @noahmanger

PR #134, opened 2015-11-20
Pin fec-style to last commit before beta release.

jmcarp:
Replaces #132.


noahmanger:
Looks right to me, but not sure what the buildpacks and engine lines are for?

jmcarp:
Oh right, those are change we've already made to the other repos to pin versions of the Cloud Foundry buildpacks, plus node and nvm versions.

noahmanger:
Cool. LGTM.

PR #139, opened 2015-12-03
Updated article template page

noahmanger:
Updates styles for the new article template.

- It includes dummy hard-coded links for "next" and "previous" buttons just for the purposes of testing.
- The new sidebar makes use of multiple StreamFields, so once we decide this is what we want, we'll need to update sidebar content areas for each page in the CMS

image

### DIFF
--- a/fec/home/templates/home/checklist_page.html
+++ b/fec/home/templates/home/checklist_page.html
@@ -6,7 +6,7 @@
 {% block content %}
 
 <header class="page-header slab--secondary">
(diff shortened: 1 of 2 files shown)


jenniferthibault:
Took a look. One request, if there's time:
Is it possible to make the previous/next arrows slightly smaller? Close to these proportions:
screen shot 2015-12-02 at 8 23 55 pm

Thanks!

noahmanger:
Got it.

PR #145, opened 2015-12-03
Removing next / prev nav

noahmanger:
This removes the previous and next navigation from articles while we wait to figure out how exactly we want it to work (https://github.com/18F/fec-cms/issues/144)

Resolves https://github.com/18F/fec-style/issues/194


noahmanger:
Ok, this is ready to go. In addition to removing the "next" / "previous" items from the article page, it also updates the navigation so that when you hover over "Campaign finance data" you get "Campaign finance data home" as an option.

PR #151, opened 2015-12-07
Manage Cloud Foundry environment with cfenv.

jmcarp:
EOM


jmcarp:
@LindsayYoung can you take a look? This is a tiny patch that uses py-cfenv to parse CF environment variables, like we're doing in the other FEC repos.

PR #153, opened 2015-12-10
Sticky side nav

### LINKED ISSUE #146: Make page-nav menu sticky
Per the new mockups for the registration and reporting pages, the side nav should be sticky and highlight the section that you're on.

noahmanger:
- Implements the component-sticky library to sticky the side nav
- Adds a module based on scroll monitor to highlight active elements of
a table of contents
- Implements on reg. & reporting landing page and checklist page

example

Resolves #146 #152

### DIFF
--- a/fec/fec/static/js/toc.js
+++ b/fec/fec/static/js/toc.js
@@ -1,0 +1,68 @@
+'use strict';
+
+/* global require, module */
+
+var $ = require('jquery');
+var scrollMonitor = require('scrollmonitor');
+
+/**
+ * Table of Contents widget
+ *
+ * 1. Takes a list of links and finds all sections with IDs matching their hrefs
+ * 2. Adds scrollwatchers to highlight the menu item when the section reaches top of viewport
+ * 3. Animates the document to scroll to the section when clicking the link
+ *
+ * @constructor
+ * @param {string} selector - Selector for the navigation menu for the TOC
+ */
+
+function TOC(selector) {
+  this.$menu = $(selector);
+  this.sections = this.getSections();
+  this.offset = (-1 * scrollMonitor.viewportHeight);
+  this.addWatchers();
+  this.$menu.on('click', 'a', this.scrollTo.bind(this));
+}
+
+TOC.prototype.getSections = function() {
+  var sections = [];
+  this.$menu.find('a').each(function() {
+    var section = $(this).attr('href');
+    sections.push(section);
+  });
+
+  return sections;
+};
+
+TOC.prototype.addWatchers = function() {
+  var self = this;
+
+  this.sections.forEach(function(section) {
+    var elm = document.querySelector(section);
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,9 +1,11 @@
 {
   "dependencies": {
     "ace-builds": "1.2.2",
+    "component-sticky": "^1.0.0",
```
jmcarp: Prefer pinning to exact versions so we don't get surprised when dependencies update.

fec/fec/static/js/toc.js
```diff
@@ -0,0 +1,68 @@
...
+ * 3. Animates the document to scroll to the section when clicking the link
+ *
+ * @constructor
+ * @param {string} selector - Selector for the navigation menu for the TOC
+ */
+
+function TOC(selector) {
+  this.$menu = $(selector);
+  this.sections = this.getSections();
+  this.offset = (-1 * scrollMonitor.viewportHeight);
```
jmcarp: What happens if the viewport height changes after the TOC gets drawn?

jmcarp: Also, no need for parentheses here.

fec/fec/static/js/toc.js
```diff
@@ -0,0 +1,68 @@
...
+  this.addWatchers();
+  this.$menu.on('click', 'a', this.scrollTo.bind(this));
+}
+
+TOC.prototype.getSections = function() {
+  var sections = [];
```
jmcarp: I haven't tested this out, but it should do the same thing as the current code in fewer lines.

```
return this.$menu.find('a').map(function(idx, elm) {
  return $(elm).attr('href');
});
```

jmcarp:
I know you already know this, but I can't help saying it: tests! We don't have javascript testing or travis configured for this repo, so we can borrow the configuration from another repo, or move the js to fec-style.

fec/fec/static/js/toc.js
noahmanger: Good point. I'm looking into it.

fec/fec/static/js/toc.js
noahmanger: :+1:

fec/fec/static/js/toc.js
```diff
@@ -0,0 +1,73 @@
...
+  var section = $link.attr('href');
+  var sectionTop = $(section).offset().top;
+  $(document.body).animate({
+    scrollTop: sectionTop
+  });
+};
+
+TOC.prototype.updateWatchers = function() {
+  var newOffset = -1 * window.innerHeight;
+  [].foreEach.call(this.watchers, function(watcher) {
```
jmcarp: Pretty sure `foreEach` isn't a thing. Also, since we're already using underscore in fec-style, might as well use `_.each` instead of the `[].forEach` hack.

PR #160, opened 2015-12-15
Enabling registration card on home

noahmanger:
- Un-grays out the registration and reporting card on the home page
- Makes the icon and words a link to /registration-and-reporting

image

@emileighoutlaw : we want to do this, right?


emileighoutlaw:
Must have missed the email notification for this, somehow. Yes please!

PR #168, opened 2015-12-17
De-linking inactive sidebar links

noahmanger:
Let’s wait until we’ve gotten the thumbs up from FEC before we merge this.

Also, not technically a hot fix as this isn’t a gitflow enabled repo, so we’ll need to manually merge to master and develop.


leahbannon:
FEC gave :+1:

PR #171, opened 2015-12-19
Calendar

jmcarp:
* [x] Implement fullcalendar
* [x] Add filter sidebar
* [ ] Write unit tests


noahmanger:
It looks like we're missing filters for committee hearings / meetings type of events like these:

image

Or am I missing something?

.travis.yml
```diff
@@ -0,0 +1,14 @@
+language: node_js
+node_js: "0.12"
```
ccostino: Shouldn't we change this to a more recent Node version?

package.json
```diff
@@ -1,18 +1,42 @@
 {
+  "scripts": {
+    "test": "karma start",
+    "test-single": "karma start --single-run --browsers PhantomJS"
+  },
   "dependencies": {
     "ace-builds": "1.2.2",
     "component-sticky": "1.0.0",
-    "fec-style": "1.4.1",
+    "fec-style": "git://github.com/18f/fec-style#styling-calendar",
```
ccostino: This dependency is now pointing to a specific branch as opposed to a version?

noahmanger: Oh yeah, we'll want to change this when we merge in https://github.com/18F/fec-style/pull/238 and bump the version number.

ccostino: @noahmanger:  Should that PR be reviewed as well and merged in first?

noahmanger: Ah, yes it should. It was assigned to @jmcarp but I don't care who does it.

ccostino:
@jmcarp and I talked a bunch about this PR and decided that more tests would come at a later time when things settle down a bit more and there's more concrete use-cases to go on.  We're merging what has been done so far.

PR #187, opened 2016-01-28
Document data export process.

### LINKED ISSUE #184: Document process for exporting CMS database
Exporting data from the CMS database is kind of a pain, since we don't have direct access to the database. I'll refresh my memory on the process, then demo to @LindsayYoung and/or @ccostino and document how it works.

jmcarp:
[Resolves #184]

To review: @LindsayYoung and/or @ccostino, try out the instructions in `README.rst`, and revise or ping me if you run into issues or find anything unclear.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,7 +14,7 @@ We welcome you to explore, make suggestions, and contribute to our code.
 This repository, `fec-cms <https://github.com/18F/fec-cms>`_: the content management system (CMS) for betaFEC.
 
 All repositories
------------------
+----------------
 - `FEC <https://github.com/18F/fec>`_: a general discussion forum. We compile `feedback <https://github.com/18F/fec/issues>`_ from betaFEC’s feedback widget here, and this is the best place to submit general feedback.
 - `openFEC <https://github.com/18F/openfec>`_: betaFEC’s API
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -31,10 +31,10 @@ We’re thrilled you want to get involved!
 
 
 Set up
-============
+======
 
 Installation
------------------
+------------
```
ccostino: Is there anything else that needs to be added to the PostgreSQL setup, such as creating a database, the name associated with it, and any users?

jmcarp: No, we haven't changed the default of using sqlite for development, so there's no setup to do here. Setting up databases in cloud.gov is a one-liner that's documented under "Deploy".

ccostino:
@jmcarp:  One thing I noticed (thought it's not with your changes) are the steps involved with running the few `./manage.py` commands are a bit out of order.  The `createsuperuser` step should come after the two migration related steps.

LindsayYoung:
I really think followed the instructions exactly with `--exclude auth.permission` But I am seeing the fec super user account and password, I think we need `--exclude auth.user` too.

Got most of this working, I do get an error when I try to load, I think it was because I already added a calendar page.

Just wanted to note that the conflict is easy and I haven't forgotten this.

PR #192, opened 2016-02-02
Updates verb tenses

emileighoutlaw:
I learned from GOV.UK that headers with verbs that sound like an imperative (something you can do right here and now) are confusing when websites are just providing content about the thing. For example, "Register now" just tells people *how* to register. It doesn't provide the service.

Have updated verb tenses in the checklist items to gerunds (-ing verbs) to clarify that we're just providing content around "Registration and reporting" in this section. 

Resolves: https://github.com/18F/fec-cms/issues/190

### DIFF
--- a/fec/home/templates/home/checklist_page.html
+++ b/fec/home/templates/home/checklist_page.html
@@ -102,7 +102,7 @@
               <li><a href="/registration-and-reporting/pre-election-reports/">Pre-election reports</a></li>
               <li><a href="/registration-and-reporting/48-hour-notices/">48-Hour Notices</a></li>
               <li><a href="/registration-and-reporting/post-general-election-reports/">Post-general election reports</a></li>
-              <li><a href="/registration-and-reporting/terminate-committee/">Committee termination</a></li>
+              <li><a href="/registration-and-reporting/terminate-committee/">Terminating a committee</a></li>
             </ul>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


fec/home/templates/home/checklist_page.html
```diff
@@ -102,7 +102,7 @@ <h4 class="sidebar__title">Already registered?</h4>
               <li><a href="/registration-and-reporting/pre-election-reports/">Pre-election reports</a></li>
               <li><a href="/registration-and-reporting/48-hour-notices/">48-Hour Notices</a></li>
               <li><a href="/registration-and-reporting/post-general-election-reports/">Post-general election reports</a></li>
-              <li><a href="/registration-and-reporting/terminate-committee/">Terminate a committee</a></li>
+              <li><a href="/registration-and-reporting/terminate-committee/">Committee termination</a></li>
```
jenniferthibault: Tiny tiny, but wouldn't this be `Terminating a committee` using this pattern? @emileighoutlaw

emileighoutlaw: Yeah, I did think about it: -tion words also take away the immediacy of action (by transforming verbs into nouns). 

In this case, I was just going for variation (my eyes were tired of looking at -ings). Do you think it looks out of place?

jenniferthibault: I don't think it _looks_ out of place, but it does seem to break a pattern of building consistent language tense to me. `Termination` is more passive than `Terminating`

emileighoutlaw: Consistency! You've won my <3. Just made the change.

PR #206, opened 2016-02-19
Preserve query params in export URLs.

### LINKED ISSUE #205: add api key to csv call
the download csv calendar option needs the API key.

jmcarp:
[Resolves #205]


noahmanger:
This works for me, but I don't get why the build is failing to find gulp. https://travis-ci.org/18F/fec-cms/builds/110427535#L315

PR #228, opened 2016-03-09
Feature/more cal tests

### LINKED ISSUE #204: Discussion: Flesh out calendar tests
The calendar widget has automated tests for basic functionality (filtering and making requests), but lots of  behavior is untested. I think we need more tests than we have, but I also think it's overkill to chase 100% coverage. Looking for input from @noahmanger @ccostino @LindsayYoung on where tests would be most useful. What's liable to break in the future--what kinds of tests would help folks feel confident that important behaviors aren't going to break?

noahmanger:
Adds more tests for calendar.js, calendar-helpers.js and calendar-tooltip.js.

We still need tests for calendar-list-view.js, but I didn't write that, so I'm not sure what to test.

Resolves #204

### DIFF
--- a/fec/fec/static/js/calendar.js
+++ b/fec/fec/static/js/calendar.js
@@ -229,6 +229,7 @@ Calendar.prototype.handleRender = function(view) {
     this.manageListToggles(view);
   } else if (this.$listToggles) {
     this.$listToggles.remove();
+    this.$listToggles = null;
   }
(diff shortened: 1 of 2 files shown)


noahmanger:
I can't tell why the tests are failing. They work for me locally. Anyone know? I tried clearing the travis cache.

noahmanger:
Merged @jmcarp 's fixes in here. Should someone else review, or are we good to merge?

PR #230, opened 2016-03-10
Clean up calendar tests.

jmcarp:
* Fix width-dependent tests
* Fix order-dependent tests
* Split up tests by category
* Delete unused imports
* Fix linter errors

@noahmanger: do you have a linter running in your editor? I fixed some linter errors in this patch.

### DIFF
--- a/fec/fec/static/js/calendar.js
+++ b/fec/fec/static/js/calendar.js
@@ -215,7 +215,7 @@ Calendar.prototype.styleButtons = function() {
 };
 
 Calendar.prototype.defaultView = function() {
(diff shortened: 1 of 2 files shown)


jmcarp:
BTW, running `npm test` uses chrome. Travis doesn't get to use chrome, so we use phantom there. To run exactly what travis runs, use `npm run test-single`.
@noahmanger

noahmanger:
Ah! Ok, thanks for clarifying. And for cleaning up.

I do have a linter set up, but I just tested it and apparently it's not working. I'll fix it. Sorry about that, I thought it was running.

PR #231, opened 2016-03-10
Feature/new accordion

noahmanger:
Adds the new accordions and glossary to the CMS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
     "test-single": "karma start --single-run --browsers PhantomJS"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -4,9 +4,11 @@
     "test-single": "karma start --single-run --browsers PhantomJS"
   },
   "dependencies": {
+    "aria-accordion": "git://github.com/18f/accordion",
```
jmcarp: How about releasing these to npm? These would ideally be pinned to a specific version, not just to a git repo.

noahmanger: Ah yes! I can do that.

noahmanger: Done.

PR #233, opened 2016-03-11
[WIP] Nesting calendar filters in accordions

noahmanger:
Alongside the work we just did to put datatable filters inside accordions, this does the same for the calendar filters.

screen shot 2016-03-11 at 2 30 20 pm

This is loosely based on @jenniferthibault 's original designs, with some slight modifications:

1. I grouped "Reporting periods", "Elections" and "State" together into the same accordion item, since they are they only events that are affected by the "State" filter.
2. Because IE and EC periods aren't affected by the "State" filter (as we don't have the data for that), I broke them out to their own filter called "Reporting periods"
3. I grouped all other filters in a single accordion item for "Other events" since I think it looks weird to have each filter be in an accordion item of their own.

This doesn't _need_ to get into the release, but wanted to propose it on the off-chance that it's an easy change that people agree with.

In particular, curious what @amykort @ChristianHilland @emileighoutlaw and @jenniferthibault think.

### DIFF
--- a/fec/fec/constants.py
+++ b/fec/fec/constants.py
@@ -7,7 +7,10 @@ election_types = OrderedDict([
 deadline_types = OrderedDict([
     ('report-M', 'Monthly'),
     ('report-Q', 'Quarterly'),
-    ('report-E', 'Pre- and Post-election'),
+    ('report-E', 'Pre- and Post-election')
+])
+
+period_types = OrderedDict([
     ('IE Periods', 'Independent expenditures'),
     ('EC Periods', 'Electioneering communications')
 ])
--- a/fec/fec/static/js/fec.js
+++ b/fec/fec/static/js/fec.js
@@ -34,12 +34,14 @@ $(document).ready(function() {
   // Initialize new accordions
   $('.js-accordion').each(function(){
     var contentPrefix = $(this).data('content-prefix') || 'accordion';
+    var openFirst = $(this).data('open-first') || false;
     var selectors = {
       body: '.js-accordion',
       trigger: '.js-accordion-trigger'
     };
     var opts = {
       contentPrefix: contentPrefix,
+      openFirst: openFirst
     };
     new Accordion(selectors, opts);
   });
--- a/fec/home/templates/home/calendar_page.html
+++ b/fec/home/templates/home/calendar_page.html
@@ -11,92 +11,107 @@
 
 <section class="main__content--full">
   <div id="filters" class="filters">
-    <div class="filters__inner">
       <button id="filter-toggle" data-slide="left" class="filters__toggle">
         <span class="filters__toggle__text js-filter-toggle-text">Show filters</span>
       </button>
       <div class="filters__hider">
-        <h2 class="filter__header">
+        <h2 class="filter__header filters__inner">
           Filter events
         </h2>
-        <form id="category-filters">
-          <div class="">
-            <button class="button--sm button--primary-contrast" type="button">Apply filters</button>
-            <button class="button--sm button--neutral js-clear-filters" type="button">Clear filters</button>
+        <form id="category-filters" class="js-accordion accordion--neutral" data-content-prefix="filter" data-open-first="true">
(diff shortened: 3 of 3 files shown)


jenniferthibault:
There is a lot to say about the visual of this: I can be the first to say that it does not aesthetically look good, but that's not Noah's fault! That's just because this is happening at a point when the calendar needs to undergo some cosmetic surgery anyways. So putting visuals aside for a moment...

Grouping state, reporting deadlines, and elections into one accordion items make sense to me, and I understand what will be in it through the title "Reporting deadlines and elections".

Along the same logic, I think we need either a more descriptive title for "Other events"  or they need to remain broken down into more groups, because it obfuscates what's inside it. @emileighoutlaw ?

emileighoutlaw:
Agree with Jen on all fronts. :smile:

Right now our other events include:
- Outreach
- Advisory opinions and rulemaking
- Commission meetings
- Reporting periods
- Other events (which are Litigation and FEA periods)

I think `Other` could make sense as two groupings:

> **Meetings and outreach**
- Outreach
- Commission meetings

> **Legal events** (maybe another name if we don't like this)
- Advisory opinions and rulemaking
- Litigation

That leaves FEA periods and reporting periods as odd ducks. But I could see them filtering back up into our first category, like so:

> **Elections and reporting**
- Elections
- Reporting deadlines
- Reporting periods
- FEA periods

AmyKort:
Let's remove "Litigation." We don't anticipate having entries under that heading.

noahmanger:
The problem with including "Reporting periods" and "FEA periods" in the first group is that the state filter doesn't actually filter them (which is a data issue). 

Another suggestion:

**Reporting deadlines and elections**
- State
- Elections
- Reporting deadlines

**Meetings, outreach and legal**
- Outreach
- Commission meetings
- Advisory opinions and rulemaking

**Reporting and election activity periods**
- Reporting periods
- FEA periods

emileighoutlaw:
I think that's a good compromise, if it's what the data allows :wink: 

Microcopy: 

I'd love to spell out FEA to say `Federal Election Activity (FEA)`

Also, could we make `Reporting and election activity periods` into `Reporting and election periods` ?

AmyKort:
"Reporting and election activity periods" is giving me pause. "Reporting and election periods" is giving me more pause. Federal Election Activity is a term of art that is very different from election activity. "Election period" doesn't mean much to me at all. 

When we put IE, EC and FEA periods together, we're lumping together very dissimilar things. You can make an IE any time and it is still an IE, but if you make one within the period, you have to report it differently. My understanding of the EC and FEA periods is that activity you conduct becomes an EC or FEA by virtue of the face that you conduct it within the period--the definition includes the fact that the activity is undertaken in proximity to an election. So, these are more than "reporting" periods, they also affect the way that you can pay for that activity. 

Can we just call them IE, EC and FEA Periods? We're pretty in the weeds at this point, so I think anyone who is interacting with this part of the calendar knows what these things are. Calling them election periods makes me think . . . is that an election cycle? I can't even attach that to anything to that term.

emileighoutlaw:
This is very helpful to know! Thanks for being so knowledgeable, Amy! We'll confer on this.

noahmanger:
This is all good. Since there's still some things we're working through on this, and it's not critical, let's leave this for the next release so we can focus on finishing up the things we need to for this one.

AmyKort:
I agree on waiting for the next release--thanks! 

From a content standpoint, how awful is it to say something like "Other Reporting and Compliance Periods" or something like that? Is it too much like the name of that casserole you make when you have a lot of unrelated leftovers lying around in your fridge at the end of week? You know, "corned beef and pad thai surprise," or whatever we had last week. @emileighoutlaw

emileighoutlaw:
Not adding anything substantive at the moment — I just didn't want to forget to say to say that your corned beef and pad thai surprise analogy is brilliant and hilarious. And I laughed for so long when I read it.

noahmanger:
This work is now contained in #256 , so I'm closing.

PR #236, opened 2016-03-12
Feature/ssf pages

noahmanger:
I know this is an ugly way of doing this, but in the interest of releasing the page this week it seemed like the simplest solution. I was starting to look at making the checklist page more of a dynamic page with the `StreamFields` and it doesn't seem like that much work, but probably more than we could get done for this release.


noahmanger:
Ok this is ready for review.

emileighoutlaw:
Looks great to me! Thanks for doing all this legwork, Noah!

PR #240, opened 2016-03-15
Making the state filter aqua-on-navy to stay consistent with data site

noahmanger:


noahmanger:
Gah, this should be targeting the release branch.

PR #256, opened 2016-03-22
Feature/calendar visual update

noahmanger:
Beginning of work to implement the designs in https://github.com/18F/fec-cms/issues/222

image

Remaining tasks:
- [x] Fix bug where you can't use `enter` to open an event after you've already used it once
- [x] Break up filters into accordions
- [x] Redesign filter toggle button
- [x] Move "Clear filter" button to tag area
- [x] Fix highlighting of "today" cell in the calendar
- [x] Update style on list view

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "glossary-panel": "0.1.2",
     "ace-builds": "1.2.2",
     "component-sticky": "1.0.0",
-    "fec-style": "git+https://github.com/18f/fec-style#filter-data-update",
+    "fec-style": "~1.8",
     "fullcalendar": "2.5.0",
     "jquery": "2.1.4",
     "js-beautify": "1.5.10",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "glossary-panel": "0.1.2",
     "ace-builds": "1.2.2",
     "component-sticky": "1.0.0",
-    "fec-style": "~1.7",
+    "fec-style": "git+https://github.com/18f/fec-style#filter-data-update",
     "fullcalendar": "2.5.0",
     "jquery": "2.1.4",
     "js-beautify": "1.5.10",
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ node_js: "5.5.0"
 
 sudo: false
 
-cache:
-  directories:
-    - node_modules
-
 before_script:
   - travis_retry npm install -g gulp
 
--- a/fec/fec/constants.py
+++ b/fec/fec/constants.py
@@ -29,7 +29,6 @@ meeting_types = OrderedDict([
 
 rule_types = OrderedDict([
     ('AOs and Rules', 'AOs and rulemakings'),
-    ('Litigation', 'Litigation')
 ])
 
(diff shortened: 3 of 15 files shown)


noahmanger:
Ok, this is getting close. @jenniferthibault can you just post the latest mocks for how the list view should look? Then we can pair on this to finish up the final bits.

I can explain the couple small departures from the mocks then too.

image

emileighoutlaw:
Just popping in to say this is looking amazing, y'all!

jenniferthibault:
@noahmanger thanks for your patience, I should have gotten ahead on that as well.

Here's what I have mocked up, the only change would be to add a column for the event category. Seeing your screenshot above, I know this might need to be tweaked a bit. Looking forward to pairing up!

screen shot 2016-03-23 at 4 16 30 pm

noahmanger:
Ok, I think this is ready for review. 

@jenniferthibault here's the new list view:

image

We're not showing separate titles above the descriptions like you have in your mock (because the data is largely redundant and awkward to read again), so I put the category on top of the description in order to give it some visual separation. What do you think?

We still need to resolve the grouping question, but I don't think that needs to hold up review or merging of this.

emileighoutlaw:
From a content standpoint, I like what you've done by including category rather than the (largely redundant) titles. It does visually seem to need that bold and regular text — so this looks good to me, words wise.

jenniferthibault:
**Some small design changes:**
- I think `Sort by: Date  |  Category` feels more accurate for a calendar than `Time  |  Category` (even though it would be also chronologically ordered by time)
- Could we please make the font size of the Date/Category buttons equal to the font site of "Sort by" ?Right now it's smaller 
screen shot 2016-03-24 at 12 02 41 pm
- This wasn't in the mockup but we mentioned it in review yesterday, is it possible to add some signifier to the current date in list view? Like the vertical bar pattern we use on browse table detail patterns?
screen shot 2016-03-24 at 12 02 15 pm


**And some information presentation thoughts:**

My sense is that people need to know what the event is more than they need to know the category, which is why it would be ideal to have event titles, and give them most visual priority (i.e., bolding and their own line). If we can't have titles, and if we can't make sure the blurb is short enough to be helpful, I'd prefer to have the category be it's own column instead of acting as an event header. 
Example:
screen shot 2016-03-24 at 12 17 53 pm


It's not super helpful to see a bunch of events in a row that seem to be titled "Election", UNLESS, we could separate the state info, and pre-pend the category with that. 
Example:

screen shot 2016-03-24 at 12 18 16 pm

noahmanger:
Ready for review! Assigning @ccostino .

package.json
```diff
@@ -8,7 +8,7 @@
     "glossary-panel": "0.1.2",
     "ace-builds": "1.2.2",
     "component-sticky": "1.0.0",
-    "fec-style": "~1.7",
+    "fec-style": "git+https://github.com/18f/fec-style#filter-data-update",
```
ccostino: Just need to update to point to next minor version, should be `~1.8` once we're ready with `fec-style`.

ccostino:
We're set now on this too, thanks @noahmanger!

PR #260, opened 2016-03-30
Legal resources placeholder page

adborden:
This adds a new `legal` module and page for the [legal resources work](https://projects.invisionapp.com/share/5N6NL76QM#/screens/145835129). This serves as a placeholder where we can put functionality until we figure out where they should live permanently.

This adds a Legal Resources tab in the nav, hidden behind a feature flag. This won't show up on production, but the plan is to have it show in a staging environment so we can get some feedback from clients.

screenshot from 2016-03-29 17-09-56

cc @anthonygarvan

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,9 +88,9 @@ Settings: ::
 Features
 ~~~~~~~~~~~~~~~~~~~~~~
 
-[settings.py](https://github.com/18F/fec-cms/blob/develop/fec/fec/settings/base.py) includes a set of `FEATURES` which can be enabled using environment flags or via `settings.local`.
+`settings.py <https://github.com/18F/fec-cms/blob/develop/fec/fec/settings/base.py>`_ includes a set of ``FEATURES`` which can be enabled using environment flags or via ``settings.local``. ::
 
-    $ FEC_CMS_FEATURE_LEGAL=1 python fec/manage.py runserver
+    FEC_FEATURE_LEGAL=1 python fec/manage.py runserver
 
 
 Run
--- a/README.rst
+++ b/README.rst
@@ -85,6 +85,14 @@ Settings: ::
 
     FEC_APP_URL = 'http://localhost:3000'
 
+Features
+~~~~~~~~~~~~~~~~~~~~~~
+
+[settings.py](https://github.com/18F/fec-cms/blob/develop/fec/fec/settings/base.py) includes a set of `FEATURES` which can be enabled using environment flags or via `settings.local`.
+
+    $ FEC_CMS_FEATURE_LEGAL=1 python fec/manage.py runserver
+
+
 Run
 -----------------
 
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -179,11 +179,8 @@ FEC_CMS_ENVIRONMENT = ENVIRONMENTS.get(os.getenv('FEC_CMS_ENVIRONMENT'), 'DEVELO
 CONTACT_EMAIL = 'betafeedback@fec.gov';
 CONSTANTS = constants
 
-def feature_enabled(feature):
-    return os.getenv(feature, False) == '1'
-
 FEATURES = {
-    'legal': feature_enabled('FEC_CMS_FEATURE_LEGAL')
+    'legal': bool(os.getenv('FEC_FEATURE_LEGAL', ''))
 }
 
 if os.getenv('SENTRY_DSN'):
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -53,6 +53,7 @@ INSTALLED_APPS = (
     'fec',
     'search',
     'home',
+    'legal',
 )
 
 MIDDLEWARE_CLASSES = (
@@ -85,6 +86,7 @@ TEMPLATES = [
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',
                 'fec.context.show_settings',
+                'fec.context.features',
             ],
         },
     },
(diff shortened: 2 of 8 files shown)


emileighoutlaw:
This looks so, so awesome! Wohoo, @aborden! 

Can I suggest two quick content tweaks before merge?
1. `Legal resources` in the nav bar and in the header above the search should have a lowercase `r`. (We're going with title case across the site because users have found it more approachable)
2. In the subhead, what if we said something like `Find the information you need` or `Find the materials you need` or `Find the documents you need`. That way we could avoid having "resources" twice on the same page (and I also think the goal of the subhead is to be more specific than the main head).

adborden:
@emileighoutlaw thanks for the feedback, I pushed a new commit with those changes. Went with "Find the information you need".

emileighoutlaw:
Yay! Thanks thanks!

noahmanger:
This sounds like a good plan. I don't have any concerns, but would like @ccostino or @LindsayYoung to review.

adborden:
I'm going to rename the env variable to `FEC_FEATURE_LEGAL`, since it's not CMS specific and will also be set for other apps.

fec/fec/settings/base.py
```diff
@@ -177,6 +179,13 @@
 CONTACT_EMAIL = 'betafeedback@fec.gov';
 CONSTANTS = constants
 
+def feature_enabled(feature):
+    return os.getenv(feature, False) == '1'
```
LindsayYoung: Small thing I noticed, we have been using true rather than 1 
Wanted to mention it incase we want that to be consistent.

README.rst
```diff
@@ -85,6 +85,14 @@ Settings: ::
 
     FEC_APP_URL = 'http://localhost:3000'
 
+Features
+~~~~~~~~~~~~~~~~~~~~~~
+
+[settings.py](https://github.com/18F/fec-cms/blob/develop/fec/fec/settings/base.py) includes a set of `FEATURES` which can be enabled using environment flags or via `settings.local`.
+
+    $ FEC_CMS_FEATURE_LEGAL=1 python fec/manage.py runserver
```
noahmanger: Use the code block syntax from above. This displays as a blockquote. Also, we're not including `$` to make pastability easier.

README.rst
```diff
@@ -85,6 +85,14 @@ Settings: ::
 
     FEC_APP_URL = 'http://localhost:3000'
 
+Features
+~~~~~~~~~~~~~~~~~~~~~~
+
+[settings.py](https://github.com/18F/fec-cms/blob/develop/fec/fec/settings/base.py) includes a set of `FEATURES` which can be enabled using environment flags or via `settings.local`.
```
noahmanger: This might be my own ignorance, but isn't the file actually called base.py? (Or is settings.py the conventional name for it?)

Also, where or what is `settings.local`?

Last, (and I'm not sure why this is), but this isn't Markdown, which that link syntax is using.

fec/fec/templates/partials/search-hero.html
```diff
@@ -0,0 +1,28 @@
+{% block content %}
+  <section class="hero hero--primary" aria-labelledby="hero-search">
+    <div class="container">
+      <div class="hero__heading">
+        <h1 id="hero-heading" class="hero__title t-display">{{ search_hero_heading | default:'Search betaFEC' }}</h1>
+        <h2 class="hero__subtitle t-ruled">Find the information you need</h2>
+      </div>
+      <form id="large-search" autocomplete="off" class="search__container js-search">
+        <div class="combo combo--search--large">
+          <select class="search__select js-search-type" name="search_type">
```
noahmanger: `select` needs an `aria-label` attribute.

fec/fec/templates/partials/search-hero.html
```diff
@@ -0,0 +1,28 @@
...
+            <option value="all">All sources</option>
+            <option value="candidates">Candidates</option>
+            <option value="committees">Committees</option>
+            <option value="aos">AOs</option>
+            <option value="murs">MURs</option>
+          </select>
+          <input type="text" name="query" class="combo__input js-search-input" placeholder="e.g. candidate, committee, advisory opinion, topic"{% if search_query %} value="{{ search_query }}"{% endif %}>
```
noahmanger: `input` needs an `aria-label` attribute.

noahmanger:
This looks good and works for me. Made a couple notes / questions. Let me know when you get to those and I'll merge.

fec/fec/settings/base.py
adborden: Hmm, I see the usage in [openFEC-web-app](https://github.com/18F/openFEC-web-app/blob/develop/openfecwebapp/config.py#L33). Does it bother anyone that `FEC_FORCE_HTTPS=false` actually _enables_ the feature?

README.rst
adborden: > This might be my own ignorance, but isn't the file actually called base.py? (Or is settings.py the conventional name for it?)

You're right, it is called `base.py`, but in a simple django app, you would just have a single `settings.py`, which is what I want to invoke in the readers' minds.

> Also, where or what is settings.local?

It's not committed to the repo, but is used to override settings in [your local dev environment](https://github.com/18F/fec-cms/blob/master/fec/fec/settings/dev.py#L13). Django convention might also call it `local_settings.py`, but in our case just drop a `local.py` in your settings dir.

> Last, (and I'm not sure why this is), but this isn't Markdown, which that link syntax is using.

Doh, yeah RST is not my thing, I'll push a fix.

adborden:
@noahmanger @LindsayYoung comments addressed, thanks for reviewing!

PR #266, opened 2016-04-05
Removes "FEC"

emileighoutlaw:
Content only change: 

We've been talking about naming and branding with regards to the 18F site, and so this content piece stuck out to me when we were doing design demo today.

There's lots of available branding and visual clues that people are on the FEC website. So having the glossary called "FEC Glossary" rather than "Glossary," feels a little redundant. And any word we can remove is one less word we to distract site visitors who don't want the cognitive burden of reading. 

Anyone have any objections to this change?

This PR also adds a needed comma after "e.g.," and changes the exampled to "authorized committee" (which is in the glossary, unlike the current "committee," which is not)

### DIFF
--- a/fec/fec/templates/partials/glossary.html
+++ b/fec/fec/templates/partials/glossary.html
@@ -5,7 +5,7 @@
       class="button button--close--inverse toggle js-glossary-close"
     ><span class="u-visually-hidden">Hide glossary</span>
   </button>
-  <h2>FEC Glossary</h2>
+  <h2>Glossary</h2>
   <label for="glossary-search" class="label">Filter glossary terms</label>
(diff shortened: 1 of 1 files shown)


jenniferthibault:
@emileighoutlaw I'm alllllll for this! Great spotting.

I believe this will need a corresponding pull request to the `openFEC-web-app` repository with the same changes. (This is one of those features where the code needs to be changed in two places.)

If you want to submit that PR, I'll review and merge them both together.

emileighoutlaw:
Yes! I'm on it! I'll cc you in the openFEC-web-app PR :)

jenniferthibault:
Looooooks good!
screen shot 2016-04-05 at 11 26 05 am

PR #267, opened 2016-04-05
Updating class name to use new styleguide class

noahmanger:
Fixing a class name to go with changes in https://github.com/18F/fec-style/pull/287


jenniferthibault:
Annnd a few things here:

This should use the call to action button style, (and result in a federal blue button)
screen shot 2016-04-08 at 6 15 08 pm

screen shot 2016-04-08 at 6 25 02 pm


I'm also seeing some mixed styles on the calendar:
screen shot 2016-04-08 at 6 21 50 pm

Should be the alt style buttons here:
screen shot 2016-04-08 at 6 24 01 pm

And feedback button styles aren't coming through on this side either:
screen shot 2016-04-08 at 6 30 05 pm

noahmanger:
Good 👀 ! Will fix.

PR #270, opened 2016-04-11
Improved Autopilot Deployment

ccostino:
This changeset adds support for the Cloud Foundry autopilot plugin, Slack notifications, and Travis build enhancements to make the deployment of our CMS mirror the deployment procedures for our API and web app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
     "test-single": "karma start --single-run --browsers PhantomJS"
   },
   "dependencies": {
+    "ace-builds": "1.2.2",
     "aria-accordion": "0.1.0",
     "glossary-panel": "0.1.2",
-    "ace-builds": "1.2.2",
     "component-sticky": "1.0.0",
     "fec-style": "~1.8",
     "fullcalendar": "2.5.0",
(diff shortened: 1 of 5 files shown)


ccostino:
@LindsayYoung, if you wouldn't mind giving this a go to see if it works well for you whenever you have a moment I'd appreciate it!  I think the Slack notification needs to still be wired up, I'm not sure what's missing just yet compared to the other repos.

package.json
```diff
@@ -40,6 +40,12 @@
     "vinyl-source-stream": "^1.1.0",
     "watchify": "^3.7.0"
   },
+  "scripts": {
+    "build": "cd fec && gulp build-js",
```
noahmanger: Ah! Thanks for adding these. What about moving the gulpfile to the parent directory so we don't have to `cd` into `fec`? With this command, if you ran it twice in a row it wouldn't work, right?

ccostino: You're welcome!  Actually, it should work because you are put back in the parent directory after running the task, at least I was in my initial test.  It would be good to verify, though.

noahmanger: Ah! Yep, that's how it works.

PR #271, opened 2016-04-11
Feature/calendar bug fixes

noahmanger:
- Adds a class for multi-day events 
- Correctly highlights only one item in the "more" pop-up when a tooltip is opened

Resolves  https://github.com/18F/fec-style/issues/300
Resolves https://github.com/18F/fec-cms/issues/268

### DIFF
--- a/fec/fec/static/js/calendar-helpers.js
+++ b/fec/fec/static/js/calendar-helpers.js
@@ -38,6 +38,8 @@ function className(event) {
   var end = event.end_date ? moment(event.end_date).format('M D') : null;
   if (end && start !== end) {
     return 'fc-multi-day';
+  } else {
+    return '';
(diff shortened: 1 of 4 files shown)


fec/fec/static/js/calendar-helpers.js
```diff
@@ -32,7 +33,16 @@ function getUrl(path, params) {
     .toString();
 }
 
+function className(event) {
+  var start = event.start_date ? moment(event.start_date).format('M D') : null;
+  var end = event.end_date ? moment(event.end_date).format('M D') : null;
+  if (end && start !== end) {
+    return 'fc-multi-day';
+  }
```
adborden: Should this return '' in the negative case?

PR #274, opened 2016-04-12
Legal resources search

adborden:
Legal resources search on CMS should proxy to the web-app search.

No UI changes for this one.


noahmanger:
Just curious: why have this page here and not as part of the web app?

adborden:
On Thu, 2016-04-14 at 11:17 -0700, Noah Manger <notifications@github.com> wrote:
> Just curious: why have this page here and not as part of the web app?

Great question.. my hunch is that this will be some landing page
instructional content that FEC might want to have more control over. My
gut says it should be part of the CMS for them to be able to do that.


--
Aaron D Borden
Human and Hacker

PR #275, opened 2016-04-12
Add Notification Support

### LINKED ISSUE #272: slack integration on deploy
We can add this in a separate PR, deployment was updated in #270

ccostino:
This changeset adds the notification hook that is currently missing in the `run` script.  Closes #272.

/cc @LindsayYoung, @anthonygarvan, @adborden, @noahmanger

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,3 +1,11 @@
+#!/bin/bash
+
+# Set environment options to exit immediately if a non-zero status code
+# appears from a command or within a pipe
+set -o errexit
+set -o pipefail
+
+# Send out Slack notification(s)
 invoke notify
 
 # Build static files
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,3 +1,5 @@
+invoke notify
+
 # Build static files
 cd fec
 gulp build-js


adborden:
Doesn't `bin/run.sh` get run on every start? I.e. not just every deploy, but every time the instance is rebooted?

bin/run.sh
```diff
@@ -1,3 +1,5 @@
+invoke notify
```
adborden: Generally best practice to include a `#!` and some options for shell scripts.

```
#!/bin/bash
set -o errexit
set -o pipefail
```

These cause the script to exit immediately after a non-zero status code appear from a command or within a pipe. Generally you don't want the script to continue if one of the commands fails.

ccostino: D'oh!  You're absolutely right, especially with the hash-bang line at the beginning; I completely failed to notice that.  We'll need to update the scripts in the other repos as well.

ccostino:
@adborden: Yes, it does - this is the way the other repos are setup right now, and my focus was just on the first step of bringing parity across the deployment procedures in each of the repos we work in (especially since this was my first introduction to Cloud Foundry, autopilot, Travis, and deploying Flask and Django apps in Cloud Foundry).

That said, we can absolutely continue to work on a better way to manage this for the future.

PR #281, opened 2016-04-15
Beginning mega menu work

noahmanger:
Prepares the nav for the mega menu from https://github.com/18F/fec-style/pull/316
- Removes the old sub-lists
- Re-arranges some nav components
- Adds a global variable for the CMS url


adborden:
:+1:

PR #284, opened 2016-04-27
Adding checklist and landing page content for party committees

noahmanger:
Adds the static content for political party content.

image

image

@ccostino or @LindsayYoung can you review this in the morning and merge to develop so partners can review for an end-of-day release?

Requires adding a new page using the "Party checklist page" template. The page should be titled "Essentials for political party committees".

### DIFF
--- a/fec/home/migrations/0011_ssfchecklistpage.py
+++ b/fec/home/migrations/0011_ssfchecklistpage.py
@@ -26,4 +26,15 @@ class Migration(migrations.Migration):
             },
             bases=('wagtailcore.page',),
         ),
+        migrations.CreateModel(
+            name='PartyChecklistPage',
+            fields=[
+                ('page_ptr', models.OneToOneField(primary_key=True, auto_created=True, to='wagtailcore.Page', serialize=False, parent_link=True)),
+                ('body', wagtail.wagtailcore.fields.StreamField((('heading', wagtail.wagtailcore.blocks.CharBlock(classname='full title')), ('paragraph', wagtail.wagtailcore.blocks.RichTextBlock()), ('html', wagtail.wagtailcore.blocks.RawHTMLBlock()), ('image', wagtail.wagtailimages.blocks.ImageChooserBlock())), blank=True, null=True)),
+            ],
+            options={
+                'abstract': False,
+            },
+            bases=('wagtailcore.page',),
+        ),
     ]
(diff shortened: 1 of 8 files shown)


noahmanger:
Note: @jenniferthibault I know some of the typography is kinda wonky here. It calls for headings where we previously hadn't really used them, so it may prompt a tweak to styles. If you're ok launching as-is and updating styles in the next release, I'd prefer that just to keep this release as simple as possible. But I'm fine if you do want to make the change.

emileighoutlaw:
Also — as a content note for whomever reviews this PR. This content raises some important questions about what we want the blurbs on this page to do. For example, should they be in second person (you) or third person (they)? What questions exactly should these blurbs answer?

The inconsistencies here are notable, but I think our plan should be to launch this content as is and iterate rapidly. Which is to say, I've got my eyes on this and want to improve, but this content is OGC blessed and can't change before release.

jenniferthibault:
🙈 🔤 👍

fec/home/migrations/0011_ssfchecklistpage.py
```diff
@@ -26,4 +26,15 @@ class Migration(migrations.Migration):
...
+            name='PartyChecklistPage',
+            fields=[
+                ('page_ptr', models.OneToOneField(primary_key=True, auto_created=True, to='wagtailcore.Page', serialize=False, parent_link=True)),
+                ('body', wagtail.wagtailcore.fields.StreamField((('heading', wagtail.wagtailcore.blocks.CharBlock(classname='full title')), ('paragraph', wagtail.wagtailcore.blocks.RichTextBlock()), ('html', wagtail.wagtailcore.blocks.RawHTMLBlock()), ('image', wagtail.wagtailimages.blocks.ImageChooserBlock())), blank=True, null=True)),
+            ],
+            options={
+                'abstract': False,
+            },
+            bases=('wagtailcore.page',),
+        ),
```
ccostino: @noahmanger:  Was this auto-generated for you?  I ask because this was an existing migration and if one had already applied it, this particular addition to it would not be run.  Django would see this migration as already being applied and therefore ignore the file.

ccostino: Actually, this migration should be its own step in any case since the `SSFChecklistPage` already exists.

PR #285, opened 2016-04-27
Updates content for Registration landing page

emileighoutlaw:
Content only: 

- Updates subhead to clarify what this page is _about_ 
- Shortens and clarifies section on "What does it mean to register and report?"
- Removes arrow for "view registration categories" because content is shortened

Resolves: https://github.com/18F/fec-cms/issues/283#issuecomment-213586159 (and discussion here)
Resolves: https://github.com/18F/fec-cms/issues/282#issuecomment-214481549 (and discussion here)

### DIFF
--- a/fec/home/templates/home/landing_page.html
+++ b/fec/home/templates/home/landing_page.html
@@ -7,10 +7,7 @@
 <section class="hero hero--secondary" aria-labelledby="hero-heading">
   <div class="container">
     <h1 class="t-display">Registration and reporting</h1>
-    <h2 class="t-ruled">Register your candidacy or committee and get started reporting</h2>
-    <div class="hero__content">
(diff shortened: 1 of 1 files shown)


noahmanger:
Just go have an eyeball test, here's what this looks like:

image

Noticed a couple things I just want to confirm:
- The label for the search box now breaks onto 3 lines, which looks awkward
- The shorter content in the gray area now makes the vertical height look a little unbalanced

What do you think @jenniferthibault

emileighoutlaw:
Just shortened that text in the "Already registered?" box a little bit (It says `Find` rather than `Search for`). Maybe that'll help with the visual some?

jenniferthibault:
I missed the window on this, but I would like to revise the design! That's a lot of empty space. Making a new issue for a later sprint.

PR #292, opened 2016-05-02
Adds the feature manifest

adborden:
- Adds the feature manifest for feature space.
- Enables legal search on `feature`, `dev`, and `stage`.

### DIFF
--- a/manifest_feature.yml
+++ b/manifest_feature.yml
@@ -1,0 +1,12 @@
+---
+inherit: manifest_base.yml
+host: fec-feature-cms
+services:
+  - fec-feature-cms
+  - fec-creds-feature
+  - cms-creds-feature
+env:
+  FEC_CMS_DEBUG: "true"
+  FEC_CMS_ENVIRONMENT: "dev"
+  NEW_RELIC_APP_NAME: beta.fec.gov | cms | feature
+  FEC_API_URL: "https://fec-feature-api.18f.gov"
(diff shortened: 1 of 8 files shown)


manifest_feature.yml
```diff
@@ -0,0 +1,13 @@
...
+inherit: manifest_base.yml
+host: fec-feature-cms
+services:
+  - cms-creds-feature
+  - fec-creds-feature
+  - fec-feature-cms
+env:
+  FEC_API_URL: https://fec-feature-api.18f.gov
+  FEC_CMS_DEBUG: true
+  FEC_CMS_ENVIRONMENT: dev
```
ccostino: This should be `feature`, right?

adborden: Hmm, well, I decided that feature isn't really new environment, it's just another dev environment deployed to a new space. I thought `dev` is more correct here. This setting is only used for the development banner that displays on the site in dev and production

screenshot from 2016-05-12 13-27-27
Displaying FEATURE here doesn't seem right. Folks (like the FEC team) have learned the difference between dev, stage, and production, and calling this a dev environment feels more correct than feature, IMHO.

ccostino: Ah, I didn't realize that's all this is in reference to - thanks for clarifying!

ccostino:
Thanks again, @adborden!

PR #294, opened 2016-05-07
Build assets on CI.

jmcarp:
* Switch to vetted python buildpack
* Build assets on travis and include in deploy
* Remove `dist` from `.cfignore`

Part of https://github.com/18F/openFEC/issues/1643.

Note--since auto-deploy won't run from a pull request, verifying this requires merging into develop. Manual deployment can be verified by running `gulp build-js` and deploying as normal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,11 @@ before_deploy:
   - travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.15.0"
   - tar xzvf $HOME/cf.tgz -C $HOME
   - travis_retry cf install-plugin autopilot -f -r CF-Community
+  - (cd fec && gulp build-js)
 
 deploy:
   provider: script
+  skip_cleanup: true
   script: invoke deploy --branch $TRAVIS_BRANCH --yes
   on:
     all_branches: true
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -10,7 +10,6 @@ invoke notify
 
 # Build static files
 cd fec
-gulp build-js
 ./manage.py collectstatic --settings fec.settings.production --noinput
 ./manage.py compress --settings fec.settings.production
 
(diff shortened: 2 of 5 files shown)


ccostino:
@jmcarp:  Same here, are you ready for this to be tested/reviewed?  If so I'll take a look ASAP.

bin/run.sh
```diff
@@ -10,7 +10,6 @@ invoke notify
 
 # Build static files
 cd fec
-gulp build-js
 ./manage.py collectstatic --settings fec.settings.production --noinput
```
adborden: Doesn't the python buildpack take care of `collectstatic`? Can we remove this too?

jmcarp: It looks like it does--I'll set the django settings module in the manifest and drop this line, and we'll see what happens.

.travis.yml
```diff
@@ -35,9 +35,11 @@ before_deploy:
   - travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.15.0"
   - tar xzvf $HOME/cf.tgz -C $HOME
   - travis_retry cf install-plugin autopilot -f -r CF-Community
+  - (cd fec && gulp build-js)
```
ccostino: `npm run build` should now work in place of this, I had added the alias a while back for parity with the other repos.

ccostino:
Only the one comment here, otherwise I'm ready to merge this along with the API changes if you're also good to go.  Thanks, @jmcarp!

.travis.yml
ccostino: Actually, one other question here:  in the API change you have an `npm install` step but here there is not.  Is that due to it being taken care of in the `before_script` section?

adborden: npm install shouldn't be necessary, because it happens as part of the test setup. That's why the `skip_cleanup` is important, so the npm deps are ready and available to jump straight to `npm run build` :+1:

The API might not have js based tests, so the npm install happens at the end. (I haven't confirmed this for sure).

ccostino: Awesome, thanks!

PR #296, opened 2016-05-10
Changing class name to go with new svg stuff

noahmanger:
Changes the name of the close button class to use the new names from the Sass SVG work.


ccostino:
Merging this as well with the recent SVG icon work.

PR #300, opened 2016-05-18
Cms link update

emileighoutlaw:
We wrote the `party committee termination page` so that it could apply to *all* committees (i.e., be re-used). 

This PR redirects the candidate and SSF termination pages to the reusable one. Once it's live, I'll remove from the other pages from Wagtail. 

cc @ccostino 

This change isn't urgent, but I'd love to get it released as soon as possible. :)


ccostino:
Awesome, thanks @emileighoutlaw!

PR #310, opened 2016-05-24
Adding support for sticky-bar component

noahmanger:
Implements the new sticky bar component introduced [here](https://github.com/18F/fec-style/pull/359) and changes the classes used to initialize the sticky side component in order to keep the two distinct.


adborden:
:+1: Looks good, just needs a bump from fec-style I suppose.

noahmanger:
Ok. Fixed the conflicts here too.

PR #331, opened 2016-06-06
fec-style version fix

ccostino:
The newer version of `fec-style` was accidentally pushed to production when it was only intended for our next release test in staging.  This hotfix fixes the issue and reverts the version back to match what is currently deployed with the web app.

Note that this hotfix is not intended to be merge back into `develop`, so I propose we ignore git flow for this one and merge this directly into `master` ourselves and leave it at that.

/cc @noahmanger, @jenniferthibault, @LindsayYoung, @adborden, @jontours

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "ace-builds": "1.2.2",
     "aria-accordion": "0.1.0",
     "component-sticky": "1.0.0",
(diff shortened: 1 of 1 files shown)


ccostino:
Argh, wrong branch, closing and recreating.

PR #334, opened 2016-06-06
Fixes `R`

emileighoutlaw:
`Legal resources` should be lowercase `r`, in keeping with our sentence case style.

Content only. 

cc @porta-antiporta @adborden @edmullen 

Anyone can merge! :)

### DIFF
--- a/fec/legal/views.py
+++ b/fec/legal/views.py
@@ -4,6 +4,6 @@ def home(request):
     search_query = request.GET.get('query', None)
 
     return render(request, 'legal/home.html', {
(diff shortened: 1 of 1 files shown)


adborden:
Yeah, the term "Legal Resources" was in a handful of places in a handful of repos, but this occurrence looks like it hasn't been changed since was first created. Sometimes small fixes like this will get lost in merge conflicts. I just grep'd `fec-cms`, `fec-style`, and `openFEC-web-app` and it looks like this is the very last one :+1:

emileighoutlaw:
Thank you!

PR #344, opened 2016-06-14
Update CloudFoundry login setup

ccostino:
This changeset accounts for recent changes with our CloudFoundry setup and only logs you in if necessary now.

h/t @adborden for the same work in the web app.

/cc @LindsayYoung, @anthonygarvan, @adborden, @jontours, @noahmanger


LindsayYoung:
replaced by #345

PR #349, opened 2016-06-16
Feature/new heroes

noahmanger:
Adds the markup for the new heroes based on https://github.com/18F/fec-style/issues/370:

Legal resources:
image

Registration and reporting:
image

Depends on https://github.com/18F/fec-style/pull/392

cc @adborden

### DIFF
--- a/fec/fec/templates/partials/search-hero.html
+++ b/fec/fec/templates/partials/search-hero.html
@@ -1,24 +1,9 @@
 {% block content %}
-  <section class="hero hero--primary" aria-labelledby="hero-search">
-    <div class="container">
-      <div class="hero__heading">
-        <h1 id="hero-heading" class="hero__title t-display">{{ search_hero_heading | default:'Search betaFEC' }}</h1>
-        <h2 class="hero__subtitle t-ruled">Find the information you need</h2>
-      </div>
-      <form id="large-search" method="get" action="{{ settings.FEC_APP_URL }}/legal/search/" autocomplete="off" class="search__container js-search">
-        <div class="combo combo--search--large">
-          <select class="search__select select--alt-primary js-search-type" name="search_type" aria-label="Select a search type">
-            <option value="regulations">Regulations</option>
-          </select>
-          <input type="text" name="search" class="combo__input js-search-input"
-          aria-label="For example, &quot;102.2&quot; or &quot;electioneering&quot;"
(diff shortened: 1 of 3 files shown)


emileighoutlaw:
Those mockups look great! One thing on registration and reporting— with the new hero content, we no longer need the**What does it mean to register and report** or the paragraph below it.

noahmanger:
So like so? 
image

emileighoutlaw:
We don't need the two paragraphs in gray either. Sorry that I didn't hone in on this sooner! :(

noahmanger:
Oh! Are we removing that whole area, search box included?

emileighoutlaw:
Errr, I didn't think at all about what happens to the "Already registered" box, but yeah, we don't need any of the other content in that area.

noahmanger:
We could put it here? 
image

noahmanger:
Ah! Just noticed on @jenniferthibault 's mockups the box was replaced by a link in the same place. I think it's good to have the search field right there though. How about this? 

image

emileighoutlaw:
oohhhh! That looks awesome to me, but I defer to you and Jen on all things visual styling!

fec/fec/templates/partials/search-hero.html
```diff
@@ -1,24 +1,9 @@
...
-      </form>
-    </div>
-  </section>
+<section class="hero hero--primary hero--legal" aria-labelledby="hero-heading">
+  <div class="hero__image"></div>
+  <div class="container">
+    <h1 id="hero-heading" class="hero__title">Campaign finance data</h1>
+    <p class="hero__content">See how candidates and committees raise and spend money in federal elections. This financial data helps voters make informed decisions.</p>
+  </div>
+</section>
```
adborden: I would kill this file and move this content inline on the landing page. Originally, this partial was created to compliment the search partial, and eventually merge with that. But now that doesn't quite make sense. Eventually I would like to merge the legal search `form` markup with what's on the campaign finance side, but that won't happen until we tackle typeahead for legal anyway.

jenniferthibault:
I think this is the mockup you're talking about, Noah.

screen shot 2016-06-16 at 2 17 11 pm


I removed the search bar in order to make the feature smaller. Since the side nav is sticky & scrolls with the screen, if users are on short windows, it might be hard to access, or miss entirely. A quick link that should be visible even on short screens would bring people to the profile search.

noahmanger:
Got it.

PR #353, opened 2016-06-17
Build assets on CI and drop buildpack-multi

adborden:
We reverted the work to move the asset building onto CI. This branch contains the revert, thus re-instating that work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,11 @@ before_deploy:
   - travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.15.0"
   - tar xzvf $HOME/cf.tgz -C $HOME
   - travis_retry cf install-plugin autopilot -f -r CF-Community
+  - npm run build
 
 deploy:
   provider: script
+  skip_cleanup: true
   script: invoke deploy --branch $TRAVIS_BRANCH --yes
   on:
     all_branches: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
 
 env:
   global:
+    - DJANGO_SETTINGS_MODULE: fec.settings.production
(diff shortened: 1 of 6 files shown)


adborden:
https://github.com/18F/fec-cms/issues/352

noahmanger:
What does this need to get merged in?

adborden:
We'll have to bump python to 3.4.5 again https://github.com/18F/fec-cms/pull/430
Let's add this to the next sprint. I can rebase it and test it once more before merging.

.travis.yml
```diff
@@ -35,9 +36,13 @@ before_deploy:
   - travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.15.0"
   - tar xzvf $HOME/cf.tgz -C $HOME
   - travis_retry cf install-plugin autopilot -f -r CF-Community
+  - npm run build
+  - DJANGO_SETTINGS_MODULE='fec.settings.production' python fec/manage.py collectstatic --noinput
+  - DJANGO_SETTINGS_MODULE='fec.settings.production' python fec/manage.py compress
```
ccostino: Do we always want to use the `production` settings or should this also be based on the environment?

manifest_base.yml
```diff
@@ -1,10 +1,11 @@
 ---
 domain: 18f.gov
 stack: cflinuxfs2
-buildpack: "https://github.com/ddollar/heroku-buildpack-multi.git"
+buildpack: python_buildpack
 applications:
 - name: cms
 env:
+  DISABLE_COLLECTSTATIC: 1
+  DJANGO_SETTINGS_MODULE: fec.settings.production
```
ccostino: Same here, should we always just assume `production` settings?

.travis.yml
adborden: Good question, this is a little confusing, but these are the Django settings, so our choices are `dev` or `production`. I believe `dev` is really referring to local development, while `production` refers to anything running in cloud foundry, so I think `production` is right here. Another way to think about it, we only build/serve assets in `production` mode, so that's really our only option for deployment.

It brings up a good point about renaming these to something other than "environment". I've seen the pattern (maybe from rails, idk):
- APP_ENV: meaning **how** the app is configured i.e. `dev`, `stage`, `prod`, etc.
- APP_INSTANCE: meaning **where** the app is running. i.e. `dev`, `production` (or maybe `dev`, `cf` would make more sense).

Anyway, conversation for another day.

ccostino: Ah okay, this makes sense, thanks for explaining!  Yes, we can revisit another day but for now, this should be fine then. :-)

ccostino:
Thanks, @adborden!

PR #356, opened 2016-06-18
Fec style 2.5.0

adborden:


adborden:
I think we're already on 2.5.1.

PR #362, opened 2016-06-20
Target cloud.gov api for deploy

adborden:
When we're using `cf auth`, you need to specify which API to use before hand. Here's the [travis failure](https://travis-ci.org/18F/fec-cms/builds/139029719#L1504).

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -92,6 +92,10 @@ def deploy(space=None, branch=None, yes=False):
     if space is None:
         return
 
+    # Set api
+    api = 'https://api.cloud.gov'
+    run('cf api {0}'.format(api), echo=True)
+
     # Log in if necessary
     if os.getenv('FEC_CF_USERNAME') and os.getenv('FEC_CF_PASSWORD'):
         run('cf auth "$FEC_CF_USERNAME" "$FEC_CF_PASSWORD"', echo=True)


tasks.py
```diff
@@ -92,6 +92,10 @@ def deploy(space=None, branch=None, yes=False):
     if space is None:
         return
 
+    # Set api
+    api = 'https://api.cloud.gov'
+    run('cf api {0}'.format(api), echo=True)
+
     # Log in if necessary
     if os.getenv('FEC_CF_USERNAME') and os.getenv('FEC_CF_PASSWORD'):
         run('cf auth "$FEC_CF_USERNAME" "$FEC_CF_PASSWORD"', echo=True)
```
ccostino: In looking back on the code that we had changed, we originally had `--a https://api.cloud.gov` as an argument in the CF call.  Would it make sense to change this line back to that as opposed to running the `cf api` command separately?

adborden: Right, but don't forget the `cf login` only happens when the env variables are set, so we'd only set the api on travis and not when run locally. It's unlikely that when running locally, you'd be logged into a non-cloud.gov cloud foundry, but having this step separate makes things more consistent between running locally vs running on travis.

I also like that having the env variables set control only a very specific thing, which is auth'ing. vs before when we were using `cf login` to do many things (auth, set space, set org, set api) which are unrelated to the env variables being set. That's just my philosophy though.

ccostino: Ah, I see what you mean; that makes sense, thanks for clarifying!

ccostino:
Thanks again, @adborden!

PR #367, opened 2016-06-22
Add active nav classes based on content page

adborden:
The nav section is based on the wagtail `Page` model. I think this makes sense.

Is there a way to add `content_section` as a model property without making it a database field? I thought we could do something like `full_name` property [described here](https://docs.djangoproject.com/en/1.8/topics/db/models/#model-methods) but it's not exposed in the template.

### DIFF
--- a/fec/fec/templates/base.html
+++ b/fec/fec/templates/base.html
@@ -24,9 +24,8 @@
     {% endblock %}
   </head>
 
-  <body class="{% block body_class %}{% endblock %} {{ self.content_section }}">
+  <body class="{% block body_class %}{% endblock %}">
     {% if settings.FEC_CMS_ENVIRONMENT != 'PRODUCTION' %}
-    {{ page }}
     <div class="banner">
       <div class="container">
         <span class="t-block t-bold t-centered">{{ settings.FEC_CMS_ENVIRONMENT }}</span>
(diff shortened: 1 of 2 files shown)


adborden:
Resolves https://github.com/18F/fec-cms/issues/330

noahmanger:
This looks good to me, but I can't answer your question as to whether or not it's the right way to do it.

ccostino:
@adborden:  Yes, that is pretty much how you would go about doing that if you are using a traditional, database-backed model in Django.  There are abstract models that don't get tied to a database table, but in this case I'm pretty certain that's not an option.

Instead, what you do is define a method on a model and wrap it with the [`@property` decorator](https://docs.python.org/3/library/functions.html#property) decorator (the example in the Django docs is a more verbose way of doing this, particularly if you need to worry about both getters and setters).  For example, the `HomePage` class would look like this:

```python
class HomePage(ContentPage, UniqueModel):
    """Unique home page."""

    @property
    def content_section(self):
        return ''
```

And as another example, the `ContactPage` model would look like this:

```python
class ContactPage(ContentPage):
    @property
    def content_section(self):
        return 'contact'
```

And so on.

adborden:
@ccostino thanks, that is exactly what I would like, however when I tried the verbose way, as well as with the `@property` decorator, the field is not actually exposed in the template. I would love your help to understand why it doesn't work, if you have time. Here's a branch of the change `bug/active-site-nav-properties` https://github.com/18F/fec-cms/compare/bug/active-site-nav...bug/active-site-nav-properties

ccostino:
@adborden sure thing!  I will be able to take a look tomorrow morning and see if I can help.

adborden:
@ccostino updated, thanks for figuring this out!

ccostino:
You're welcome, thanks @adborden!

PR #372, opened 2016-06-23
Fix/accessibility fixes

noahmanger:
- Adds the proper `aria-describedby` and `id` properties for the learn more and download links on calendar events
- Wraps the filter + data-container elements to ensure correct sizing

Goes with https://github.com/18F/fec-style/pull/411


ccostino:
Thanks, @noahmanger!

PR #386, opened 2016-07-08
Replacing form with div

noahmanger:
Replaces `form` with `div`, per https://github.com/18F/Accessibility_Reviews/issues/15#issuecomment-228167208

Requires https://github.com/18F/fec-style/pull/424


xtine:
👍

PR #391, opened 2016-07-12
Moves release notes to right hand side

jenniferthibault:
Moves release notes to right-hand column card. 

screen shot 2016-07-12 at 11 05 34 am

Remaining task for @noahmanger in `fec-style` : vertically align text with center of star image


noahmanger:
Style PR coming with the star adjustment.

PR #398, opened 2016-07-15
Feature/upgrade with table

### LINKED ISSUE #395: Update Wagtail
We're currently using version 1.0, and we need to update to version 1.5.2.

Among other benefits, this improvement will help us improve the experience for content managers, which will get us closer to a place where we can train FEC folks to use and manage content in Wagtail. 

This is one blocker for getting FEC folks involved in the content migration.

## New features
Functionalities we'll use on this project when we update include:
- Editors can now access a list of previous page revisions and preview or roll back to any earlier revision from the page editing interface.
- TableBlock, for adding tables (http://docs.wagtail.io/en/v1.5.2/reference/contrib/table_block.html)
- The user experience around inserting, editing and removing links inside rich text areas has been greatly improved: Link destinations are shown as tooltips, and existing links can be edited as well as unlinked.
- The page editor interface now produces a warning if the user attempts to navigate away while there are unsaved changes.

LindsayYoung:
- Upgraded Django and Wagtail
 - Added tables (to custom page)

closes #395


LindsayYoung:
The tests are failing because, wagtail needs html5lib and that requires setuptools version 18.5 or above; and Travis is on 12.0.5

LindsayYoung:
@emileighoutlaw this now works- newer wagtail version and tables!

LindsayYoung:
fix  18F/openFEC-web-app#1370

emileighoutlaw:
Yay hooray!!

.travis.yml
```diff
@@ -12,8 +12,12 @@ env:
...
     - secure: "UYdDAAF93m3vvsfYKWyVhiEf0RKTtZriY4OM6fmRCDGCwVNJMnqu2eoHjzjaSc81UtFdHbFjwsBifr1CeMjTswJZGqIKnItkO5vd5/b8molIOUzJkieWA/A1mLb5mbF4XDr/s9hI2l3ChjRoXwD5S/SOnfQulmmOoSwq+Gl3O5g5D3j2HQjvMf9HIbRVBY1Ex/jSNy+unxRIUtIsqwhgzPjbEC7oQQG4lTZMwvYMnwLedig/69LAfD4RMfpCb5t+4lbaI/SJsgI1hTvrCfqjdNxrbm/ssT0pPFDInkxkoNTsi/CCaOkk+xawUKjxuG1DYsK1F2NgScxFOhVpFthKRW4lWxcqVNP8kGavDyUD9KgdqTprNr15okSnlV6ziq/jFn06/X8Zue/WK172hxjBtkZpHdH46oCEBtNcpTUKNliqRmnLUeM+dXcyZCMVmMEqrPdIZ5M8b0coce5nimLKWZjzr5DGcjS0QtpYfgBGenwXLojVFmlFbREf2VqU+4E6G1OI2YBv3a0F6KoHDtdbDz/l6DLwyHrUDF+1gWqyyZD4tqGAKFjQ7G90jofaXF05mx1L/74B5NmaeC2c30vgDR5q6bMHlhB5BpUoqlzVRNj1SYi3TPerSgtTGTimFjwk+j1JsqydrDijUu8+NiFVxAc+JAT5IH4Ooy6+JD0Yzvg=" # FEC_CF_PASSWORD
 
+before_install:
+  - "pip install --upgrade pip setuptools"
```
ccostino: Nice way to get the latest versions in!  We should probably do the same in the other projects.

.travis.yml
```diff
@@ -12,8 +12,12 @@ env:
...
+
 before_script:
   - travis_retry pip install -U pip wheel
+  - pip install -U setuptools
```
ccostino: This needs to be done twice?

LindsayYoung: good catch! I'll take this out

.travis.yml
LindsayYoung: that is a good idea

.travis.yml
ccostino: It may very well be necessary (and if so, I just realized it should be prefaced with a `travis_retry`) - I think I've seen instances where depending on what step Travis was on it needed to repeat something.  That said, if it is needed, `setuptools` could also just be added to the line above. :-)

LindsayYoung:
@ccostino this PR is ready when you are :)

ccostino:
Thanks @LindsayYoung!

PR #401, opened 2016-07-19
Adding All, AO as legal search options

porta-antiporta:


porta-antiporta:
This is to address: https://github.com/18F/fec-eregs/issues/187

porta-antiporta:
Intended changes to be made:
* Add options for `All, Advisory Opinions`
* Retask `features.legal` to control those drop down values rather than the legal_resources section itself
* Disable features.legal in prod (as we don't want advisory opinions to show up in prod)

adborden:
Paired with @porta-antiporta on this one. We moved the feature flag to the user provided service `fec-creds-*`.  This has the following advantages:
- Flipping the feature flag no longer requires a code change.
- All apps read from the same variable, so no more editing manifests in each app to control a feature.

Picking up the new flag _does_ still require a `restage`, but at least that can be done with `cf zero-downtime-push`.

Note, to unset a feature flag, it should be set to the empty string. Setting it to `'0'` will parse as `True`.

adborden:
@anthonygarvan we already tested this locally and cloud.gov, but would still like another pair of eyes, do you mind doing a quick review?

anthonygarvan:
Looks good! Congrats @porta-antiporta , we'll start expecting more PRs from you 🎉

porta-antiporta:
:tada: :musical_note: :two::ear:

PR #402, opened 2016-07-19
Adding loading state to calendar filters

noahmanger:
Adds the loading / success states to calendar filters.

Requires https://github.com/18F/fec-style/pull/435

### DIFF
--- a/fec/fec/static/js/calendar.js
+++ b/fec/fec/static/js/calendar.js
@@ -146,6 +146,15 @@ Calendar.prototype.filter = function() {
 
 Calendar.prototype.success = function(response) {
   var self = this;
+
+  setTimeout(function() {
+    $('.is-loading').removeClass('is-loading').addClass('is-successful');
+  }, 500);
+
+  setTimeout(function() {
+    $('.is-successful').removeClass('is-successful');
+  }, 2000);
+
   return response.results.map(function(event) {
     var processed = {
       category: event.category,


fec/fec/static/js/calendar.js
```diff
@@ -146,6 +146,15 @@ Calendar.prototype.filter = function() {
 
 Calendar.prototype.success = function(response) {
   var self = this;
+
+  setTimeout(function() {
+    $('.is-loading').removeClass('is-loading').addClass('is-successful');
+  }, 500);
+
+  setTimeout(function() {
```
xtine: At first I feel kind of iffy that the second `setTimeout` relies on the first one to finish, but not sure if nesting setTimeouts is that much better. However I think setTimeouts are ever stopped without explicitly doing so.

noahmanger: What would you recommend?

noahmanger:
Just noting that this will also need to be merged into the release branch: https://github.com/18F/fec-cms/pull/421

ccostino:
@noahmanger and @xtine:  Just double checking that this is ready to go - it looks fine to me (and noted about the `setTimeout` chaining, but I've had to do similar things in the past), but just wanted to confirm.

fec/fec/static/js/calendar.js
xtine: It was just a general note, I don't think I have anything better to suggest. The rest looks good to me! 👍

xtine:
👍

PR #423, opened 2016-08-02
Updating to include advisory opinion content

porta-antiporta:


porta-antiporta:
Addresses the update to `legal-resources` for: https://github.com/18F/fec-eregs/issues/196

emileighoutlaw:
I've had some time to ponder this today, and I believe the buttons should all follow the same format: `Explore xxx`. Further, I think the `xxx` should align with the section title, which will be important when we have several sections, a lá:

> **Advisory opinions**
_Intro text_
[ Explore advisory opinions ] 


> **Regulations**
_Intro text_
[ Explore regulations ] 

> **Statutes**
_Intro text_
> [ Explore statutes ]

So I just made that change as a commit here. Feel free to 👋 at me and let me know if you think that's wrong @porta-antiporta

PR #424, opened 2016-08-02
Rearranging calendar page

noahmanger:
Supports the new designs and styles from https://github.com/18F/fec-style/pull/446

Unfortunately, because a lot of the title area and controls are generated by the FullCalendar library we're using, I had to use some gunky jQuery to move elements around and get them into their proper place. (Not much new from the way this code has been working, though).

image

cc @jenniferthibault @xtine

### DIFF
--- a/fec/fec/static/js/calendar.js
+++ b/fec/fec/static/js/calendar.js
@@ -42,6 +42,7 @@ function Calendar(opts) {
   this.opts = $.extend({}, this.defaultOpts(), opts);
 
   this.$calendar = $(this.opts.selector);
+  this.$head = $('.data-container__head');
   this.$calendar.fullCalendar(this.opts.calendarOpts);
   this.url = URI(this.opts.url);
   this.exportUrl = URI(this.opts.exportUrl);
@@ -56,7 +57,6 @@ function Calendar(opts) {
 
   this.$download = $(opts.download);
   this.$subscribe = $(opts.subscribe);
-  this.$toolbar = this.$calendar.find('.fc-toolbar');
 
   this.$calendar.on('click', '.js-toggle-view', this.toggleListView.bind(this));
 
@@ -71,11 +71,8 @@ function Calendar(opts) {
   this.filter();
   this.styleButtons();
 
-  this.$toolbar.after($('.js-data-widgets'));
-  this.$toolbar.append($('.js-calendar-action'));
-
   if (!helpers.isLargeScreen()) {
-    this.$toolbar.after($('#filters'));
+    this.$head.after($('#filters'));
   }
 }
 
@@ -89,7 +86,7 @@ Calendar.prototype.defaultOpts = function() {
     calendarOpts: {
       header: {
         left: 'prev,next,today',
-        center: 'title',
+        center: '',
         right: 'monthTime,month'
       },
       buttonIcons: false,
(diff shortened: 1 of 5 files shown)


jenniferthibault:
@noahmanger I added some general comments to the style PR, but here is one specific to the calendar implementation:

In the default loading state, should this text say "Viewing all events" ?
screen shot 2016-08-04 at 10 55 01 am

fec/fec/static/js/calendar.js
```diff
@@ -69,6 +69,13 @@ function Calendar(opts) {
 
   this.filter();
   this.styleButtons();
+
+  this.$calendar.find('.fc-toolbar').after($('.js-data-widgets'));
+  this.$calendar.find('.fc-toolbar').append($('.js-calendar-action'));
```
adborden: At this point, I would stick this in a local variable like `$toolbar` just for readability.

fec/fec/static/js/calendar.js
```diff
@@ -214,6 +221,19 @@ Calendar.prototype.handleRender = function(view) {
     this.$listToggles = null;
   }
   this.$calendar.find('.fc-more').attr({'tabindex': '0', 'aria-describedby': this.popoverId});
+
+  // Replace the h2 with an h1
+  this.$calendar.find('.fc-center h2').replaceWith(function(){
+    return $('<h1>', {html: $(this).html(), class: 'data-container__title'});
+  });
```
adborden: Hmm, I wonder if this would be better handled by just adding a class that implements the `h1` styles.

noahmanger: Not a bad idea. But part of this is that the page doesn't actually have an `<h1>` now, which it should.

adborden:
Yeah, it's a little weird, but that's the API they expose...

fec/fec/static/js/calendar.js
adborden: Ugg. Yeah, that makes more sense then.

Another thought, have you considered using CSS for some of this? Another
approach would be to hide the `h2`, and have your own `h1` on the page,
outside of the calendar. Or even have the fc's `h2` include the FEC `h1`
mixin. For other elements like the buttons, you could `@extends` the fc
classes with the button classes.

I wonder how maintainable the two approaches are. Testing the CSS
approach would involve browser testing, while the JS can be unit
tested.


Quoting Noah Manger (2016-08-10 12:55:56)
>> @@ -214,6 +221,19 @@ Calendar.prototype.handleRender = function(view) {
>>      this.$listToggles = null;
>>    }
>>    this.$calendar.find('.fc-more').attr({'tabindex': '0', 'aria-describedby': this.popoverId});
>> +
>> +  // Replace the h2 with an h1
>> +  this.$calendar.find('.fc-center h2').replaceWith(function(){
>> +    return $('<h1>', {html: $(this).html(), class: 'data-container__title'});
>> +  });
> 
> Not a bad idea. But part of this is that the page doesn't actually have an `<h1>` now, which it should.
> 
> ---
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/18F/fec-cms/pull/424/files/39386bc698d171bdb5d330172099dece6433fc67#r74317962

--
Aaron D Borden
Human and Hacker

noahmanger:
Ok, I think I addressed everything: 
- I made it so that it now says `Viewing all events` when no tags are applied
- I cleaned up some of the messy DOM manipulation, including just providing our own `h1` and setting the text based on the title of the calendar view when it renders

Requires https://github.com/18F/fec-style/pull/464

adborden:
Nice, much cleaner!

PR #442, opened 2016-08-19
Add statutes to legal search

adborden:


adborden:
This is paired with https://github.com/18F/openFEC-web-app/pull/1531

fec/legal/templates/legal/home.html
```diff
@@ -11,13 +11,12 @@
       <form id="large-search" method="get" action="{{ settings.FEC_APP_URL }}/legal/search/" autocomplete="off" class="search__container js-search">
         <div class="combo combo--search combo--search--large">
           <select class="search__select select--alt js-search-type" name="search_type" aria-label="Select a search type">
-          {% if features.legal %}
             <option value="all">All</option>
+            {% if features.legal_statutes %}
+            <option value="statutes">Statutes</option>
+            {% endif %}
```
porta-antiporta: This may be a moot point if we wind up shipping Statutes at the end of this sprint.  But as long as we have the statutes feature flag, can we add an `else` condition that points to a with a "Coming soon.  Statutes are available on [FEC.gov](http://www.fec.gov/law/feca/feca.shtml#statutes)." message.  Similar to Enforcement Matters.

adborden: Sure thing. This is only the `select` option to put statutes in the
search field. I've added the feature-off content to the statute landing
page story https://github.com/18F/fec-eregs/issues/208


Quoting Porta Antiporta (2016-08-19 14:58:09)
> >              <option value="all">All</option>
> > +            {% if features.legal_statutes %}
> > +            <option value="statutes">Statutes</option>
> > +            {% endif %}
> 
> This may be a moot point if we wind up shipping Statutes at the end of this sprint.  But as long as we have the statutes feature flag, can we add an `else` condition that points to a with a "Coming soon.  Statutes are available on [FEC.gov](http://www.fec.gov/law/feca/feca.shtml#statutes)." message.  Similar to Enforcement Matters.
> 
> 
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/18F/fec-cms/pull/442/files/49880fe725ab2759209443f61390c6a0df029886#r75557457

--
Aaron D Borden
Human and Hacker

porta-antiporta: 👍

porta-antiporta:
:shipit:

PR #447, opened 2016-08-23
Fix migrations and deprecation warnings

ccostino:
This changeset resolves a couple of migration conflicts that came up recently and also addresses a couple of Django deprecation warnings we started receiving with the recent Django upgrade.

/cc @LindsayYoung, @anthonygarvan


ccostino:
Note that this does not fix the template deprecation warning that still appears - that can come after the micropurchase!

PR #450, opened 2016-08-26
Clarifies language

emileighoutlaw:
I think we should be consistent in our naming of statutes. 

By that I mean either we refer to them as `Title 26 and Title 52` or `26 USC and 52 USC`. That'll be less for readers to keep in mind when trying to understand statutes. 

I'm going with `Title 26 and Title 52`. 

Content only. Should be a fast merge!

cc @adborden @porta-antiporta


porta-antiporta:
LGTM :shipit:

adborden:
@emileighoutlaw we also use the USC language in the search results, should we make the same change there as well?
screenshot from 2016-08-30 17-11-10

porta-antiporta:
@adborden we should keep it as is given: https://18f.slack.com/archives/fec-partners/p1467229654000403

_though, perhaps it should be `U.S.C.` instead of USC.  Submitting PR for that change.

PR #476, opened 2016-09-02
Adding printer-only logo

noahmanger:
Adds the print logo from https://github.com/18F/fec-style/pull/496

Because certain browsers don't render background images it has to be included as an `img`.

### DIFF
--- a/fec/fec/templates/base.html
+++ b/fec/fec/templates/base.html
@@ -38,6 +38,7 @@
           <span class="disclaimer__left">This site is in beta, visit <a href="http://www.fec.gov">FEC.gov</a></span>
           <span class="disclaimer__right">An official website of the United States Government <img src="{% static 'img/us_flag_small.png' %}"  alt="US flag signifying that this is a United States Federal Government website"></span>
         </div>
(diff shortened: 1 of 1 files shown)


fec/fec/templates/base.html
```diff
@@ -38,6 +38,7 @@
           <span class="disclaimer__left">This site is in beta, visit <a href="http://www.fec.gov">FEC.gov</a></span>
           <span class="disclaimer__right">An official website of the United States Government <img src="{% static 'img/us_flag_small.png' %}"  alt="US flag signifying that this is a United States Federal Government website"></span>
         </div>
+        <img src="{% static 'img/print-logo.png' %}" class="u-print-only" alt="FEC logo">
```
ccostino: Should the alt text be something like `Print` instead (for accessibility improvement)?

ccostino: Should the alt text be something like `Print` instead (for accessibility improvement)?

ccostino: Whoops, never mind this, it's the print version of the FEC logo, not a print icon/button!

ccostino:
Thanks, @noahmanger!

PR #477, opened 2016-09-03
Feature/checklist-template

noahmanger:
This creates a new page model and template called `CollectionPage`, which is used for all of the essentials checklists as well as the new press page (and likely others in the future). 

The admin interface exposes lots of fields that are used to set the related pages in the sidebar, each of the list items, and the style of the list (checklist vs bullets).

image

So it can generate pages like:
image

And:
image

Note: I wasn't sure the protocol on how to deal with migrations, so I checked them all in, but maybe that's not necessary?


fec/home/models.py
```diff
@@ -239,3 +239,38 @@ class CustomPage(Page):
...
         StreamFieldPanel('sidebar'),
     ]
+
+class CollectionList(blocks.StructBlock):
+    title = blocks.CharBlock(required=True)
+    style = blocks.ChoiceBlock(default='list--bulleted',
+        choices=[
+            ('list--checks list--checks--secondary', 'Checklist'),
+            ('list--bulleted', 'Bulleted list')
+        ])
```
adborden: Eh, this feels a bit funny to me, storing class names in the DB when a single char will do. It does mean a little more work in the frontend when you're going to switch based on the choice, but I think that's the more standard approach.

noahmanger: Ack, yeah.🙈  That totally makes sense.

adborden:
Excited to see wagtail getting some love! Migrations ran smoothly too.

noahmanger:
Yeah! Finally took the time to learn up on it and it's pretty slick. First foray into Django-land for me.

fec/home/models.py
```diff
@@ -239,3 +239,38 @@ class CustomPage(Page):
...
         StreamFieldPanel('sidebar'),
     ]
+
+class CollectionList(blocks.StructBlock):
+    title = blocks.CharBlock(required=True)
+    style = blocks.ChoiceBlock(default='bullet',
+        choices=[
+            ('check', 'Checklist'),
+            ('bullet', 'Bulleted list')
+        ])
```
ccostino: Minor nitpick here with the `choices` - these usually get defined as separate items and then set here, e.g., like the example in the Django docs:  https://docs.djangoproject.com/en/1.10/ref/models/fields/#choices

So in this case, it'd look like this here:

```python
class CollectionList(blocks.StructBlock):
    CHECKLIST = 'check'
    BULLET = 'bullet'
    STYLE_CHOICES = (
        (CHECKLIST, 'Checklist'),
        (BULLET, 'Bullet'),
    )
    title = blocks.CharBlock(required=True)
    style = blocks.ChoiceBlock(default='bullet', choices=STYLE_CHOICES)
    intro = blocks.RichTextBlock(blank=False, null=False, required=False)
    items = blocks.ListBlock(blocks.RichTextBlock(classname="nothing"))
```

Given that this is such a small thing here though (and the choice list below as well), I'm not inclined to hold things up but did want to point it out for future reference!  The main reason is that these choice lists tend to change more often than not and it's easier to manage them in this fashion (and keep the data related to the model it's being used for).

noahmanger: Thanks for the tip! Just updated. Definitely appreciate tips about Django conventions, which I'm not familiar with at all.

Issue #474, opened 2016-09-02
Implement "latest updates" page


noahmanger:

emileighoutlaw:
👋 . I got interrupted when we opened this issue in sprint check-in. Maybe @xtine can let me know what content things I can help with here?

noahmanger:
Resolved by #478

PR #478, opened 2016-09-04
Latest Updates landing page

xtine:
## Latest Updates
This sets up a simple view to show Press Releases, News Digests, and Records on one landing page for #474.


emileighoutlaw:
I pushed some capitalization changes here! Very small! Not trying to be a grammar bully :hurtrealbad: ❤️

noahmanger:
Nice! This is a great start. I can take it from here.

noahmanger:
I hit a stride, so I just kept running with this. This PR now includes the new press landing page, complete with press release feed:

image

And all the makings of a filterable feed of all the pages (placeholder design for now):

image

noahmanger:
Ok, I have the visual designs from https://github.com/18F/fec-cms/issues/454 implemented, with a couple suggested variations purely for the sake of getting an MVP version out the door.

The feed of items is as designed (minus the "read more" links in the blurbs, which would take more work, but can happen later). 
image

- Press release and digest type updates say "From: Press office" and FEC Record type updates are "From: Information division"
- Links go to a filtered view for that update type
- Titles link to pages
- All posts are sorted most recent first

Where I'm diverging from the mocks is in the filters:
image
- For the sake of simplicity, I wanted to do this with as little javascript as possible. This means two things:
  - Using native form inputs rather than our special dropdowns (which would need to be updated to support slightly different behavior)
  - Relying on page reloads to show updated data, rather than doing an asynchronous request to get results, which would take more time to set up. So the page reloads after every form change. 
- Because regular html `select` elements can't support multiple values, I use checkboxes here
- By default, all checkboxes are empty and the "subjects" select is disabled
- Checking "press releases" reloads the page, showing only press releases and enables a "press release subjects checkbox:
image
- Unselecting it disables the checkbox again
- The same works for "FEC Record":
image

- If both "FEC Record" and "Press releases" are checked, then two selects are shown, one for each type:
image
- Filtering by year works as expected

I'd like to start getting a code review on this now that it just _works_ because it's pretty big. We can discuss interaction specifics here though. I'm totally open to changes, but would like to try to minimize dev time.

onezerojeremy:
Probably I'm reading this wrong, and I don't reallllly know what i'm talking about here, but I don't think we need to support selection of multiple values in the content type selection. Users can pick: 
- for media pros : reloads page with a new query string that has `&type=weekly_digests,news_releases`
 - weekly digest: reloads page with a new query string that has  `&type=weekly_digests`
 - news releases : reloads page with a new query string that has `&type=news_releases`
- for committees: reloads the page with a new query string that has `&type=record`
 - the record: reloads the page with a new query string that has `&type=record`

noahmanger:
I see what you're saying, and your technical rec is _almost_ right, except:
- When grouping items in a select using `<optgroup>`, the label itself isn't clickable:
image
- You can't have `<option>`s in a `<select>` with multiple values

BUT: Your solution is totally feasible and not that hard to engineer as a next iteration, it just requires more custom javascript.

onezerojeremy:
ah. Thanks for explaining! I figured there was something I was missing (and forgetting). Really appreciate it!

emileighoutlaw:
This is looking great! 

I made some style changes and added some content, so you don't have to work in lorem ipsum on the `Latest updates` page.

There's still some content needed for the Press landing page, but I want to work with Judy to develop it— she's a fantastic writer, and I want her to feel excited about this. I'll ask Amy what the best path foward is tomorrow a.m.

jenniferthibault:
+1 to please no checkboxes. 

@onezerojeremy / @nickykrause Are we intending to have validation happening with these dropdown & year entry fields?

cc @nickykrause just for visibility (I only found this issue by chance when @emileighoutlaw mentioned she put a content note in here, and there are interaction things being discussed.)

noahmanger:
So, not doing checkboxes means that only one option can be selected at time. But I did figure out a way to have options (like "for media professionals") that the server can use to render multiple update types:

image

So selecting "for media professionals" returns a list of press releases and digests, and shows the press release category. Selecting "For committees" or "FEC Record" does the same thing.

onezerojeremy:
that looks great to me! Thanks for the fancy footwork, @noahmanger !

noahmanger:
@jenniferthibault ? @nickykrause ? what do you think?

nickykrause:
@noahmanger 

This looks good to me, although I do wonder whether or not users will realize that "For media professionals:" is selectable (rather than just a label). But, either way, I do like what you've done with that option to enable users to select both press releases and digests, if they desire.

Seems good for now, at least, I think!

jenniferthibault:
+1 to what the smart people above said 😺

fec/home/models.py
```diff
@@ -172,6 +170,10 @@ class RecordPage(ContentPage):
     def content_section(self):
         return ''
 
+    @property
+    def get_update_type(self):
+        return 'FEC Record'
```
ccostino: Should this be referring to the `update_types` `OrderedDict` defined above instead?

fec/home/models.py
```diff
@@ -197,6 +199,10 @@ class DigestPage(ContentPage):
     def content_section(self):
         return ''
 
+    @property
+    def get_update_type(self):
+        return 'Weekly Digest'
```
ccostino: Same question here.

fec/home/models.py
```diff
@@ -225,6 +231,10 @@ class PressReleasePage(ContentPage):
     def content_section(self):
         return ''
 
+    @property
+    def get_update_type(self):
+        return 'Press release'
```
ccostino: And here.

fec/home/templatetags/filters.py
```diff
@@ -7,3 +7,7 @@
 @register.filter
 def clean_whitespace(value):
     return re.sub(r'\s+', '-', value)
+
+@register.filter
+def lookup(dict, arg):
```
ccostino: Is this custom filter being used anywhere in the app?

noahmanger: Is there a way to get a dictionary value from a key in a template? I looked and couldn't find anything but maybe I missed it. Seemed weird to me that this wasn't built-in.

fec/home/models.py
noahmanger: Good call. Fixed!

fec/home/templatetags/filters.py
ccostino: D'oh, apparently you're right, sorry about that!  Okay, so with this being setup as a filter, which means you'd use it like `{{ my_dict|lookup:arg }}`.  That should still be a `dict.get(arg, '')` (or some other useful default value supplied), though, otherwise if a `KeyError` is thrown it won't be handled gracefully.

noahmanger:
Updated this to be against the release branch.

ccostino:
Awesome, thanks @noahmanger!

PR #483, opened 2016-09-09
[WIP] Feature/load content

LindsayYoung:
This works locally, I am going to try it out on staging and probably add some improvements.


noahmanger:
Just curious: why do these need to be checked in?

noahmanger:
I'm trying to run this locally and finding that it's breaking my admin interface. I can log in but once I try to navigate to a subpage, I get an error. Which I think is the same thing happening on staging that's resulting to getting kicked to a 404 error.

Any idea what's up?

PR #489, opened 2016-09-13
Release fixes

noahmanger:
This makes  a few fixes to the latest updates page in preparation for the release:

1. It only renders the latest updates view if either the user is authenticated or a feature flag is set. This will allow it to exist on production and be visible to authenticated users, and also on staging / dev / local and be visible to everyone. This page is just created in the code, not the DB, so I had to hide with some mechanism until they're ready.

Two questions (maybe fore @adborden or @ccostino ):
- Am I doing this right? cc @adborden 
- Where do the flags get set for the different environments? 

2. Adds the "updates preview" title to the press page above the feed
3. Removes the "FEC Record" from the filter options
4. Programmatically generates breadcrumb links on the `CollectionPage` template, which is used for the 'Resources for media professionals' page

### DIFF
--- a/fec/home/views.py
+++ b/fec/home/views.py
@@ -1,5 +1,6 @@
 from django.shortcuts import render
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
+from django.conf import settings
 from itertools import chain
 from operator import attrgetter
 from home.models import DigestPage
@@ -37,6 +38,8 @@ def get_press_releases(category_list=False, year=False):
   return press_releases
 
 def updates(request):
+  # Only render view if the user is authenticated or there's a feature flag
+  if request.user.is_authenticated() or settings.FEATURES['latest_updates']:
     digests = ''
     records = ''
     press_releases = ''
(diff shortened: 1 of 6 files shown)


fec/home/views.py
```diff
@@ -37,6 +38,8 @@ def get_press_releases(category_list=False, year=False):
   return press_releases
 
 def updates(request):
+  # Only render view if the user is authenticated or there's a feature flag
+  if request.user.is_authenticated() or settings.FEATURES['latest_updates']:
```
adborden: Not a huge deal, but I prefer the "fail fast" style functions because they're easier to read. Avoids having a chunk of indented code in if/else blocks.
```python
def updates(request):
  if not (request.user.is_authenticated() or settings.FEATURES['latest_updates']):
    return render(request, '404.html')

  # render code here
```

noahmanger: Ah totally right. Will fix.

adborden:
:+1: To enable the feature flag, you'll want to [update the credential service](https://github.com/18F/openFEC/wiki/switch-out-cf-service-environment-variables) for `feature`, `dev`, and `stage`.

adborden:
I wish I could emoji that commit :smile:

ccostino:
Awesome, thanks @noahmanger!

PR #500, opened 2016-09-16
Feature/improve update display

noahmanger:
Replaces the breadcrumbs on CustomPage templates with a new partial that programmatically generates breadcrumb links based on ancestor pages. If a page's ancestor is "Registration and reporting" it's an "Essentials" page and so "Essentials" is added as an inactive item. This allows press content pages to show the correct breadcrumbs:
image

Cleans up press release and digest pages by:
- Adding the correct class to the tag
- Adding "Latest updates: " to the breadcrumbs
- Activating the links in the footer to the index page

And last, restricts the feed on the main press page to just the last 5 press releases and last 2 weekly digests (these numbers can be tweaked of course).

Resolves https://github.com/18F/fec-cms/issues/495

cc @LindsayYoung @emileighoutlaw

### DIFF
--- a/fec/fec/templates/partials/body-blocks.html
+++ b/fec/fec/templates/partials/body-blocks.html
@@ -16,7 +16,7 @@
 
 {% with blocks_name=blocks_name|default:"body" %}
 
-  <div class="{{ blocks_name }}-blocks">
+  <div class="{{ blocks_name }}-blocks content__section content__section--narrow">
     {% for block in blocks %}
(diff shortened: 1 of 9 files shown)


emileighoutlaw:
Content looks perfecto on this one. Thanks!

LindsayYoung:
The code looks fine to me and the one page I have locally where can check this behavior looks like it is taking the inheritance very literally. So locally, I get a home and a home page- I think this will work fine on production, I just checked and the hierarchy is not as nested as my local.

PR #502, opened 2016-09-19
Cleaning up the wagtail editor interface

noahmanger:
This includes a few changes to improve the wagtail editing experience:
- Removes the "undo" / "redo" buttons, which aren't necessary and take up space
- Forces the rich text menu bar to be full width when possible, rather than breaking onto two lines
- Removes extraneous padding on rich text fields
- Enables the Halo.js plugin to clean HTML when pasted into the editor. This required adding Rangy and jQuery HTMLClean as vendor assets

Before:
image

After:
image

More of these WYSIWIG controls can be removed as well.

cc @adborden for review, @emileighoutlaw for visibility.


fec/fec/static/js/customize-editor.js
```diff
@@ -0,0 +1,11 @@
...
+    (function($) {
+      // Force hallotoolbar to be full width when possible
+      $(document.body).on('halloactivated', function() {
+        var $toolbar = $('.hallotoolbar');
+        $toolbar.width('auto');
+        // Remove re-undo buttons
+        $toolbar.find('.halloreundo.ui-buttonset').remove();
+      });
+    })(jQuery);
+}).call(this);
```
adborden: I've never seen this, what's the difference between using `call` and just `(function() {})(this)` ?

adborden: oh sorry, feel like I just woke up. re-phrase, why bind use the outer function at all?

noahmanger: Good catch. I don't think it was necessary. It was in the example code I started from. Just reverted to the more traditional form.

adborden:
I haven't dug too far into this, but why the need for htmlclean? Does Wagtail not sanitize the HTML the way we want? Judging from this https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/whitelist.py they encourage you to subclass the whitelister to tweak the behavior.

noahmanger:
So, I saw that, and we're even doing a version of it https://github.com/18F/fec-cms/blob/feature/customize-editor/fec/fec/wagtail_hooks.py#L6 . But I wasn't clear on if it's possible to have it strip attributes as well?

noahmanger:
Ok, after some experimenting that does look like a better approach. I replaced the old code with a verbose list that includes the `allow_without_attributes` necessary to strip them.

Unless you think otherwise, I think it's helpful to also include the client-side cleaning so that you don't see the extra cruft in the editor when pasting, but it does add some extra scripts to the page.

emileighoutlaw:
This is terribly exciting. That is all. ❤️

adborden:
Sweeet.

PR #509, opened 2016-09-21
Feature/reconcile h1s

noahmanger:
Adjusts classes and markup throughout to accommodate new header styles from https://github.com/18F/fec-style/pull/520

## Examples:
Custom page:
image


jenniferthibault:
Few checks:

I think the Contact page is missing the new styles: it should be underlined

screen shot 2016-10-04 at 3 38 55 pm


And should we apply this class system on the homepage?

screen shot 2016-10-04 at 5 08 28 pm

noahmanger:
Ok, fixed it for contact. I don't think we want to use it for "New features" though. It seems different to me.

noahmanger:
Ok, this is all ready.

PR #511, opened 2016-09-22
Pinning to new minor update of glossary

noahmanger:
Makes use of the new version of the glossary panel that sets focus on the element that opened the glossary.

cc @xtine


xtine:
👍

PR #512, opened 2016-09-22
Feature/adding watch

noahmanger:
This uses [nswatch](https://github.com/egoist/nswatch) to watch for changes to js and scss files and rebuilds them automatically. 

It also includes the proper instructions for using `npm link` for including fec-style, and configures `watch` so that it watches for changes on the linked fec-style files. 

And last, it replaces REAMDE.rst with README.md so all our repos have the same format.


adborden:
:tada: have you tried adding the handlebar templates as well? I would think you want those to be watched too.

noahmanger:
Ooh I didn't think of that. Good call.

PR #516, opened 2016-09-23
Feature/remove calendar template

noahmanger:
This hard-codes the contact and calendar pages rather than relying on them to be populated on the CMS, which makes it slightly easier to standup instances of the CMS.

It also adds the `content_section` property to the `legal_view.home` so that the parent nav item is highlighted on the landing page. 

cc @adborden @xtine


adborden:
This fixes the cms part of https://github.com/18F/fec-eregs/issues/170

noahmanger:
Cool cool. I thought we decided to check in migrations? @adborden and @ccostino ?

adborden:
Yeah, you definitely want to keep the migrations, otherwise the CF databases won't get updated.

PR #530, opened 2016-09-29
Changes verb to "search"

emileighoutlaw:
Our pattern on the site so far has been to use "search" as a verb and "filter" as a noun.

This is a one-word PR, so I'm thinking maybe @jenniferthibault would be a good 👀 ?


jenniferthibault:
Dig it 😻

PR #533, opened 2016-09-29
[MERGE WITH GITFLOW] Removing then re-adding feed_intro field

noahmanger:
This removes then re-adds the `feed_intro` field on the `presslandingpage` model in order to resolve an issue causing the admin panel to crash at certain pages.


LindsayYoung:
:+1:

PR #536, opened 2016-10-04
Adding polyfill for WeakMap

noahmanger:
Adds a polyfill for `WeakMap` to patch a bug in IE 10 caused by an included library that used `WeakMap`. 

Resolves https://github.com/18F/openFEC-web-app/issues/1194


ccostino:
Thanks, @noahmanger!

PR #543, opened 2016-10-07
Updates content

emileighoutlaw:
This changes the content per FEC feedback:

> Also, wondering about "can" as in "who can register and report?" It is a legal requirement so I'm tempted to suggest "must" even though that's not the friendliest of greetings. Does suggesting an optional choice stir resentment? If the IRS said "Who can pay taxes?" I'd growl.

Resolves: https://github.com/18F/FEC/issues/489#issuecomment-252088579


emileighoutlaw:
Content only. Anyone can review

PR #547, opened 2016-10-11
Makes digest plural

emileighoutlaw:
"Press releases" is plural on the Press landing page, so "Weekly Digests" should be plural to match.

This PR makes them parallel and plural.


noahmanger:
Changed this to go into the release.

emileighoutlaw:
Ah, thank you.

PR #571, opened 2016-10-28
Search consistency

emileighoutlaw:
Removes placeholder text and adds an example. I'm not feeling 100% confident I did this HTML right, so 

cc @noahmanger and @jenniferthibault


emileighoutlaw:
Ugh — no idea why this failed :(

PR #584, opened 2016-11-07
Adding hidden press boilerplate to 2016 press releases

noahmanger:
Adds the press release boilerplate to the press release template and adds a `@property` to the `press_release_page` model to flag when a release is from 2016 or later, and then adds the `.u-print-only` class so that it will only be visible when print.

For earlier press releases, the boilerplate will stay in the generated content and will continue to be visible online as well as print.

Resolves https://github.com/18F/fec-cms/issues/580


fec/home/models.py
```diff
@@ -223,6 +223,15 @@ def content_section(self):
...
 
+    """
+    Because we removed the boilerplate from all 2016 releases
+    this flag is used to show it in the templates as a print-only element
+    """
+    @property
+    def no_boilerplate(self):
+        if self.date.year >= 2016:
+            return True
+        return False
```
ccostino: A bit nit-picky, but this could be simplified to be just `return self.date.year >= 2016`.  If this is easier to read though, then don't worry about changing it!

noahmanger: Ah thanks!

PR #590, opened 2016-11-08
Only show published updates

noahmanger:
This only shows published updates rather than all updates.

Fixes https://github.com/18F/fec-cms/issues/589#issuecomment-259191571


ccostino:
Thanks, @noahmanger!

Issue #582, opened 2016-11-07
Upgrade Django


noahmanger:
As part of our migration to the new hosting, we need to upgrade Django on the fec-cms. 

@LindsayYoung anything to add to this?

LindsayYoung:
Looks good, it shouldn't be a big lift

PR #602, opened 2016-11-22
Upgrading Django and Wagtail worked for me locally.

LindsayYoung:
It wanted me to do a migration I did a migration merge locally, but I don't think it was related to this so I did not commit it.


LindsayYoung:
closes #582

jontours:
Worked for me as well.

PR #614, opened 2016-11-25
This adds validation for users passwords

LindsayYoung:
This adds some checks for to user password strength. 

Note that superusers do not get the additional password validation.


ccostino:
Awesome, thank you @LindsayYoung!  For future reference to help others, the documentation around this lives here: https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#module-django.contrib.auth.password_validation

Issue #607, opened 2016-11-22
CMS Documentation Updates


ccostino:
Part of #541 

As part of our work to improve the development workflow for the CMS, we need to update our documentation to reflect recent and upcoming changes.  These should include the following:

* [x] Bringing in as much parity between our API and web app docs as possible
* [x] Updating info about databases from SQLite to PostgreSQL

ccostino:
I'm closing this issue as the work done so far has been merged in.  Remaining documentation updates will come as a part of the work they are coupled with.

PR #615, opened 2016-11-28
Update invoke and related tasks

### LINKED ISSUE #609: Update invoke and add invoke tasks for local development
Part of #541 

Another aspect of improving our local development workflow with the CMS is to add some additional support for [`invoke`](http://www.pyinvoke.org/) tasks.  Again, we'd like to achieve parity with our other projects as much as possible and help make managing local development a lot easier by automating repetitive tasks where possible.

These changes should include the following:

* [x] Updating the `invoke` dependency to the latest release
* [x] Modifying the existing `invoke` tasks to account for the new release
* [ ] Adding new `invoke` tasks:
   * [ ] Retrieving a database dump from the cloud.gov RDS service hooked up to the CMS via a private S3 bucket
   * [ ] Pushing a new database dump to a private S3 bucket to account for new migrations created during development
* [ ] Documenting any instructions (handled by #607)

Regarding the database backup management tasks, one idea that came up that we'd like to pursue is pulling down the RDS service backups via a private S3 bucket instead of just storing copies in the repo itself.  Unlike the data behind the API, some pieces of the CMS data could be private and we don't want to risk exposing it.  In theory, the backups would be created and exported directly to the private S3 bucket, which is where a developer would retrieve them from for local use.

It just so happens that @jcscottiii, who is on the cloud.gov team, is working on a CLI plugin to do just this very thing.  If possible, we'd very much like to leverage this as it'd make things a lot easier and smoother for us, much like the [`autopilot` plugin](https://github.com/contraband/autopilot) does for our deployments.

ccostino:
Part of #607
Part of #609

This changeset updates invoke to the latest stable release and updates our corresponding tasks to work correctly with it.  It will also provide new tasks for retrieving and pushing database dumps to an S3 bucket for the CMS app used by the FEC project.  These tasks will make use of the `cg-migrate-db` plugin for Cloud Foundry written by @jcscottiii.

Huge h/t and thank you to @jcscottiii for all of the hard work and help with the plugin!

**UPDATE (01/11/2017)**

I'm opting to take this PR out of WIP status as it is still in good shape, and work has been postponed on the other pieces for the time being.  Furthermore, `invoke` has been updated again with some other changes and I'd rather see these things go in now as opposed to having to redo them again, this way there will be less work when upgrading again in the future.

/cc @LindsayYoung, @jontours, @dogweather

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,5 +11,5 @@ whitenoise==2.0.3
 wagtail>=1.6,<1.7
 
 # Development
-invoke==0.10.1
+invoke==0.13.0
 GitPython==1.0.1
--- a/tasks.py
+++ b/tasks.py
@@ -12,15 +12,15 @@ env = cfenv.AppEnv()
 
 
 @task
-def add_hooks():
-    run('ln -s ../../bin/post-merge .git/hooks/post-merge')
-    run('ln -s ../../bin/post-checkout .git/hooks/post-checkout')
+def add_hooks(ctx):
+    ctx.run('ln -s ../../bin/post-merge .git/hooks/post-merge')
+    ctx.run('ln -s ../../bin/post-checkout .git/hooks/post-checkout')
 
 
 @task
-def remove_hooks():
-    run('rm .git/hooks/post-merge')
-    run('rm .git/hooks/post-checkout')
+def remove_hooks(ctx):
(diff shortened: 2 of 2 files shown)


ccostino:
D'oh, yeah, we probably could - I'll submit a separate PR for that after!  Thanks @vrajmohan!

PR #619, opened 2016-12-01
Fix nav

noahmanger:
This removes the parentheses from "About" in the utility nav and removes the "Contact" link from the utility nav.

### DIFF
--- a/fec/fec/templates/base.html
+++ b/fec/fec/templates/base.html
@@ -90,16 +90,12 @@
             <li>
               <a href="{{ cms_url }}/contact-us/">Contact us</a>
             </li>
-          {% if features.press %}
             <li>
               <a href="{{ cms_url }}/press/">Press</a>
             </li>
-          {% endif %}
-          {% if features.latest_updates %}
             <li>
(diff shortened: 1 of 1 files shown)


fec/fec/templates/base.html
```diff
@@ -62,10 +62,7 @@
               <a href="/legal-resources" class="site-nav__link{% if self.content_section == 'legal-resources' %} is-parent{% endif %}">Legal resources</a>
             </li>
             <li class="site-nav__item utility-nav__item is-disabled">
-              <a href="#" class="site-nav__link">(About)</a>
-            </li>
-            <li class="site-nav__item utility-nav__item">
-              <a href="/contact-us/" class="site-nav__link is-disabled">Contact</a>
+              <a href="#" class="site-nav__link">About</a>
```
ccostino: Is the `About` link going to appear as disabled?  It looks like it falls under that list item style now.

noahmanger: Yep, that's what we're going for.

ccostino: Ah okay, just wanted to confirm and make sure I understood.  Thanks!

PR #578, opened 2016-11-03
Improve CMS README, documentation, and instructions

ccostino:
Part of #609 
Part of #608
Part of #607 

This changeset represents the beginning of updates to our local development workflow, including updates to our documentation.  As more work is completed this will continue to be updated to account for new/updated commands and information.

This PR currently accounts for all of the documentation updates to correspond with #620.  Going forward we'll incorporate the documentation changes directly with the other PRs.

/cc @jontours, @LindsayYoung, @xtine, @emileighoutlaw

PR #620, opened 2016-12-01
Switch local DB to PostgreSQL

### LINKED ISSUE #608: Switch from SQLite to PostgreSQL
Part of #541 

As a part of improving our development workflow, we'd also like to establish parity with our CMS deployments as well as other parts of the overall beta FEC project.  We're currently using SQLite right now for local development with the CMS, but all other parts of our projects use PostgreSQL.

This work entails the following:

* [x] Changing our local development settings to use PostgreSQL instead of SQLite (please refer to [Django's `DATABASES` setting documentation](https://docs.djangoproject.com/en/1.10/ref/settings/#databases) for help)
* [x] Modifying our local database to be a PostgreSQL dump instead (only for 18F and FEC folks)
* [x] Updating documentation as necessary (handled by #607)

ccostino:
Part of #607 
Part of #608 

This changeset switches our local development setup to make use of PostgreSQL instead of SQLite.  The impetus behind this change is to bring parity across our projects (the API also runs on PostgreSQL) and the fact that the CMS as deployed also runs on PostgreSQL.  We want to take advantage of being able to pull down snapshots of production data to help with future development.  By having the same database backend locally we will be able to accomplish that.

As a part of this change, the following have been addressed here:

* Updating our Travis config to reflect the database change (this also updates the Python reference and tweaks the log output so we can actually see the end results of a build more easily).
* An update and minor cleanup to the Django settings to account for the change.
* A fix to the tests to account for the database having a different `auto increment` behavior.
* An update to the `django-libsass` package on account of other failing tests previously.
* Updated `cfenv` and `dj-database-url` packages.

Making use of a backup copy will come in a separate PR as that will involve integration and instructions with a Cloud Foundry plugin for use with cloud.gov that I've been working with @jcscottiii on.

**Note:**  The documentation around this change is in #578.

/cc @LindsayYoung, @jontours, @xtine


ccostino:
It looks like there's a problem with tests failing locally with some SCSS misconfiguration.  I'm not sure if it's just me or a problem in general (though @LindsayYoung tried and had the same result I did).

The error ultimately results in this error message.

```
sass.CompileError: Error: ".body-blocks .block-heading" failed to @extend ".heading__title".
       The selector ".heading__title" was not found.
       Use "@extend .heading__title !optional" if the extend should be able to fail.
        on line 22 of static/scss/fec.scss
>>     @extend .heading__title;
   ----^
```

I thought upgrading `django-libsass` might help since it was fairly old but that didn't fix the issue locally for me.  It does work just fine in Travis CI, though.

@noahmanger and @xtine, I'd like to pair with either or both of you just to investigate this a bit further to make sure I'm not introducing another issue into our setup before calling this ready to go.

noahmanger:
It looks to me like something is just really out of date. The error is ostensibly from this line, which was removed in September: https://github.com/18F/fec-cms/commit/b2e961a016e4c8121639c4dae33bafd27e629b9b#diff-0f7e18d7d5b1f370145401a1891b46f5L22

ccostino:
Oh wow, I wonder why I'm seeing that locally then...  even doing a `grep` through the codebase isn't producing that for me.

ccostino:
Figured out the issue - an old `static` directory sitting at the `fec-cms/fec` level was somehow getting read and sucked in.  Deleting the directory fixed the issue.  Not sure how that got there in the first place.

fec/fec/settings/base.py
```diff
@@ -105,10 +107,9 @@
 # https://docs.djangoproject.com/en/1.8/ref/settings/#databases
 
 DATABASES = {
-    'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
-    }
+    # Be sure to set the DATABASE_URL environment variable on your local
+    # development machine so that the local database can be connected to.
```
jontours: I don't have any background with the CMS, but if we deployed it locally using our local Postgres installation I don't see how the name of the database is specified, I'm guessing that's defined elsewhere and remains unchanged for this PR?

fec/home/tests/test_pages.py
```diff
@@ -25,6 +25,12 @@ def create_page(self, data=None):
         """
         if data is None:
             data = self.create_datas[0]
+
+        # We need to make sure we reference the correct author to account for
+        # foreign key relationships.
+        if 'authors-0-author' in data:
```
jontours: this seems rather hacky but as long as there are no worries of this naming convention changing, i.e. we control the naming of `authors-0-author`.  I have no background in the CMS so not sure.

fec/fec/settings/base.py
ccostino: Good question!  That is handled by the `dj-database-url` package for us, which expects an environment variable called `DATABASE_URL` to be set.  This is how our cloud.gov instances are currently running, so I just adapted it to the local setup too.

The documentation updates explicitly mention this and how to set it, so no worries about missing anything here, it is accounted for! :-)

fec/home/tests/test_pages.py
ccostino: It does feel hacky, but it was the cleanest way to address this without spending more time overhauling things in the tests (which may still be worthwhile, but not at the moment I don't think).

Thankfully, this is also something under our control, or at least predictable.  This key represents the ID that is constructed via Django's form API, which Wagtail utilizes for its admin pages (actually making use of something called [formsets](https://docs.djangoproject.com/en/1.10/topics/forms/formsets/)).  These field names in a form within a formset (if not customized) follow a predictable pattern of form name, form instance number, and model property name (which typically becomes the form field name if the defaults are left alone).

So in this case, this is entirely okay, but I can definitely understand the confusion and code smell to it! :-)

PR #621, opened 2016-12-01
Feature/rich-text-titles

noahmanger:
Ok! I think I have a way to do this. Basically:

- I added a new `CharField` to the the `PressReleasePage` model for a "Rich title"
- To make a portion of a title italicized, simply wrap it in `<i>` tags like so:
image

- I then made a custom template tag that checks if the page model passed in has a `rich_title` property, and if so, it returns that value, sanitized through Django's `format_html()` function
- This template tag is included in the press release page template as well as in the partial that's used to generate the blurbs in the feed for a result that looks like:
image

And:
image

The way I wrote the template tag code feels a little weird. I'm not sure if there's a better convention for what's pretty simple logic, so feedback welcome.

Resolves https://github.com/18F/fec-cms/issues/575


ccostino:
Oh nice, @noahmanger!  I think this is a smart way to go and seems to get the job done!  In thinking about this two things come to mind:

* You renamed the property name but forgot to update the migration to correspond as well.
* It might be better to use `<em>` instead of `<i>` for semantics (thinking of accessibility and screen readers), but the end visual result will be the same (italicized text).

Thanks for taking this on!  I'll review the template tag in more detail in a bit but for the most part it appears okay.  They are a bit wonky to create to begin with so it's not surprising it feels a little weird. :-)

noahmanger:
@ccostino thanks! Updated the help text to suggest using `<em>` instead. I also renamed the field back and checked in the migrations.

ccostino:
In looking at this more closely I honestly think you're okay with your template tag definition, and like that you're also checking for the existence of the property first before doing anything with it.  Nicely done!

The only thing I'd recommend here is maybe moving it into its own file called `tags.py` instead of putting it in `filters.py`, but that's nitpicking at the moment.  These methods are all small enough (and there's only the three) that it's easy to keep track of them at the moment.  I bring this up only because tags and filters are different things in a Django template, so for larger apps that define a bunch of each, it makes more sense to do so. :-)

PR #654, opened 2016-12-09
Adding AO process page

noahmanger:
This implements the AO process page as part of https://github.com/18F/fec-eregs/issues/290

This first pass simply implements it as a hard-coded template in the CMS, rather than making a new page model that would allow this to be edited via the admin interface. I'm open to doing that, but want to hear more about if the content here would change enough to warrant it...or if we just want to adopt the approach that everything should be editable via the Wagtail interface.

[Screenshot](https://cloud.githubusercontent.com/assets/1696495/21068080/4c7f19aa-be24-11e6-9a9f-1f140950f4b1.png)

This is ready for visual review.

### DIFF
--- a/fec/fec/templates/long_page.html
+++ b/fec/fec/templates/long_page.html
@@ -21,7 +21,6 @@
         <nav class="sidebar sidebar--neutral sidebar--left side-nav js-toc">
           <ul class="sidebar__content">
               {% block toc %}
-
               {% endblock %}
           </ul>
         </nav>
@@ -31,7 +30,6 @@
       <div class="main__content--right">
         {% block sections %}
         {% endblock %}
-
         <aside id="legal-citations" class="sidebar sidebar--primary">
           <h4 class="sidebar__title">Legal citations</h4>
           <div class="sidebar__content">
--- a/fec/fec/templates/long_page.html
+++ b/fec/fec/templates/long_page.html
@@ -9,7 +9,7 @@
 
 <article class="main">
   <div class="container" id="sections">
-    <h1 class="main__title">{{ self.title }}</h1>
+    <h1 class="heading--main">{{ self.title }}</h1>
     <div class="content__section">
       <div class="main__content">
         {% block intro %}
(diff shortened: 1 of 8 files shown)


noahmanger:
Ok @jenniferthibault I made those changes. 

[Check it out here](https://cloud.githubusercontent.com/assets/1696495/21168542/9dbec642-c169-11e6-92eb-4c2706dbeec6.png)

jenniferthibault:
Thanks @noahmanger ! 

I'm not sure if this is an artifact of the screenshot process, but the second section is looking a little garbled for me, just wanted to double-check with you in case. 

screen shot 2016-12-14 at 9 48 30 am


I missed one thing yesterday. The hand-delivery icon should be a little bit larger within the circle so that it bleeds to the edge. Here's a screenshot from the mockups:

screen shot 2016-12-14 at 9 48 30 am

Otherwise it all looks good! Thank you for your infinite patience with the tiny details!

noahmanger:
@jenniferthibault fixed

fec/home/templates/home/contact.html
```diff
@@ -42,10 +38,7 @@ <h5 class="contact-item__title">Teletypwriter (TTY) for the hearing impaired</h5
       </ul>
       <h2>Mailing address</h2>
       <ul>
-        <li class="contact-item">
-          <div class="contact-item__icon">
-            <img src="{% static "img/i-mail--primary.svg" %}" alt="Icon of a stamp">
-          </div>
+        <li class="contact-item contact-item--mail">
```
xtine: I like the use of making a specific class for the contact icons since mail/email/etc are going to be re-used again for other contact item lines.

fec/fec/templates/long_page.html
```diff
@@ -0,0 +1,55 @@
...
+        {% block intro %}
+        {% endblock %}
+      </div>
+    </div>
+    <div class="sidebar-container sidebar-container--left">
+      <div class="js-sticky-side" data-sticky-container="sections">
+        <nav class="sidebar sidebar--neutral sidebar--left side-nav js-toc">
+          <ul class="sidebar__content">
+              {% block toc %}
+
```
xtine: rm space

noahmanger:
Removed that extra space and also disabled the links that aren't available yet.

Issue #639, opened 2016-12-06
Implement "Commissioners" portion of home page


noahmanger:
So that users can understand the FEC better, implement a portion of the home page to feature the commissioners and link to the "About the FEC" section.

---

[Full home page design](https://cloud.githubusercontent.com/assets/1696495/20908227/2e474eea-bb08-11e6-9b55-e552a3b5c1cb.png) | This is the section for this issue:

![image](https://cloud.githubusercontent.com/assets/1696495/20908850/a3b824de-bb0c-11e6-93f0-cb521f5a7901.png)

The links for this section won't be active yet, so we can link to the legacy site for now: http://www.fec.gov/about.shtml and commissioner pages http://www.fec.gov/members/members.shtml

The order here is the chair goes first, then the vice chair, and then alphabetical by last name. So on mobile that order will need to be maintained.

**Completion criteria:**
- [x] Design is implemented as shown and all links go to the corresponding page on the legacy site

jenniferthibault:
Design tasks to prepare for implementation:

- [ ] Export headshot images

@xtine do you prefer I export them as circles, or as squares that you use a circular layer mask on the front end for?

xtine:
@jenniferthibault: if the pictures aren't going to be used elsewhere, then circles would be easier.

jenniferthibault:
[Google document](https://docs.google.com/a/gsa.gov/document/d/1bdbScZfwVtlauZRLtXh_-16h1BUUR26hPHC9pCLc-Os/edit?usp=sharing) with all the up-to date content so that it can be more easily copy & pasted 🔒 

[InVision](https://gsa.invisionapp.com/share/TD95OJRX3#/209035572_Home-V1_Copy_5) with final design and link paths

Here's how I was thinking the layout could stack on mobile:

![screen shot 2016-12-07 at 3 25 22 pm](https://cloud.githubusercontent.com/assets/11636908/20985190/67a411b6-bc91-11e6-874c-1f0730c7e8c0.png)

xtine:
@jenniferthibault / @noahmanger: I'm trying to come up with a name to describe the new type of headings next to the headshots. The most similar component would be the `contact-item`. The best I could come up with is "profile-item" ... any other suggestions?

noahmanger:
Ugh naming. Are you trying to think of a name for the component as a whole (pic + text)? Or just the name?

jenniferthibault:
Another similar bit is:

[...]

PR #662, opened 2016-12-15
New Homepage Commissioners section

xtine:
Issue #639 

screen shot 2016-12-15 at 5 18 08 pm
screen shot 2016-12-15 at 5 19 06 pm

Relies on https://github.com/18F/fec-style/pull/586

### DIFF
--- a/fec/home/templates/home/home_page.html
+++ b/fec/home/templates/home/home_page.html
@@ -88,7 +88,7 @@
           <div class="icon-heading">
             <img class="icon-heading__image" src="{% static "img/headshot--peterson.png" %}" alt="Headshot of Matthew S. Peterson">
             <div class="icon-heading__content">
-              <div class="t-lead t-ruled-bottom"><a href="http://www.fec.gov/members/petersen/petersen.shtml">Matthew S. Peterson</a></div>
+              <div class="t-lead"><a href="http://www.fec.gov/members/petersen/petersen.shtml">Matthew S. Peterson</a></div>
               <div class="t-note">Chairman</div>
               <div class="t-sans">Republican</div>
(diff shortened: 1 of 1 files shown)


fec/home/templates/home/home_page.html
```diff
@@ -74,6 +74,75 @@
...
+        <p>Established in 1975, the FEC is composed of six Commissioners who are appointed by the President and confirmed by the Senate. By law, no more than three can represent the same political party.</p>
+        <a href="http://www.fec.gov/about.shtml">Learn more about the FEC’s history and mission »</a>
+      </div>
+
+      <div class="grid grid--3-wide grid--no-border">
+        <div class="grid__item">
+          <div class="icon-heading">
+            <img class="icon-heading__image" src="{% static "img/headshot--peterson.png" %}" alt="Headshot of Matthew S. Peterson">
+            <div class="icon-heading__content">
+              <div class="t-lead t-ruled-bottom"><a href="http://www.fec.gov/members/petersen/petersen.shtml">Matthew S. Peterson</a></div>
```
noahmanger: Why do you need the `t-ruled-bottom` class? Doesn't the `<a>` come with underlining?

xtine: Good catch, I had that on there before I added the links and forgot to take them out.

xtine:
I prefer a simpler approach to just having the html markup for the commissioners as-is. I think there would be a case of storing it somewhere else if they were referenced on different pages or sections, but if it's just on the homepage then I don't see the need.

ccostino:
Thanks, @xtine!  Merging in since there are no blockers here. :-)

PR #666, opened 2016-12-16
Add import support for scraped content

### LINKED ISSUE #466: Import FEC Record data into CMS
Once the data has all been scraped and the new templates have been completed, import all FEC Record data into the Wagtail database.

Criteria for completion:
- [x] All FEC Records are included in the database as instances of the record page template

For each record, these properties should be populated:
- [x] Title
- [x] Published date (the original date it was published)
- [x] Categories
- [x] Keywords (will need to add a model property for this)
- [x] Monthly issue (will need to add a model property)
- [x] Author(s): These are database models, so we'll need to create new instances for each author
- [x] Author roles
- [x] Body content: If possible to import the HTML into a RichText field (rather than an HTML field) that would be best, in order to make it easier for people to edit, but it's not the end of the world

ccostino:
Closes #466 

This changeset adds a couple of missing pieces for running the scraper locally:

* Missing requirements entries
* Adjusted import statements a bit
* Modified the file setup to work as a Django Management Command

It also adds support for importing the data into the CMS.  The following pages types are supported:

* Weekly Digest Pages
* Press Release Pages
* Record Pages

**NOTE:** This PR requires that you have the scraped content available locally!  You can retrieve the record pages using the included scraper but for the other two importers the scrapers are external.

To run the weekly digest and press release importers, use these commands in the `fec/` directory:

```bash
./manage.py import_press_releases [--delete-existing] [--import-raw]
./manage.py import_weekly_digest [--delete-existing] [--import-raw]
```

To run the importer for the record pages, it is slightly different:

```bash
./manage.py import_record_pages <path_to_json_file> [--delete-existing] [--import-raw]
```

The `--delete-existing` flag will remove all existing models first before importing new content, and the `--import-raw` flag will import the body content as raw HTML instead of rich text.

To run the record page scraper itself, you need to go into the `fec/data_loader/management/commands` directory and run it by itself (it hasn't been converted to a management command yet):

```
./scrape_record_articles.py
```

**REVIEW ITEMS**

* Please see the code for `# TODO` comments as I had some questions around certain things.
* Identify and clean messy data as necessary.

To that last point, I've already noticed a couple of things:

* Weird `lockquote` elements and `blockquote` elements as text.
* Some incorrect keywords due to the parsing and cleanup that had to be done.

It'd be great to get some eyes and more testing around this, so any help is greatly appreciated!

**FINAL NOTE**

This PR contains a lot of work done by @LindsayYoung and @tadhg-ohiggins originally and collects all of these previous efforts.  Thank you both very much for the hard work!

/cc @LindsayYoung, @jontours, @noahmanger, @xtine, @tadhg-ohiggins


ccostino:
Oh interesting GitHub, you're pulling in comments from the other PR...  Cool, well, no loss of history. :-)

noahmanger:
These are looking really good! I had one small hiccup with running the scraper, but I had a previous download I was able to use.

One thing that would be good to do is remove the title text from the body of the record, since we already display it as a separate field. Otherwise it's duplicated:

image

Is that possible?

ccostino:
@noahmanger oh good catch!  I'll see what I can do to remove that.

ccostino:
What was the hiccup you encountered when running the scraper?

noahmanger:
@ccostino I can't recall exactly. I think it was that it didn't know where to save the new JSON output.

ccostino:
Ah, perhaps you need to make the `record_json` directory inside of `fec/data_loader/data/`?  If you don't have it yet try doing that and running the scraper again.  If it still doesn't work I can convert it into a management command like the importers.

noahmanger:
Hmm tried running it again and this is the error `'../../data/record_json/2016-12-28-164317.json'` . I made the `data/record_json/` directory before running.

ccostino:
Strange!  Is it just not able to write the file?  I'll take a look this morning and convert it to a management command too to see if that helps alleviate the issue.

ccostino:
This is now ready for a full review!

fec/data_loader/management/commands/import_record_pages.py
```diff
@@ -0,0 +1,224 @@
...
+
+            category = self.validate_category(
+                category,
+                DEFAULT_CATEGORY,
+                record_page_categories,
+                **options
+            )
+
+            slug = slugify(str(item_year) + '-' + category + '-' + title)[:225]
+            # TODO:  Is this path correct?
```
jontours: Um, wow there is a lot of code here in the PR, but I'll comment on maybe overlooked stuff: is this TODO resolved?

ccostino:
@jontours yes, there is a lot of code here, I'm sorry about that, thanks so much for going through!  It is a build-up of a lot of work done by several folks. :-)

As far as tests, you're right, this is intended to be "one-and-done" stuff that is going to be removed once the content is in.  We have no intention of maintaining this code at all, it's just to help us get the content as is for right now.

fec/data_loader/management/commands/import_record_pages.py
ccostino: It is not, but it's an easy enough tweak to make.

PR #680, opened 2016-12-19
Feature/local db setup instructions

patphongs:
Updated some of the postgres database set up to include initialization instructions and also auth migration instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,12 @@ npm install
 ### Create local databases
 Before you can run this project locally, you'll need a development database:
 
+Initialize postgres database into local folder where you want to run the database
+
+```pg_ctl init -D [local_pg_folder]```
+
+```/usr/local/Cellar/postgresql/9.5.0/bin/pg_ctl -D /data/postgres -l logfile start```
+
 ```bash
 createdb cfdm_cms_test
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -91,6 +91,12 @@ npm install
 ### Create local databases
 Before you can run this project locally, you'll need a development database:
 
+Initialize postgres database into local folder where you want to run the database
+
+```pg_ctl init -D [local_pg_folder]```
+
+```/usr/local/Cellar/postgresql/9.5.0/bin/pg_ctl -D /data/postgres -l logfile start```
```
ccostino: Ah, I missed this part due to [Homebrew](http://brew.sh/) taking care of it automatically.  Thanks for adding it in!  We might want to include a note with this that it is a step dependent on how you go about installing PostgreSQL (either with the installer or via Homebrew or some other means).

README.md
```diff
@@ -110,6 +116,7 @@ project setup by running these commands:
 npm run build
 cd fec/
 ./manage.py createsuperuser
+,/manage.py migrate auth
```
ccostino: I believe the line below with just the `./manage.py migrate` should take care of also handling the internal `auth` migrations along with any other migrations that ship with both Django and Wagtail.  Did it not do that for you?  We can run through a quick exercise to see for sure if you'd like!

PR #687, opened 2016-12-23
Removes recruiting message from legal home page

jenniferthibault:
## Summary

Removes unique instance of recruiting message, since we have recruiting streams going specifically to recruit legal users for upcoming sprints, and we have no way to prove that this message has actually resulted in any volunteering from users.

## Screenshots

screen shot 2016-12-23 at 1 31 42 pm


noahmanger:
I moved this page in https://github.com/18F/fec-cms/pull/693 so I just removed this section with that change. Closing this.

Issue #667, opened 2016-12-16
Make "About the FEC" landing page


noahmanger:
So that users can find their way to all kinds of organizational information about the FEC, create a landing page for the "About the FEC" section that introduces them to all of the sub-sections.

---

🖌 [Designs](https://cloud.githubusercontent.com/assets/11636908/21236524/5ad66f04-c2c9-11e6-916b-6a700b8d6451.png) | 🖊 [Content](https://docs.google.com/document/d/1z7ROcgSmbJHjmPIFDjL5Gdn722Ann2zdkbms9xdj_M8/edit) |  🗣 [Discussion](https://github.com/18F/fec-cms/issues/627)

This task is to make a new model and template in the CMS that can act as the "About the FEC" landing page. This page is basically identify to the [Press](https://beta.fec.gov/press) landing page, though that page has more custom components. It seems like we should figure out a good way to generalize this layout, perhaps with some sort of template inheritance behind the scenes.  

Because we're still sorting out the sub-pages of this section, these links may not be active at first. So this issue will be done when the basic page is made with disabled links to all of the subsections. I think it would make sense to keep this all behind a feature flag until we have at least some of the sub-sections populated, but open to suggestions.

**Completion criteria:**
- [x] Implement page as designed with disabled links for all sections
- [x] Page should be editable as a CMS template
- [x] Add link to this page in CMS nav
- [x] Add link to this page in web app nav
- [x] Add link to this page in eregs nav
- [x] Put page and all links behind a feature flag

noahmanger:
@patphongs did you say you had the links to the about page in a branch for the web app?

noahmanger:
Implemented by #690

PR #690, opened 2016-12-27
Added About landing page template

patphongs:
Issue #667 

Added About landing page

- Implemented page as designed with disabled links for all sections
- Page is editable as a CMS template
- Added link to footer of CMS, web app, and regulations pages
- Put about link behind a feature flag. About landing page can be left in draft mode until ready to publish live.

screen shot 2016-12-27 at 6 26 22 pm


fec/home/templates/home/about_landing_page.html
```diff
@@ -0,0 +1,87 @@
+{% extends "base.html" %}
+{% load wagtailcore_tags %}
+{% load staticfiles %}
+{% load updates %}
+{% block body_class %}template-{{ self.get_verbose_name | slugify }}{% endblock %}
+
+{% block content %}
+<section class="hero hero--secondary hero--press" aria-labelledby="hero-heading">
```
noahmanger: This should be `hero--about` instead of `hero--press` and should use the new image. @jenniferthibault can you share the About page hero background?

fec/home/templates/home/about_landing_page.html
```diff
@@ -0,0 +1,87 @@
...
+    <div class="sidebar-container sidebar-container--left">
+      <div class="js-sticky-side" data-sticky-container="options">
+        <nav class="sidebar sidebar--neutral sidebar--left side-nav js-toc">
+          <ul class="sidebar__content">
+              <li class="side-nav__item"><a class="side-nav__link" href="#mission-and-history">Mission and history</a></li>
+              <li class="side-nav__item"><a class="side-nav__link" href="#leadership">Leadership and structure</a></li>
+              <li class="side-nav__item"><a class="side-nav__link" href="#careers">Careers</a></li>
+              <li class="side-nav__item"><a class="side-nav__link" href="#reports">Organization reports</a></li>
+              <li class="side-nav__item"><a class="side-nav__link" href="#doing-business">Doing business with the FEC</a></li>
+              {% for option in self.option_blocks %}
```
noahmanger: Instead of hard-coding the different menu items and sections, you could just use the `option_blocks` for everything (both in the menu and in the content). It would make for a simpler template. What do you think?

fec/home/templates/home/about_landing_page.html
jenniferthibault: Just shared the hero image with  @patphongs via Slack.

patphongs: Thanks, @jenniferthibault has provided me with the hero image and I have updated accordingly.

patphongs:
@emileighoutlaw @nickykrause In terms of placement for the about page link. Did we only want the link to be in the footer of the page, which it currently is? I've also seen comps where the about link was in the header of the page, so I wanted to get your thoughts on it. 

about-link

nickykrause:
I recall that we were planning to use the header only for "utilities" (calendar, glossary, etc.), whereas the footer was for ancillary content. This would mean that `About` wouldn't be in the header.

@emileighoutlaw @jenniferthibault - Is my recollection sounding right to you?

emileighoutlaw:
I remember our research being split on whether the "About" section should be in the header or footer— the most important piece was that it either be in one or the other.  Putting it in the footer (and ideally testing to be sure users can find it there) seems consistent with our thinking and logical to me. 👍

emileighoutlaw:
@patphongs — where does this page live in Wagtail? I noticed a small typo (my fault) that I wanted to fix

patphongs:
@emileighoutlaw I've only just created the page model, so that needs to be merged into production in order for it to be an available landing page template. If it's in the introduction text under the separate headings, that can be edited when the new page is created in production, otherwise, I will have to make the update in the code. Go ahead and make the typo correction from the content document itself https://docs.google.com/document/d/1z7ROcgSmbJHjmPIFDjL5Gdn722Ann2zdkbms9xdj_M8/edit and let me know if you need me to update the code.

emileighoutlaw:
Sure thing!

Under `Leadership and structure` the first sentence should have a comma `,` after "Commissioners", not a semicolon `;` :upside_down_face: 

 I made that change in the Google document.

jenniferthibault:
@emileighoutlaw & @nickykrause yep! I'm 👍  with the first round of navigational testing linking to the About page from the footer.

noahmanger:
@patphongs were you still planning on refactoring this so that all the sections were dynamically generated?

patphongs:
@noahmanger Not right now. I think I just want to deploy it as is for now and then refactor later.

noahmanger:
There was a migration conflict with some other work, so I just pushed a fix for that and also removed a few stray instances of the old "about" links.

PR #691, opened 2016-12-27
Adding FEC record to update select

noahmanger:
Adds an option to the latest updates filter for FEC Records. Also adds the `.row` class to clear floats when the update blurb includes an image.

Wait to merge this until we've added the records.

But it's here if you want it for testing @ccostino .


ccostino:
Thanks, @noahmanger!

noahmanger:
Changed this to target the release.

noahmanger:
I just updated this to hide FEC record articles behind a feature flag, with one exception: if users are authenticated, the option to filter to only FEC records on latest updates will be visible and when specifically requested will be displayed. This way, we can publish the articles but they will only be accessible to authenticated users while they review.

PR #692, opened 2016-12-28
Up production instances

LindsayYoung:
should take care of the small periods of downtime we have seen


ccostino:
Thanks, @LindsayYoung!

PR #693, opened 2016-12-29
Feature/complaints page

noahmanger:
This is my attempt at answering the question from issue https://github.com/18F/fec-cms/issues/668 about how turn a common page template into more editable CMS fields. If it works, this will also be the template that we can use for the complaints page as well as other pages.

First, it creates a new page model, which I called `ResourcePage`. This page extends the `long_page.html` template. The model includes a few other models for the different components of the page:
- A `DocumentBlock` for generating thumbnails and links to documents like:
image

- A `AsideLinkBlock` for links to searches or calendar views:
image

- A `ContactItemBlock` for displaying blurbs of contact info like:
image

- And a `CitationsBlock` for adding citations at the bottom:
image

These can be combined to form a page that [looks like this](https://cloud.githubusercontent.com/assets/1696495/21534101/df8e66c8-cd16-11e6-8d29-bc5f72827292.png), all without editing the HTML. 

(I know that we need some style refinements on this).

---

While I was mucking around with this stuff I did a few other things:
- In order for a page like "Complaints process" to be nested in the right place, its parent pages need to exist in the CMS. In this case those pages are the Legal Resource landing page and Enforcement Matters page. The first one currently only existed as a view, not a page model, so I turned it into one. The second one existed as a static file in the web app. It doesn't need to do any web-app-y things, so I moved it to the CMS and made it a page model as well.
- Since I did this, it also made sense to take another step towards simplifying the code-base by combining the separate `legal` app with the `home` app. The separate app didn't do anything other than have a separate urls.py and views.py file, which were really small. Combining them simplifies sharing templates.

---

Super interested in feedback from anyone, but extra ping to @xtine @patphongs @johnnyporkchops @anthonygarvan or @ccostino . 

Willing to back off on any of these changes if there's a better way to go.


fec/home/models.py
```diff
@@ -323,3 +326,109 @@ class CollectionPage(Page):
...
+        ('link', AsideLinkBlock())
+    ],
+    template='blocks/section-aside.html',
+    icon='placeholder')
+
+class CitationsBlock(blocks.StructBlock):
+    """Block for a chunk of citations that includes a label and the citation (in content)"""
+    label = blocks.CharBlock()
+    content = blocks.RichTextBlock()
+
```
ccostino: I wonder if it'd make sense to pull these blocks out of the `models.py` file and move them into their own file, e.g., `blocks.py`, just to help keep things manageable and organized.

fec/home/templates/home/legal/legal_resources_landing.html
```diff
@@ -20,8 +20,7 @@
             {% endif %}
           </select>
           <input type="text" name="search" class="combo__input js-search-input"
-          aria-label="Search"
-          {% if search_query %} value="{{ search_query }}"{% endif %}>
+          aria-label="Search">
```
ccostino: Did you intend for this change?  By removing the `if` statement here, it won't pre-fill the form field with a search term if the user had previously searched for something.

ccostino:
Awesome work, @noahmanger!  I think you have a solid approach here and like the idea of creating the different blocks to help with customizing parts of the page but maintaining the same underlying base template(s).  👍 on the consolidation of the legal pieces back into the main app as well, it's not large or different enough to warrant being in its own app at this point.

Two things jumped out at me in the work done so far, but overall this is shaping up nicely. :-)

fec/fec/urls.py
```diff
@@ -22,8 +21,7 @@
     url(r'^calendar/$', home_views.calendar),
     url(r'^contact-us/$', home_views.contact),
     url(r'^search/$', search_views.search, name='search'),
-    url(r'^legal-resources/$', legal_views.home, name='legal'),
```
ccostino: Is this URL just flat out no longer needed or should it also have changed to reference a view moved to the `home` app?

fec/home/models.py
```diff
@@ -323,3 +326,109 @@ class CollectionPage(Page):
...
+class ResourcePage(Page):
```
ccostino: Just to be clear, this shouldn't inherit from our existing `ContentPage` class due to the unique nature of this page, right?  Or should it/would it be beneficial to do so?

fec/home/models.py
```diff
@@ -323,3 +326,109 @@ class CollectionPage(Page):
...
+        ))
+    ], null=True)
+    content_panels = Page.content_panels + [
+        StreamFieldPanel('intro'),
+        StreamFieldPanel('sections'),
+        StreamFieldPanel('citations'),
+        StreamFieldPanel('related_topics')
+    ]
+
+class LegalResourcesLanding(ContentPage, UniqueModel):
```
ccostino: Ah, one thing to bear in mind when [inheriting from multiple classes](https://docs.python.org/3/tutorial/classes.html#multiple-inheritance) - this doesn't work they way you likely expect in Python.  This looks like code that was just brought over from legal when you consolidated, so tagging @anthonygarvan and @vrajmohan as well as this is good to review since we hit it on the data side a few times as well.

[Long story short](https://www.python.org/download/releases/2.3/mro/), there's this concept known as the MRO (Method Resolution Order) and it has an impact on how class inheritance/hierarchy works.  In a very basic sense, the thing to bear in mind is that the class/object that you want to serve as the "base class" should be the _last_ class listed.  Any other abstract classes or mixins (just another name for an object in Python that folks use) should come before that.  The primary reason for this is that if you have methods named the same in these underlying classes you want to make sure Python references the correct one when either calling a method only defined in one of the parent classes or when referencing something via `super()`.

Now in this case, I'm not entirely sure what should serve as the base class, but I just wanted to note it in case they should actually be switched if the `ContentPage` is the base class.  If so, this should be written as follows:

```python
class LegalResourcesLanding(UniqueModel, ContentPage):
```

If this was already accounted for by others, then we're all good here!  Like I said, we've run into this recently on the data side too and it's definitely something worth reviewing, so I just wanted to make sure we took it into account properly. :-)

fec/home/models.py
```diff
@@ -323,3 +326,109 @@ class CollectionPage(Page):
...
+    subpage_types = ['ResourcePage', 'EnforcementPage']
+    template = 'home/legal/legal_resources_landing.html'
+    @property
+    def content_section(self):
+        return 'legal-resources'
+
+class EnforcementPage(ContentPage, UniqueModel):
```
ccostino: See note above about class inheritance.

fec/home/models.py
noahmanger: I like that idea. I was feeling like `models.py` was getting unweildy. Will do.

fec/home/templates/home/legal/legal_resources_landing.html
noahmanger: Yep that was my intention. As far as I can tell, this was never actually used, since queries are submitted to a different URL to view results.

noahmanger:
Update: Jen and I are going to pair on a few details for this later today, and then we can do the full review.

fec/fec/urls.py
noahmanger: No longer needed. I made this page an actual Wagtail page so that we can nest children underneath it.

fec/home/models.py
noahmanger: Yep. That page has a different structure so I thought it made more sense to just inherit from the base `Page`.

fec/home/templates/home/legal/legal_resources_landing.html
ccostino: Oh okay, good to know, thanks!

fec/home/models.py
noahmanger: Good to know. I actually modeled this after the way `class HomePage()` was implemented on line 66. It doesn't look like these two classes have any method name overlap.

fec/home/models.py
ccostino: Yeah, just moving these things into a `blocks.py` file in the same directory and importing them in the `models.py` file (likely with a statement like `from home.blocks import ...`) will help cut down on clutter in there. :-)

fec/home/models.py
ccostino: Ah perfect - just went back and looked at that code myself, I think this is correct and no need to worry here.

fec/home/models.py
ccostino: Disregard!

noahmanger:
Ok, reviewed with Jen and now this is all good to go.

.gitignore
```diff
@@ -59,5 +59,8 @@ npm-debug.log
 
 .DS_Store
 
+# Image uploads
+fec/media/
```
ccostino: If we're ignoring anything in this directory in the repo, should we also ignore it in Cloud Foundry as well?  In other words, make sure that if we ever do a manual deploy these files don't also go up with it?  If so, we should add this to `.cfignore` as well.

fec/fec/settings/base.py
```diff
@@ -185,9 +185,6 @@
 CONSTANTS = constants
 
 FEATURES = {
-    'latest_updates': bool(env.get_credential('FEC_FEATURE_LATEST_UPDATES', '')),
-    'legal_murs': bool(env.get_credential('FEC_FEATURE_LEGAL_MURS', '')),
-    'press': bool(env.get_credential('FEC_FEATURE_PRESS', ''))
```
ccostino: It's awesome to see these now just on and a part of the app. :-)  🎉

fec/fec/urls.py
ccostino: My mistake, I didn't put two-and-two together with the note about that before, thanks!

fec/home/blocks.py
```diff
@@ -0,0 +1,92 @@
...
+    contact_items = blocks.ListBlock(ContactItemBlock())
+
+    class Meta:
+        template='blocks/contact-info.html'
+        icon='placeholder'
+
+class CitationsBlock(blocks.StructBlock):
+    """Block for a chunk of citations that includes a label and the citation (in content)"""
+    label = blocks.CharBlock()
+    content = blocks.RichTextBlock(help_text='Use Shift + Enter to add line breaks between citation and description')
```
ccostino: Nice, thanks for adding the help text here and elsewhere!

fec/home/models.py
```diff
@@ -18,6 +18,9 @@
 from wagtail.contrib.table_block.blocks import TableBlock
 
 from fec import constants
+from home.blocks import (DocumentBlock, AsideLinkBlock, ContactInfoBlock,
+                        ContactInfoBlock, CitationsBlock, ResourceBlock,
+                        OptionBlock, CollectionBlock)
```
ccostino: Beautiful, much, much better having these in their own home!

ccostino:
The one thing I had mentioned about the `.gitignore` and `.cfignore` files is not worth keeping this from going on; we can deal with that separately if need be.  Merging!

PR #723, opened 2017-01-10
Commissioners landing and bio pages

xtine:
- Create model for Commissioner Page
- Template for each Commissioner Page
- View and template for All Commissioners 
  - Commissioner Chair will appear first. I set a query that finds any title with "Chair" but does not contain "Vice." So it will return any variation of _Chair/Chairman/Chairwoman/etc_
  - Commissioner Vice Chair will appear second. This query is any title that begins with "Vice" so it will accommodate variations of _Vice Chair/Vice Chairman/Vice Chairwoman/etc_
  - The rest of the current commissioners will then be sorted alphabetically by last name.
  - The past commissioners are sorted by term expiration date.

Needs twitter icon from: https://github.com/18F/fec-style/pull/604

https://github.com/18F/fec-cms/issues/702
https://github.com/18F/fec-cms/issues/669


xtine:
screencapture-127-0-0-1-8000-commissioners-1484020823629
screencapture-127-0-0-1-8000-caroline-c-hunter-1484020830006

xtine:
Commissioner Page in Wagtail
screencapture-127-0-0-1-8000-admin-pages-4348-edit-1484020934306

xtine:
@noahmanger: currently the urls for All Commissioners is `/commissioners` and page for each commissioner their slugified title (their full display name) like `/caroline-c-hunter`. Is this correct or should it be different?

Also I'm assuming commissioners links (Statements and opinions, Media and appearances) will be added in later?

noahmanger:
@xtine is it possible so that they're at `/commissioners/caroline-c-hunter`? I hadn't thought to look for a way to do this without having to make a new page model, but it looks like this might do the trick: http://docs.wagtail.io/en/v1.8/reference/contrib/routablepage.html

xtine:
@noahmanger: I figured out that URLs in Wagtail can be dependent on their page hierarchy in the CMS. So just creating a Custom page called "Commissioners" and then having all the Commissioner Pages live in there will set their URLS to `/commissioners/<commissioner-name>`.

xtine:
## Updated screenshots

All commissioners page with some past commissioners
screencapture-127-0-0-1-8000-commissioners-1484091360588

Commissioner page
screencapture-127-0-0-1-8000-commissioners-caroline-c-hunter-1484091386844

fec/home/models.py
```diff
@@ -336,6 +336,69 @@ class AboutLandingPage(Page):
         StreamFieldPanel('sections')
     ]
 
+class CommissionerPage(Page):
+    first_name = models.CharField(max_length=255, default='', blank=False)
+    middle_initial = models.CharField(max_length=255, blank=True)
+    last_name = models.CharField(max_length=255, default='', blank=False)
+    picture = models.ForeignKey(
```
noahmanger: I think you can just replace this with Wagtail's `ImageChooserBlock()`

fec/home/models.py
```diff
@@ -336,6 +336,69 @@ class AboutLandingPage(Page):
...
+        FieldPanel('commissioner_twitter'),
+    ]
+
+    def get_context(self, request):
+        context = super(CommissionerPage, self).get_context(request)
+
+        # Breadcrumbs for Commissioner pages
+        context['ancestors'] = [
+            {
+                'title': 'About',
```
noahmanger: This should be `About the FEC` for the title.

fec/fec/urls.py
```diff
@@ -15,13 +15,13 @@
     url(r'^django-admin/', include(admin.site.urls)),
 
     url(r'^admin/', include(wagtailadmin_urls)),
-    url(r'^documents/', include(wagtaildocs_urls)),
-
-    url(r'^updates/$', home_views.updates),
     url(r'^calendar/$', home_views.calendar),
+    url(r'^commissioners/$', home_views.commissioners),
```
noahmanger: I believe we want this nested like `/about/leadership-and-structure/commissioners`

fec/home/views.py
```diff
@@ -150,3 +151,30 @@ def ao_process(request):
...
+  past_commissioners = CommissionerPage.objects.filter(commissioner_title__exact='', \
+    term_expiration__isnull=False).order_by('-term_expiration')
+
+  page_context = {
+    'title': 'All Commissioners',
+    'chair_commissioner': chair_commissioner,
+    'vice_commissioner': vice_commissioner,
+    'current_commissioners': current_commissioners,
+    'past_commissioners': past_commissioners,
+    'ancestors': [
```
noahmanger: This should also include `Leadership and organization` / `/about/leadership-and-organization/` and `About` should be `About the FEC`.

fec/home/models.py
xtine: `ImageChooserBlock()` seems to only be used within StreamFields, whereas setting `wagtailimages.Image` is the way in the documentation to set images in Page models.

fec/fec/urls.py
xtine: Added so either URL works. Especially if a Commissioners custom page is made on the CMS, then any hit to `/commissioners` will return the correct page than a blank one.

fec/home/models.py
noahmanger: Ah gotcha.

PR #728, opened 2017-01-11
Feature/link-block

noahmanger:
This adds a couple more capabilities to the CMS. 

First, it adds a new button to the rich text controls that is used to make a link sans-serif:

demo

Which gets us this:
image

Second, it adds a new block to the `section` StreamField to insert links to other pages, which will be used to get this sort of component on the new Candidate and Committee Services page (ignore the copy...these are just random pages in a random place for illustration purposes):
image

### DIFF
--- a/fec/fec/static/css/customize-editor.css
+++ b/fec/fec/static/css/customize-editor.css
@@ -6,6 +6,14 @@ li.sequence-member .struct-block .sequence-container {
   width: 83%;
 }
 
+li.sequence-member .struct-block .sequence-container .field-content {
+  width: 100%;
+}
+
 li.sequence-member .sequence-member-inner .sequence-type-list .sequence-container-inner {
   width: 100%;
+}
+
+.t-sans {
+  font-family: sans-serif;
 }
--- a/fec/fec/static/js/hallo-edit-html.js
+++ b/fec/fec/static/js/hallo-edit-html.js
@@ -12,13 +12,6 @@
             populateToolbar: function(toolbar) {
                 var button, widget;
 
-                var getEnclosing = function(tag) {
-                    var node;
-
-                    node = widget.options.editable.getSelection().commonAncestorContainer;
-                    return $(node).parents(tag).get(0);
-                };
-
                 widget = this;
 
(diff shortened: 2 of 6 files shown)


fec/fec/static/js/hallo-edit-html.js
```diff
@@ -12,13 +12,6 @@
             populateToolbar: function(toolbar) {
                 var button, widget;
 
-                var getEnclosing = function(tag) {
```
noahmanger: Just noting that this is removed because it wasn't actually being used.

fec/fec/wagtail_hooks.py
```diff
@@ -5,29 +5,7 @@
 @hooks.register('construct_whitelister_element_rules')
 def whitelister_element_rules():
     return {
-        # Commenting out disallowed tags so its easier to remember & revert
-        'a': attribute_rule({'href': check_url}),
```
noahmanger: I learned that these rules are the default ones, so we don't need to include them... just the ones that we want to add.

fec/fec/static/css/customize-editor.css
```diff
@@ -6,6 +6,14 @@ li.sequence-member .struct-block .sequence-container {
...
+li.sequence-member .struct-block .sequence-container .field-content {
+  width: 100%;
+}
+
 li.sequence-member .sequence-member-inner .sequence-type-list .sequence-container-inner {
   width: 100%;
 }
+
+.t-sans {
+  font-family: sans-serif;
```
xtine: Does it not get this class from fec-style?

emileighoutlaw:
Woohoo! Just popping into this PR to say how exciting this work is 💟

fec/fec/static/css/customize-editor.css
noahmanger: This stylesheet is loaded by the admin interface, which doesn't load fec-style. I added this as a one-off style so that you can know that the button works.

xtine: Gotcha!

PR #732, opened 2017-01-11
Updates hero content

emileighoutlaw:
Content managers want to update to the hero content so that our first, giant message to the world is one that works harder to introduce the FEC (rather than summarize its complex mission). This PR updates the hero content to a more plain language introduction. 

Resolves: https://github.com/18F/FEC/issues/574#issuecomment-271988829


noahmanger:
Because this shortened the length to just one line, it broke the display:

image

This will require a quick patch to fec-style.

PR #733, opened 2017-01-11
Commissioner Page (reappointment date + ex officio) updates

xtine:
- Add `reappointed_dates` as a text field in the CMS below term expiration date to show if a Commissioner was reappointed during any time period. Appears in both All Commissioners and Commissioner page.
- Show Ex officio Commissioners on the All Commissioners page. Because these are past positions only, I think we can just have them hardcoded rather than controlled in the CMS.

https://github.com/18F/fec-cms/issues/669


xtine:
## Screenshots
screencapture-127-0-0-1-8000-about-leadership-and-structure-commissioners-1484173500842

---

screencapture-127-0-0-1-8000-commissioners-david-m-mason-1484173510197

PR #735, opened 2017-01-12
Upgrade Wagtail + commissioner block

noahmanger:
This makes it possible to show a block of the current commissioners on the "Leadership and structure" page (and anywhere else).

image

To do this, though, requires updating Wagtail in order to make use of the new `StaticBlock` type. This is a way to add plain HTML inside a StreamField without any configuration. It works like this:

static block demo

This basically inserts a template partial (`blocks/commissioners.html`) which in turn calls a new `{% current_commissioners %}` template tag, which renders a list using the `partials/commissioner.html` template we use on the main commissioner page.

I added an option to that partial so the dates / "currently serving" text only shows if told to, and also fixed a couple broken URLs.


noahmanger:
Tagging both @xtine and @ccostino in for the review just because of the wagtail upgrade.

fec/home/templates/partials/commissioner.html
```diff
@@ -12,18 +12,20 @@
...
       {% endif %}
       <div class="u-padding--top t-sans">{{ commissioner.get_party_affiliation_display }}</div>
-      <div class="t-italic">
-      {% if commissioner.term_expiration %}
-        {{ commissioner.sworn_in|date:'F Y' }} to {{ commissioner.term_expiration|date:'F Y' }}
-        {% if commissioner.reappointed_dates %}
-      </div>
-      <div class="t-italic">
-        ({{ commissioner.reappointed_dates }})
+      {% if show_dates == True %}
```
xtine: This can be shorted to just `{% if show_dates %}`

noahmanger: 👍

ccostino:
Awesome, thanks @noahmanger!  Looks like there are some migration issues, and the upgrade to Wagtail 1.8 will require a couple of additional changes.  I'll add some updates to this PR to address these things and tag folks for another review when done.

ccostino:
Okay, I made the one change that was necessary with this upgrade, but there's another we could potentially do.  However, I don't think it's necessary as it's related to reverting to older behavior around who has permission to delete things in bulk:  http://docs.wagtail.io/en/v1.8/releases/1.8.html#non-administrators-now-need-bulk-delete-permission-to-delete-pages-with-children

I think as a precaution it might make sense to not add this.  What do you think, @xtine, @jenniferthibault, and @noahmanger?

jenniferthibault:
I'd agree @ccostino that it's not necessary for addressing this issue, but might be worth filing as an issue for upgrading later. If you agree, I'm glad to file the issue in the backlog.

ccostino:
Awesome, thanks @jenniferthibault!  Yeah, that makes sense to file a small issue just to fully investigate to see if it's a warranted thing in the near future.  But definitely not needed as a part of this.

ccostino:
@xtine:  I believe this is in good shape now, but I want to make sure the migration changes I had to make are good to go; would you mind taking another look, please?  Thanks!  If you find anything just let me know and I'll take care of it ASAP.

xtine:
@ccostino: all the changes look good, and I didn't get any issues with the migrations. Is it safe to merge?

ccostino:
@xtine:  great, thank you!  Yes, this is good to go now.  Unfortunately anything else in flight with migrations will have conflicts again and need to be resolved, but hopefully we're nearing the end of all of the churn.

PR #738, opened 2017-01-13
Only show published digests and releases on press page

noahmanger:
This fixes a bug where draft press releases and weekly digests show up on https://beta.fec.gov/press/

cc @AmyKort


ccostino:
Nice catch, thanks @noahmanger!

PR #747, opened 2017-01-18
Capability to upload documents and images from Wagtail to s3

patphongs:
Issue https://github.com/18F/fec-cms/issues/714

Used django-storages and boto3 backends to provide this capability


fec/fec/settings/base.py
```diff
@@ -210,3 +210,11 @@
     RAVEN_CONFIG = {
         'dsn': os.getenv('SENTRY_DSN'),
     }
+
+AWS_QUERYSTRING_AUTH = False
```
LindsayYoung: Can we preface these with `CMS_` so it is clear which bucket we are using, we have 3 buckets across the environments.

LindsayYoung:
I am setting up the infrastructure for this, I think I can get decent urls going so we don't have to worry about switching them now.

Pat and I talked bout how to get local environments using file folders, instead of s3.

fec/fec/settings/base.py
patphongs: added the `CMS_` prefix for this

patphongs:
@LindsayYoung @noahmanger I added some logic to check a local environment variable so that when you start up Wagtail, your images can still be uploaded to your local rather than to the s3 bucket.

Let me know what you think about the solution. Ideally in the future we should probably try to default `ENVIRONMENTS.get(os.getenv('FEC_CMS_ENVIRONMENT'), 'DEVELOPMENT')` from DEVELOPMENT to LOCAL since those who clone this repo will be developing locally. I just haven't looked at the code in detail as to what the DEVELOPMENT environmental variable is pulling in.

fec/fec/settings/base.py
```diff
@@ -210,3 +211,13 @@
     RAVEN_CONFIG = {
         'dsn': os.getenv('SENTRY_DSN'),
     }
+
+if FEC_CMS_ENVIRONMENT != 'LOCAL':
```
noahmanger: I love this, but what about instead of setting a new environment we just check to make sure that the env isn't `dev`, `stage`, or `prod`? One less step when getting it going.

README.md
```diff
@@ -95,13 +95,19 @@ Before you can run this project locally, you'll need a development database:
 createdb cfdm_cms_test
 ```
 
-You will also need to set a connection string for the local database as an
+You will also need to set 2 environmental variables:
```
noahmanger: Thanks for updating the README!

LindsayYoung:
Infrastructure stuff is in reasonable shape.

made a  s3 bucket in each environment in govcloud 
set cors policy in each bucket
added the s3 credentials in each corresponding environment in e/w

Use the variable names:
 "CMS_AWS_ACCESS_KEY_ID"
 "CMS_AWS_SECRET_ACCESS_KEY"
 "CMS_AWS_STORAGE_BUCKET_NAME"
 "CMS_AWS_DEFAULT_REGION"

Josh Carp said he could take a look tonight or tomorrow about proxying the url names. I think it might be a real pain to change those later. (In theory it should be easy to proxy the bucket, but I did not have luck earlier.)

fec/fec/settings/base.py
LindsayYoung: We use develop as the default environment. So, it will be used by default locally and by the development server, in this case we want them to act differently. 

We could add a new local default but we would need to look carefully at the code to make sure that did'n have unintended consequences.

patphongs:
@noahmanger The 403 error should be fixed now with the latest commit. HUGE shout out to @LindsayYoung for helping me figure out the issue! Problem was that boto3 was not passing in region name correctly. Please test it out and see if it's ok for you all. 

**One note is that we should use the image upload feature sparingly for now because the URL for the buckets will eventually change once @jmcarp and @LindsayYoung are able to figure out how to proxy the buckets to a more friendly URL. Once that happens, we'll need to delete the images in Wagtail and re-add so that all of the mapping and references through the application are correct.

fec/fec/settings/base.py
patphongs: Did a quick search for FEC_CMS_ENVIRONMENT variable in the code and found that it is currently only being used for the top banner to print out whether or not you are in dev or prod. So I feel like it would be safe to export default as local? What do you think @LindsayYoung

LindsayYoung:
depends on this, which is ready for review https://github.com/18F/fec-proxy/pull/11

fec/fec/settings/base.py
```diff
@@ -210,3 +212,13 @@
...
+    AWS_QUERYSTRING_AUTH = False
+    AWS_ACCESS_KEY_ID = os.environ['CMS_AWS_ACCESS_KEY']
+    AWS_SECRET_ACCESS_KEY = os.environ['CMS_AWS_SECRET_ACCESS_KEY']
+    AWS_STORAGE_BUCKET_NAME = os.environ['CMS_S3_BUCKET_NAME']
+    MEDIA_URL = 'https://%s.s3-us-gov-west-1.amazonaws.com/' % AWS_STORAGE_BUCKET_NAME
```
LindsayYoung: I am pretty sure for this to work with the proxy pr we want it different for each environment, so we should make AWS_STORAGE_BUCKET_NAME an env var and put the following values in the cf service
for dev:
https://fec-proxy.18f.gov/resources-dev/
for staging:
https://fec-proxy-stage.18f.gov/resources-stage/
for prod:
https://beta.fec.gov/resources/

patphongs:
@LindsayYoung I just tried to access an image uploaded to the bucket via this URL: https://fec-proxy.18f.gov/resources-dev/cms-content/images/fec-ftw.max-165x165.png Is that the way the URL is supposed to be constructed or am I missing something? 

Image is accessible from this URL: https://s3-us-gov-west-1.amazonaws.com/cg-47928592-406c-4536-8234-99b896e8d57d/cms-content/images/fec-ftw.max-165x165.png

ccostino:
@patphongs, it seems you uploaded the image to the `prod` S3 bucket, so the URL should have been https://fec-proxy.18f.gov/resources/cms-content/images/fec-ftw.max-165x165.png instead.  However, this will no longer work either as that is the `dev` proxy, which will not point to the `prod` S3 bucket.

If you upload the image to the `dev` S3 bucket with the same name and in the same location, https://fec-proxy.18f.gov/resources/cms-content/images/fec-ftw.max-165x165.png should then work.  Once we have the proxy app fix up in `prod`, then https://beta.fec.gov/resources/cms-content/images/fec-ftw.max-165x165.png should work as well. :-)

Please see https://github.com/18F/fec-proxy/pull/11 for more info!

fec/fec/settings/base.py
ccostino: This is now done as of the merge in the `fec-proxy` and the content-bucket PR: https://github.com/18F/fec-proxy/pull/11

Now there is only ever the `/resources/` URL for any given environment, and under the hood it is connected to the S3 bucket associated with that environment.  It looks like we'd just need to expose that same value in an environment variable for the CMS app as well, at least until we shift things over into GovCloud and put it in the `VCAP_SERVICES`.

README.md
ccostino: Yes, always nice to see, thank you! :-)

fec/fec/settings/base.py
ccostino: I think there's an opportunity to improve this for ourselves and make it easier to get things going locally.  @patphongs and I will investigate at some point soon! :-)

patphongs: @ccostino @LindsayYoung -- Based on my conversations with Carlo today, it seemed like it would be safe to make local our new default. This will make it so that your local dev will just upload assets locally by default. Please test and verify.

patphongs:
@ccostino Just tested the changes out locally using the new env.get_credential method and it works ok. Please review.

ccostino:
Ah because of the branch switch, we'll have to do a quick merge of `develop` into this PR - I can take care of that!

ccostino:
Thank you, @patphongs!

PR #752, opened 2017-01-20
Cleaning up aspects of the fec record articles

noahmanger:
This makes a few of the clean up changes from https://github.com/18F/fec-style/issues/614

- Adds a "Read more" link (needed to use javascript to get it inline)
- Capitalizes the first letter of the categories
image

- When clicking on a tag, it now takes you to the `#feed` div so that the filters are at the top of the screen. 
- Removes the `p--lead` sizing on the first paragraphs.
- If title and email are 'N/A' then they're not displayed on posts

Last, it adds space between the "next item" and the link back to the latest updates page, resolving https://github.com/18F/fec-cms/issues/712:
image

### DIFF
--- a/fec/fec/static/js/fec.js
+++ b/fec/fec/static/js/fec.js
@@ -137,7 +137,7 @@ $(document).ready(function() {
   // Post feed
   // Move the read more links to be inline with the snippet from the post
   $('.js-post-content').each(function() {
-    var $p = $(this).find('p');
+    var $p = $(this).find('p:last-of-type');
     var $link = $(this).find('.js-read-more');
     if ($p.text() !== 'PDF') {
       $p.append($link);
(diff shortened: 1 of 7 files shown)


jenniferthibault:
I should flag that once these changes are made, it's probably best to transfer to @xtine for technical review. I'm 🙅 enough for reviewing js

noahmanger:
Got it! 

image

The reason the ellipsis is showing up as part of the link is because we're truncating after a certain number of words and the link in this case contains more words beyond that limit, so it puts the ellipsis in the middle. That's going to be kind of a weird edge case that just depends on where a link falls in the body of text and how long it is, so I'd prefer if we can just leave it as it would be pretty complicated to implement logic to avoid.

noahmanger:
But I did make it so the underline lines up:
image

jenniferthibault:
Thanks @noahmanger that makes sense to me. Transferring to @xtine for technical review 👋

PR #490, opened 2016-09-14
Agenda Page model and template

johnnyporkchops:
Proposing the addition of a Wagtail model, AgendaPage and corresponding template: agenda_page.html. 
Example of current fec.gov page: http://www.fec.gov/agenda/2016/agenda20160915.shtml

PR #756, opened 2017-01-23
Agenda page model/template

johnnyporkchops:
This replaces **https://github.com/18F/fec-cms/pull/490**:
Changes requested in #490 are done as well as additional enhancements

For design notes see: 
**https://github.com/18F/fec-style/issues/509 for design notes**

Below is a screenshot of an example page published on local instance using the template. The page is sparingly styled, anticipating input from design/UX folks.

agenda_page


fec/home/models.py
```diff
@@ -429,6 +432,52 @@ class CollectionPage(Page):
         StreamFieldPanel('sections'),
     ]
 
+
+
+class AgendaPage(Page):
+    author= models.CharField(max_length=255)
```
noahmanger: Does this need an author?

fec/home/models.py
```diff
@@ -429,6 +432,52 @@ class CollectionPage(Page):
...
+    date = models.DateField('Post date')
+    mtg_date = models.DateField(default=datetime.date.today)
+    mtg_time = models.CharField(max_length=255, default ='10:00 AM')
```
noahmanger: I don't have a strong opinion, but might be better to store this as an actual datetime object instead of a string?

fec/home/templates/home/agenda_page.html
```diff
@@ -0,0 +1,110 @@
...
+            {% endif %}
+          {% endfor %}
+         </ul>
+        {% endif %}
+       </li>
+      {% endfor %}
+   
+  </ol>
+    </div>
+    {% if self.sidebar %}
```
noahmanger: Does this page model actually have a sidebar? Doesn't look like it does.

fec/home/models.py
```diff
@@ -429,6 +432,52 @@ class CollectionPage(Page):
...
+    mtg_media = StreamField([
+        ('full_video_url', blocks.TextBlock()),
+        ('full_audio', DocumentChooserBlock(required=False)),
```
noahmanger: Your call: do we want to use Wagtail's built in document panels, or just paste in the URLs to assets in S3 manually?

fec/home/models.py
```diff
@@ -429,6 +432,52 @@ class CollectionPage(Page):
         StreamFieldPanel('sections'),
     ]
 
+
+
+class AgendaPage(Page):
+    author= models.CharField(max_length=255)
+    date = models.DateField('Post date')
```
noahmanger: Oh, and I don't think we need this date field, either.

dogweather:
@ccostino @noahmanger Do you think this code can be merged, and then we can build on it? I'm not familiar enough with the frameworks to render an opinion. There are no tests, and the test coverage is really low, so I can't be sure this won't break anything.

johnnyporkchops:
I was holding off on making any of the requested changes until meeting with @dogweather, I can make these changes now

fec/home/models.py
johnnyporkchops: Regarding: Using Wagtail's built in document panels, or just paste in the URLs to assets in S3 manually for audio file upload: I think recent discussions about S3 assets will inform this, I will circle back with @patphongs

fec/home/templates/home/agenda_page.html
johnnyporkchops: No, this does not have a sidebar, good call

fec/home/models.py
johnnyporkchops: Does not need author

noahmanger:
Yeah, once the changes are made, I think it's ok to merge. @ccostino / @xtine there's migrations, though, so not sure how to factor the pending squash in. Do we need to wait for that?

Re: tests, I'm fine with merging this in and then adding tests so that the scraper work can be unblocked. (And I know we also need to step up coverage for the rest of the models as well 😖 ).

johnnyporkchops:
I may need to resolve conflicts locally before I can make changes and update PR, right now, even if I pull and makemigrations --merge, I get error when navigating through Wagtail Admin. So I could make the changes but I am afraid if I update the PR, it will fail tests again. @ccostino, Do you have any input?

dogweather:
@johnnyporkchops Normally, to handle new code with schema changes vis-a-vis your local dev database, the best practice I know of is to just delete and re-create the local dev database. You shouldn't need to make changes to the code, unless these errors are true errors that we've committed and need to be fixed. 

Run the tests; do they pass?

ccostino:
@noahmanger:  Don't worry about any other migrations in here impacting the squashing work.  It's easy enough to incorporate in and since this is already in flight, I'd rather see it get in before the squash happens.  As it is, we had do to this step anyway as a part of that work so it's just happening here instead. :-)

@johnnyporkchops:  First of all, do not fear changing the code if it's a matter of getting something done, fixing something, or trying something new to see if it meets a need!  If the tests fail, that is valuable feedback and lets you know that either your code needs to be fixed or the test(s) need(s) to be updated. :-)

In any case, what error are you seeing in the admin?  Did you apply the migrations (`./manage.py migrate`) after running `makemigrations --merge`?  If you have some time tomorrow, we should be able to pair a bit if that'd be easier.

@dogweather:  I agree for some situations, but that shouldn't be necessary in this case.  It sounds like something else is going on here.

fec/home/models.py
ccostino: I'd err more on the side of whatever provides a better workflow and is easier to use for folks who will be managing this content in the admin.

fec/home/models.py
```diff
@@ -609,4 +658,4 @@ def content_section(self):
 
     @property
     def hero_class(self):
-        return 'services'
+        return 'services'
```
ccostino: Nit-picky code style thing:  Python files (pretty much any file format when it comes to software for that matter) typically have a blank new line as the last line in a file.  This icon you see here is GitHub's way of letting you know that one is not present in this file.  Adding a blank new line will address this and make the code more adherent to things like [PEP 8](https://www.python.org/dev/peps/pep-0008/)! :-)

fec/home/templates/home/agenda_page.html
ccostino: Should it have a sidebar or is it okay without one?  Sorry, I'm still confused based on the comments here. :-)

johnnyporkchops:
Ok, I will make the requested changes to the code and update PR ASAP. @ccostino, yes I get Django error page after doing both mamemigratipns --merge and migrate. It's an old error message that I remember having to do with press landing page so it must be something simple, we can pair tomorrow at your convenience, thanks!

johnnyporkchops:
Re- sidebar: No there is no sidebar so I can take that out of template

fec/home/models.py
johnnyporkchops: @ccostino wrote:" Are the binary files supposed to be a part of this PR?" 
Answer: If you are referring to the audio (and other documents) that were added-- NO, that was a mistake I made in the beginning, I thought they needed to be added for example purposes b/c at there were no documents in Wagtail upload area.

ccostino: Ah okay - you can remove them easily from the branch using `git rm path/to/files` and committing and pushing those changes to the PR, and they'll be removed!

ccostino: I just saw @noahmanger's remark in his earlier review that I missed when looking things over - adding `fec/media/` to the `.gitignore` file (which is just a listing of a file name, path, or regex matching some semblance of a file name or path) will also help fix the issue and is probably a good thing to do in general as it'll help prevent things like this from happening in the future (Git doesn't bother trying to track things listed in the `.gitignore` file).

ccostino:
Sure thing, @johnnyporkchops!  If it's an old error you've seen before, I think I know what it is you're talking about.

I'm hoping our efforts around the squashed migration and ability to pull production database backups for local development will prevent that from happening in the future.

dogweather:
FYI @noahmanger and @johnnyporkchops, Tadhg had finished the scraping script. It produces this data. A list of `Meeting` objects:

```python
    "Meeting",
    [
        ("agenda_document_links", Links), # list (Links)
        ("approved_minutes_date", Date),
        ("approved_minutes_link", Link),
        ("body", str),                    # HTML,
        ("closed_captioning_link", Link),
        ("draft_minutes_links", Links),   # list (Links)
        ("link_title_text", str),         # Plain text
        ("meeting_type", str),            # "open" or "executive"
        ("pdf_disclaimer", str),          # HTML (outer HTML)
        ("posted_date", Date),
        ("primary_audio_link", Link),
        ("old_meeting_url", str),         # URL
        ("secondary_audio_links", Links), # list (Links)
        ("sunshine_act_links", Links),    # list (Links)
        ("title_text", str),
        ("video_link", Link)
    ])
```

johnnyporkchops:
I made requested changes to the model and template and updated the PR. I tested these locally and there does not seem to be any issue, but it is failing travis checks as we knew it would because of the current migration discrepencies. I was advised to update the PR even though it will not pass checks.

noahmanger:
@ccostino what do you recommend around the migrations issue? 

@johnnyporkchops can you remove the files that were checked in in `/media/`? 

And thanks @dogweather for that. I see that there's discrepency between the model built in this PR and that, but I think rather than trying to continue to juggle these as two separate PRs it'd be good to merge in this PR ASAP and then handle changes to the page model that are needed to mirror the scraper in the scraper PR. How does that sound?

johnnyporkchops:
I removed the media files

ccostino:
@noahmanger:  For the migrations issue, I'm not sure.  Were these migrations with the lower numbers in the beginning supposed to be created, are they expected?  Or should we really only be seeing things from `0065` and beyond?  Or were those created when this work started and therefore are in the right place, it's just a matter of timing?

dogweather:
@noahmanger I agree — let's merge this in, delete the branch, and I can work on the import script and update the models as needed.

johnnyporkchops:
@noah, I am not sure how to answer your migrations question. Yes this is an fairly old OR , although it was already recreated once to get past migration issues. (from PR #490 to #756) I told @ccostino in a conversation about this issue, that I would gladly kill and recreate the PR yet again( from a fresh checkout of /develop if need be.

johnnyporkchops:
@dogweather wrote: "let's merge this in, delete the branch, and then I can work on the import script and update the models as needed."
Agreed!

noahmanger:
This PR only adds a new model which shouldn't be causing this many migration headaches. I started a clean branch here that just includes the model and template changes from this and checks in a fresh migration https://github.com/18F/fec-cms/pull/853 . Closing this in favor of the new one.

johnnyporkchops:
Thats exactly what I was going to do, thanks @noahmanger

PR #765, opened 2017-01-26
Commissioner template logic

xtine:
Remove nested template logic for Commissioner grids so HTML shows up cleanly.


jenniferthibault:
I am so excited that I think I can actually translate this to plain English! Thanks @xtine !

PR #772, opened 2017-01-30
Adding select option for tips for treasurers

noahmanger:
And moving the backend logic behind its own feature flag:

image


ccostino:
Thanks, @noahmanger!

PR #792, opened 2017-02-02
Variable names to match the env vars

LindsayYoung:
Looked at the env in the production cms and fixed the variables accoddingly

I feel like it is easier to change here since we have to replace the whole thing in the env.


patphongs:
Thanks @LindsayYoung! I pulled your branch and tested it with the new names and everything still worked. Do you think dev and staging also have the same environmental variable names as prod? @ccostino has been trying to figure out why we get a 500 error when trying to access the images page. Maybe this would help?

ccostino:
Thank you, @LindsayYoung!  Always helps to have another set of eyes. :-)  I will merge this in and check staging again and continue on from there if we still run into an issue.

PR #793, opened 2017-02-02
Making author title and email non-required

noahmanger:
This makes the author email and title fields for latest updates posts
non-required. And it updates the template partial to only show email /
title / phone numbers if the values are present *and* not `N/A`.


fec/fec/templates/partials/author-list.html
```diff
@@ -10,17 +10,19 @@
...
             {{ author.name }}
           {% endif %}
         </li>
-        {% if author.title != 'N/A' %}
+        {% if author.title and author.title != 'N/A' %}
         <li>{{ author.title }}</li>
         {% endif %}
-        <li>{{ author.phone }}</li>
+        {% if author.phone %}
+          <li>{{ author.phone }}</li>
```
jenniferthibault: @noahmanger Would we also want the phone field to display only if the values are present _and_ not `N/A`?

noahmanger: Not really an issue in this case because there's no phone numbers that have been set as `N/A`, as far as I know.

PR #801, opened 2017-02-03
Add support for and switch to the GovCloud environment

ccostino:
This changeset includes the necessary changes required to migrate the FEC CMS app to the GovCloud cloud.gov environment from the old East/West environment.  Some of these changes may end up being unwound later on, but for the time being these are the pieces that are required in order to get the app running in the new environment.


ccostino:
As part of this work I'm bumping the Python version to `3.5.2` - we've been running the app in Python 3.5.2 at least locally and in Travis as well to some degree, so I see no reason not to bump it up.

`3.6.0` is the latest version, which is also supported in cloud.gov now, but we haven't done testing with it yet.  Once the migration is done, we can investigate this separately to get up and running on the latest version.

ccostino:
I added the `cms-creds-stage` (and `cms-creds-prod`) service, so we should no longer see errors with the Travis builds.

LindsayYoung:
- make a mini-release
- review and merge
- check deploys in the tasks.py

jontours:
@LindsayYoung Just to ensure I'm understanding the steps here, it's first cut the release from the state dev is in now, and _then_ review and merge this branch.  I promise I listen during our meetings. ;).

fec/fec/settings/base.py
```diff
@@ -235,3 +235,18 @@
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_LOCATION = 'cms-content'
     AWS_S3_REGION_NAME = env.get_credential('CMS_AWS_DEFAULT_REGION')
+
+UAA_CLIENT_ID = env.get_credential('CMS_LOGIN_CLIENT_ID', 'my-client-id')
+UAA_CLIENT_SECRET = env.get_credential('CMS_LOGIN_CLIENT_SECRET', 'my-client-secret')
+#fake uaa server deploys locally on port 8080.  Will be needed to login for local use
+#TODO: These will have to have a explicit reference until we can figure out how
```
jontours: @ccostino Wanted to let you know I made this change, and to solicit your approval here.  In short, unless we can figure out how to silence the error in the tests, the urls for these variables have to be `https`, or the test errors out.  Maybe we can revisit later but for now I say we keep them as is, and add as environment variables later on in a pull request?

ccostino:
@jontours yes that's correct, we'll make a release off of `develop` as it is now before merging this in!

fec/fec/settings/base.py
ccostino: I think that's fine, thanks @jontours! 👍

ccostino:
Ah just to avoid confusion @jontours and @LindsayYoung - the branch in this PR needs to be merged into #897 - #897 is the one that should be merged into `develop` once our mini release is done!

PR #803, opened 2017-02-07
Feature/tips-improvements

noahmanger:
This adds a step to the import script for tips for treasurers that wraps the body in a single `<p>` tag and makes a couple small changes to the template. 

Also, I noticed that the `escape_quotes` method was replacing all single quotes with two single quotes. I changed the replace method to replace a single quote with an HTML entity code instead which got the desired effect, but let me know if that's wrong.

Resolves the issues already caught in https://github.com/18F/fec-cms/issues/773 but there may be more.


jenniferthibault:
The screenshots in https://github.com/18F/fec-cms/issues/773#issuecomment-278464060 look good to me! Signed off on the visual side ✅

fec/home/templatetags/updates.py
```diff
@@ -23,22 +24,29 @@ def weekly_digests():
 
 @register.inclusion_tag('partials/home-page-updates.html')
 def home_page_updates():
-    press_releases = PressReleasePage.objects.filter(homepage_hide=False, has_unpublished_changes=False).order_by('-date')[:4]
+    press_releases = PressReleasePage.objects.live().filter(homepage_hide=False, has_unpublished_changes=False).order_by('-date')[:4]
     if settings.FEATURES['record']:
-        records = RecordPage.objects.filter(homepage_hide=False).order_by('-date')[:4]
+        records = RecordPage.objects.live().filter(homepage_hide=False).order_by('-date')[:4]
```
xtine: Should also check for unpublished changes for both records and tips for treasurers so they won't show up on Latest Updates if they are in draft mode.

fec/home/templates/home/updates/tips_for_treasurers.html
```diff
@@ -49,12 +49,14 @@ <h1 class="heading__left">{{ self.title }}</h1>
 
 <div class="slab slab--neutral">
   <div class="container">
-    <div class="content__section">
-      <h3>Read next:</h3>
-      {% if self.read_next %}
+    <h3>Read next:</h3>
```
xtine: Should we also hide the Read Next text here if there isn't one?

fec/home/templatetags/updates.py
noahmanger: Isn't this what `objects.live()` does?

fec/home/templates/home/updates/tips_for_treasurers.html
noahmanger: 👍

fec/home/templatetags/updates.py
xtine: Oh snap, that's a wagtail specific queryset! Hm, then do we still need to show `has_unpublished_changes=False` on PRs as well?

noahmanger: Oh! Yeah, I don't think so. I'll remove it.

fec/home/templatetags/updates.py
```diff
@@ -23,22 +24,34 @@ def weekly_digests():
...
 def home_page_updates():
-    press_releases = PressReleasePage.objects.filter(homepage_hide=False, has_unpublished_changes=False).order_by('-date')[:4]
+    press_releases = PressReleasePage.objects.live().filter(homepage_hide=False).order_by('-date')[:4]
     if settings.FEATURES['record']:
-        records = RecordPage.objects.filter(homepage_hide=False).order_by('-date')[:4]
+        records = RecordPage.objects.live().filter(homepage_hide=False).order_by('-date')[:4]
+    if settings.FEATURES['tips']:
+        tips = TipsForTreasurersPage.objects.live().filter().order_by('-date')[:4]
     else:
         records = []
```
xtine: Move up 3 lines

PR #804, opened 2017-02-09
Feature/about-reports

noahmanger:
This adds support for the "About > Reports about the FEC" pages.  

image

I talked with @patphongs and @johnnyporkchops as part of https://github.com/18F/fec-cms/issues/779 yesterday and we identified that the method for adding these on the CMS needed to allow for organization within the Wagtail admin (i.e. we couldn't just add them as native Wagtail "Document" assets) as well as within the S3 bucket they live in.

This PR does that by making a new page model for `DocumentPage` which [looks like this](https://cloud.githubusercontent.com/assets/1696495/22766097/733575a6-ee28-11e6-9082-01b6f4485d8c.png). If the document is just a PDF, you include the S3 URL in the field and it gives you something like:

image

If the DocumentPage is actually an HTML page, it prints that out like:
image

These `DocumentPages` will be children of another new page model, `DocumentFeedPage`. This inherits from the same template as the latest updates page (which I refactored to extend a base template). The document feed page includes a year filter and then calls an inclusion tag which pulls in all of this pages child pages. 

Last, I added an loader script so that we could take a spreadsheet of S3 urls and with a little cleanup import them all into the system.


johnnyporkchops:
nice!

patphongs:
This is great! I will look up where I uploaded these docs to the bucket and add the new URL path to the spreadsheet. That way when you run your script, it will be pointing there instead.

noahmanger:
@ccostino it's the `import_report_documents.py` file at the top. It just parses json, so there will just need to be a manual step to convert the spreadsheet to json.

ccostino:
Ah okay - alternatively we could have the import script just parse the spreadsheet directly (if it's a simple thing to work with like just a CSV file), but whatever will be less work. :-)

I pulled this down locally and the migrations are working, the pages are available in the admin, and templates appear to be okay, so I will merge.  Thanks!

PR #806, opened 2017-02-09
Adding sub-category filters

noahmanger:
This adds the filters for the various report sub-categories. 

DocumentFeedPages must set a parent category which then pulls in a `<select>` with the appropriate child categories:
image

image

cc @ccostino


fec/fec/constants.py
```diff
@@ -132,17 +132,62 @@
...
+    'agency_operations': agency_operations_reports
+}
+
+# Create an OrderedDict of all the categories
+report_parent_categories = OrderedDict((x, x.replace('_', ' ')) for x in report_category_groups.keys())
+
+# Combine all the above categories into a single dict
+report_child_categories = {}
+for category in report_category_groups.keys():
+    report_child_categories.update(report_category_groups[category])
```
ccostino: Just making sure I understand this correctly, these are split apart due to the `category_filter` property below, is that correct?  So as you select a parent category, only the child categories will appear that are associated with it?

I was a bit confused by the separate of categories, only to have them combined together at the end; is this the reason for that?

fec/fec/constants.py
```diff
@@ -132,17 +132,62 @@
...
+
+report_category_groups = {
+    'oig': oig_reports,
+    'strategy_budget_performance': strategy_budget_performance_reports,
+    'foia': foia_reports,
+    'privacy': privacy_reports,
+    'procurement_contracting_reports': procurement_contracting_reports,
+    'annual_anniversary': annual_anniversary_reports,
+    'agency_operations': agency_operations_reports
+}
```
noahmanger: It's for that and also to provide a reference here https://github.com/18F/fec-cms/pull/806/files#diff-c7dd542c74f42ef1820b79640c812d87R356

ccostino:
Thanks for going over this with me, @noahmanger!  Much clearer now and the documentation will help; if we can come up with a better solution in the future we can revisit. :-)

PR #825, opened 2017-02-21
Building reports landing page

noahmanger:
This adds the landing page for the "Reports about the FEC" section:

image

It's an editable CMS page in which you add a link to a sub-page and a short blurb:
image

The [designs](https://cloud.githubusercontent.com/assets/11636908/21940543/e0b98d1a-d992-11e6-8eb1-1b1368e12f07.png) called for showing the number of results. This could be either done by hand or automatically. Doing it by hand seems tedious. Doing it automatically requires 7 separate queries to the database to see how many reports of each type there are. 

Q: Curious sure if folks have strong feelings about the need to include it or not, and if so, what the best way is. cc @jenniferthibault @xtine .

Requires https://github.com/18F/fec-style/pull/635 for styles

Resolves https://github.com/18F/fec-cms/issues/778


jenniferthibault:
I should have read this before commenting in full on https://github.com/18F/fec-style/pull/635#issuecomment-281444883 about the # of documents, sorry @noahmanger !

The intent of that feature is to give the user an idea of how many documents there are in the category before they dive in. It could be advantageous to the user if they know how deep the hole is before they start digging— @emileighoutlaw was enthusiastic about this feature when it was proposed, so I'd wonder what the functional trade-off of the seven queries is. In other words, I'm not sure what 7 database queries means for user experience. Can you help translate a bit @noahmanger ?

noahmanger:
@jenniferthibault discovered that wasn't really hard so I added the counts (and also underlining the title on hover): 

image

jenniferthibault:
That is exciting and awesome! Thanks for the tweaks

emileighoutlaw:
Just echoing Jen that this is super exciting! Thanks @noahmanger!

Issue #743, opened 2017-01-17
Database migration cleanup


xtine:
I think it would be helpful if could assess the current state of DB migrations for the CMS. Maybe decide on squashing migrations if we want a clean slate to prevent past references that don't exist?

We should also look at the `django_migrations` table on both staging and prod if they match up with the migrations we actually do have in the repo.

Tagging @ccostino and @LindsayYoung to help in this effort.

ccostino:
Agreed, @xtine!  I started looking into this a bit and it doesn't look too bad, but I didn't want to just do it either.  I'm more than happy to work with you and anyone else interested on this. :-)

xtine:
@ccostino: let's plan a pairing session, I would like to get more confident with these DB problems.

ccostino:
Perfect!  I'll find time on our calendars. :-)

ccostino:
@noahmanger:  After reviewing more of the squashing mechanisms and what's involved with them, @xtine and I feel fairly confident that we should be able to move forward with this next sprint.  With the work on the scraping and importing being almost complete in this sprint, we should be a in a good spot where there isn't as much churn in the database with new models being added and other adjustments being made.

Furthermore, if we set things up to get this working at the start of next sprint, we can coordinate with everyone and actually do the squashing and initial testing while we're all in town together.  This would be a great learning and knowledge sharing opportunity for the FEC folks.

If this sounds good to you, we can start working out some specific dates and the people that should be involved and plan from there!  Please let us know if you have any questions or if anything is unclear.

noahmanger:
Sounds good to me. Thanks!

ccostino:
@xtine and I tried to go through the exercise of squashing the migrations with folks at the CMS today and made some progress, but there's more work to be done.  The squashed migration required one manual edit after the fact (which the migration framework indicated to us was the case!), and seems to work fine if you are working in an existing database.

However, when we tested the squashed migrations on a new database, we ran into an issue that we're currently unsure of.  The same issue occurs in Travis as well since each build starts with an empty database.  The current work is found in PR #826.

ccostino:
We're still working through this, more detailed notes are in the PR itself (#826).

ccostino:
We've postponed work on this as the squashing was not working and happening at a time when there were an influx of new migrations.  We may resume this in the future once things settle down again.

PR #826, opened 2017-02-21
[WIP] Squash CMS home app migrations

ccostino:
**WIP:  PLEASE DON'T MERGE**

Addresses #743

This changeset is an attempt to clean up our migrations and start from scratch with them.  Our app has grown significantly and the state of the database is very different from when we started.  We hope that in doing this it provides a cleaner base to start from now that we have gotten through a lot of heavy feature development.


ccostino:
Well, we're still trying to figure out what's going on with this, but we've uncovered where the problem is at least!  It is the migration that adds the `keywords` field to the `RecordPage` models, which is the `0034_auto_20161221_1920` migration.  What is happening is that the migration processing gets down to the point of adding the field itself, and encounters this line (line 392 in `django/db/backends/base/schema.py` within the Django migration framework):

```python
if field.many_to_many and field.remote_field.through._meta.auto_created:
```

`field.remote_field.through` is returning just a string (`home.RecordPageTag`), which does not have a `_meta` attribute on it like a model or object would.  We're trying to figure out what the breakdown is here and why this is the case.  I've asked in the #django Slack channel and we'll continue to do some more research into it.

ccostino:
Also, why this works with our existing migrations but not with this new squashed migration on a fresh database is still a mystery.

LindsayYoung:
Maybe this is Django blasphemy, but what if we look at a db that is up to date, delete all the migration files locally. Then, make a new initial migration, add the "replaces" section  and then name it fec/home/migrations/0001_squashed_0065_auto_20170221_2055.py

ccostino:
🤔 it's cross my mind too, I think it's worth a shot at this point just to see.

ccostino:
I recommend closing this PR and if we decide to come back to it we start fresh, how do others feel?

noahmanger:
Sounds good.

PR #832, opened 2017-02-24
Adding a deploy for gov cloud

LindsayYoung:
so we can use ci and keep the branch separate


ccostino:
Thanks, @LindsayYoung!

PR #834, opened 2017-02-25
Adding beautify-html.js to vendor scripts

noahmanger:
As @johnnyporkchops noted in https://github.com/18F/fec-cms/issues/830, in production, the Wagtail editor isn't able to find the `beautify-html.js` script, which is preventing the "edit HTML" button from working. It _should_ be able to locate this inside `node_modules/js-beautify/js/lib/`, as it does when developing locally, but for some reason it's not. 

I'm stumped, so here's a workaround. This just moves that single script to the `/static/js/vendor/` directory. It works locally for me, and seems like a safer bet that it should work in production. 

Definitely open to other ideas though.

### DIFF
--- a/fec/fec/static/js/vendor/beautify-html.js
+++ b/fec/fec/static/js/vendor/beautify-html.js
@@ -1,0 +1,952 @@
+/*jshint curly:true, eqeqeq:true, laxbreak:true, noempty:false */
+/*
+
+  The MIT License (MIT)
+
+  Copyright (c) 2007-2013 Einar Lielmanis and contributors.
+
+  Permission is hereby granted, free of charge, to any person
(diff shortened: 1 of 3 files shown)


noahmanger:
Now that we have the feature env back, we should test this by deploying to there and seeing if it solves it.

patphongs:
@noahmanger Are you able to check whether or not js-beautify is within node_modules on prod? If it's not there, maybe it's worth it to run `npm -g install js-beautify` to explicitly install it. If that doesn't solve it, this PR is our best bet for now I think.

noahmanger:
Just deployed this to `feature` and it looks like it did the trick.

noahmanger:
I think this solved the problem. Should we just merge?

PR #856, opened 2017-03-08
Report importing

noahmanger:
This is the final piece necessary before we can import the reports about the FEC to resolve https://github.com/18F/fec-cms/issues/828

- Adds JSON files for each of the sets of reports
- Adds a new property to the `DocumentPage` model for `display_date` which returns either `month year` or just `year` as some reports have a `year_only` flag because we don't want them associated with a specific month
- Adds a new property for showing the file extension, rather than hardcoding it into the template (so that XML and PDF extensions are shown for the correct docs


ccostino:
Is this ready for review, @noahmanger?

noahmanger:
@ccostino yep!

fec/home/migrations/0067_documentpage_year_only.py
```diff
@@ -0,0 +1,20 @@
...
+        ('home', '0066_remove_documentpage_file_name'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='documentpage',
+            name='year_only',
+            field=models.BooleanField(default=False),
+        ),
+    ]
```
ccostino: Just a heads up that we will have another situation where we'll have to merge migrations again.  I saw some migrations in `develop` that began with `0066` and `0067`.  Not an issue in this case, though!

PR #857, opened 2017-03-08
Feature/meetings scraper

dogweather:
This is the code for the scraper only. It's completed.


README.md
```diff
@@ -150,6 +150,13 @@ cd fec/
 ./manage.py test
 ```
 
+It's necessary to specify the Postgresql URL, which can be done on the
+command line, e.g.:
+
+```bash
+env DATABASE_URL=postgresql://:@/cfdm_cms_test ./manage.py test
+```
+
```
ccostino: Just a heads up that these [instructions for the `DATABASE_URL`](https://github.com/18F/fec-cms/blob/develop/README.md#create-local-databases) already exist in the README just a little further up, specifically [line 113](https://github.com/18F/fec-cms/pull/857/files#diff-04c6e90faac2675aa89e2176d2eec7d8R113). :-)

fec/data_loader/management/commands/meetings-scraper-notes.txt
```diff
@@ -0,0 +1,35 @@
...
+
+
+for meeting in goodmeetings:
+    for slink in meeting.sunshine_act_links:
+        sunshines.append(slink.url)
+usunshines = set(sunshines)
+
+
+
+b'<tr>\r\n       <td>April 15, 2004<br>\r\n      (Public Hearing)</td>\r\n       <td><a href="http://www.fec.gov/pdf/nprm/political_comm_status/trans_04_15_04.pdf">Transcript Available</a></td>\r\n       <td align="center"><a href="http://www.fec.gov/sunshine/2004/04-8429.pdf">Notice</a></td>\r\n     </tr>\r\n     '
```
ccostino: I'm guessing we don't need to include this file as a part of this PR?  This looks like a scratch pad of notes for local testing.  If you think it'd be useful for others, though, would it be possible to format and structure it better to be of more help?

.gitignore
```diff
@@ -66,3 +68,10 @@ fec/media/
 fec/data_loader/data/**/*.json
 
 fec/media
+
+# PyCharm
+.idea
+
+# Parsing output
+fec/data_loader/management/commands/meetings--*.json
+fec/data_loader/management/commands/out/*.html
```
ccostino: Nit-pick:  No blank new line at the end of the file (PyCharm, Sublime, Atom, etc., all have configuration options to automatically add it if one isn't already there).

.python-version
```diff
@@ -0,0 +1 @@
+3.5.3
```
ccostino: Just noting that this file is used for local development, but won't be used by cloud.gov at all (the `runtime.txt` file is used instead).  I'm not sure it's worth including, either, as it may introduce some odd behavior for some folks depending on their local setup.

fec/data_loader/management/commands/meetings-scraper-notes.txt
dogweather: I opted to leave it as is because it's @tadhg-ohiggins 'd notes. It may turn out that we need to change the info we're scraping, and it will be valuable.

.python-version
dogweather: I think it'd be good to lock down the Python version. Or do you think we should leave the python version open?

.gitignore
dogweather: thanks!

README.md
dogweather: Yep, it's a little redundant, but `DATABASE_URL` isn't namespaced with some app name, so I don't/can't set it as a shell env var. I added it because I like to keep that magic invocation that just works, since I multitask between projects.

.python-version
ccostino: Agreed, and we do lock it down for all of the environments we build/deploy to.  I'm less inclined to do so locally, though.  We do have this [written in the README](https://github.com/18F/fec-cms/blob/develop/README.md#project-prerequisites) in terms of what version(s) is required, but `.python-version` is [Pyenv specific](https://github.com/pyenv/pyenv#choosing-the-python-version) as far as I know, and we do not impose the use of Pyenv as a requirement on anyone or as a prerequisites for working on the project (we simply suggest it as tool that might be useful for folks).

fec/data_loader/management/commands/meetings-scraper-notes.txt
ccostino: Okay, makes sense!  If we're going to include it as a part of this though, is there a way to organize them just a bit better to provide some context for folks looking at them and/or quick instructions on what to do with them?

.python-version
dogweather: IMO locking it down between dev and production is the most important place to. I don't see this as a place where there's a benefit to allowing inadvertent version mismatches. ?

Since we don't require pyenv, `.python-version` will only affect those who opt in.

ccostino: True, so my only other question then is what happens when someone is using Pyenv but doesn't have `3.5.3` installed?  Will it error out and inform them to install the correct version of Python or automatically do that on the user's behalf?  Or something else?

dogweather: I don't know, but since it's a fork of the ruby equivalent, I'm guessing that it will error out and tell them they need to install 3.5.3. That's the desired behavior, IMO. I.e., fail fast and clearly. That's the best user experience.

fec/data_loader/management/commands/meetings-scraper-notes.txt
dogweather: @tadhg-ohiggins ^^ ?

.python-version
dogweather: Ah hah, `runtime.txt` is `3.4.5`. I'm going to set the dev environment to that version, re-run the tests, and commit it.

fec/data_loader/management/commands/meetings-scraper-notes.txt
tadhg-ohiggins: I'll do a very rough cleanup and submit to this branch.

dogweather: Thanks @tadhg-ohiggins !

.python-version
dogweather: NB: I found that our Travis config was out of sync with production as well.

dogweather:
Thanks @ccostino

PR #861, opened 2017-03-09
Added blockquote functionality for wagtail editor

patphongs:
Implementation of issue https://github.com/18F/fec-style/issues/632

screen shot 2017-03-08 at 11 40 38 pm

### DIFF
--- a/fec/fec/static/css/customize-editor.css
+++ b/fec/fec/static/css/customize-editor.css
@@ -17,3 +17,7 @@ li.sequence-member .sequence-member-inner .sequence-type-list .sequence-containe
 .t-sans {
   font-family: sans-serif;
 }
+
+.icon-blockquote:before {
+  content: open-quote;
+}
--- a/fec/fec/static/css/customize-editor.css
+++ b/fec/fec/static/css/customize-editor.css
@@ -18,6 +18,9 @@ li.sequence-member .sequence-member-inner .sequence-type-list .sequence-containe
   font-family: sans-serif;
 }
 
-.icon-blockquote:before {
-  content: open-quote;
+blockquote {
+	 border-left: 10px solid #f1f1f1;
+	 padding-left: 30px;
+	 margin-left: 0;
+	 font-style: italic;
 }
--- a/fec/fec/static/js/hallo-blockquote.js
+++ b/fec/fec/static/js/hallo-blockquote.js
@@ -1,0 +1,44 @@
+/* Blockquote button for hallo.js
+ * Adapted from https://gist.github.com/SalahAdDin/347e4fab78a64eaadd5c
+ */
+ 
+(function() {
+  (function($) {
+    return $.widget("IKS.blockquotebutton", {
+      options: {
+        uuid: '',
+        editable: null
+      },
+      populateToolbar: function(toolbar) {
+        var button, widget;
+ 
+        widget = this;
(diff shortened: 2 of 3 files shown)


patphongs:
@jenniferthibault style comments have been addressed in this PR: https://github.com/18F/fec-style/pull/645

fec/fec/static/js/hallo-blockquote.js
```diff
@@ -0,0 +1,44 @@
...
+      populateToolbar: function(toolbar) {
+        var button, widget;
+ 
+        widget = this;
+        button = $('<span></span>');
+        button.hallobutton({
+          uuid: this.options.uuid,
+          editable: this.options.editable,
+          label: 'Blockquote',
+          icon: 'icon-blockquote',
```
noahmanger: You could just set `icon-openquote` which will use Wagtails default icons and then you don't need to add the `.icon-blockquote` style above.

fec/fec/static/js/hallo-blockquote.js
```diff
@@ -0,0 +1,44 @@
...
+          command: null
+        });
+        toolbar.append(button);
+        return button.on("click", function(event) {
+          var insertionPoint, lastSelection;
```
noahmanger: In `hallo-sans-serif.js` there's a slightly simpler version of this click handler that might work. 

Also, there should be a way to set the "edited" state on the blockquote button to show that it's had an effect.

fec/fec/static/css/customize-editor.css
```diff
@@ -17,3 +17,7 @@ li.sequence-member .sequence-member-inner .sequence-type-list .sequence-containe
 .t-sans {
   font-family: sans-serif;
 }
+
```
noahmanger: What do you think about adding a blockquote style here to add the border and/or italics so that it's extra clear in the editor when it's been applied?

jenniferthibault: ^ something that shows it's been applied in the editor would be SUPER fantastic! It's hard to tell when styles have been applied already (Jeremy and Dorothy have mentioned the same)

patphongs: Thanks, I've added the blockquote styles to here to more closely match changes on the published page.

screen shot 2017-03-09 at 8 07 02 pm

fec/fec/static/js/hallo-blockquote.js
patphongs: Thanks for showing me where the wagtail styleguide is. I've changed the class reference.

fec/fec/static/js/hallo-blockquote.js
patphongs: Thanks, that was a lot simpler. Also, it wraps the p tags within the blockquote, which was nice as I did not have repeat the font size now. cc @jenniferthibault :)

noahmanger:
Nice!

fec/fec/static/css/customize-editor.css
jenniferthibault: I AM SO EXCITED!

jenniferthibault:
Really nice work on this @patphongs ! Not only the technical side, but the communication & GitHub workflow too. & Awesome review @noahmanger that flagged really helpful improvements for people entering content on Wagtail. 🎉 🎉 teamwork!

PR #863, opened 2017-03-09
remove Ravel from homepage

johnnyporkchops:


noahmanger:
Does this need to be deployed to production ASAP?

johnnyporkchops:
I don't think so

AmyKort:
Sorry, weighing in late. Yes. It needs to be deployed to production as fast as possible.

AmyKort:
My apologies to John for being a slow communicator.

noahmanger:
Since this is going into production, this should be branched off of `master` rather than `develop`. Closing this to be replaced be a clean hotfix.

PR #868, opened 2017-03-13
Import Meeting Agenda

dogweather:
The scraper creates a JSON file containing of list of objects with these attributes:

```
            ("agenda_document_links", Links), # list (Links)
            ("approved_minutes_date", Date),
            ("approved_minutes_link", Link),
            ("body", str),                    # HTML
            ("closed_captioning_link", Link),
            ("draft_minutes_links", Links),   # list (Links)
            ("link_title_text", str),         # Plain text
            ("meeting_type", str),            # "open" or "executive"
            ("pdf_disclaimer", str),          # HTML (outer HTML)
            ("posted_date", Date),
            ("primary_audio_link", Link),
            ("old_meeting_url", str),         # URL
            ("secondary_audio_links", Links), # list (Links)
            ("sunshine_act_links", Links),    # list (Links)
            ("title_text", str),
            ("video_link", Link)
```

This code imports these by instantiating [`AgendaPage`](https://github.com/18F/fec-cms/blob/develop/fec/home/models.py#L664-L698) models.


fec/data_loader/management/commands/import_meeting_agendas.py
```diff
@@ -0,0 +1,144 @@
...
+        if options['delete_existing']:
+            self._log('Deleting existing records...')
+            self.delete_existing_records(AgendaPage, **options)
+
+    @staticmethod
+    def _open_json_file(options: Dict[str, Any]) -> TextIOWrapper:
+        return open(options['json_file_path'], 'r')
+
+    def _log(self, message: Any) -> None:
+        self.stdout.write(repr(message))
```
ccostino: Ah, nice seeing this method split out, keeps the logging statements cleaner elsewhere!

fec/data_loader/management/commands/import_meeting_agendas.py
```diff
@@ -0,0 +1,144 @@
...
+        with self._open_json_file(options) as json_text:
+            self._create_pages(json_text, self._parent_page(options), options)
+
+
+    def _delete_existing_records(self, options: Dict[str, Any]) -> None:
+        if options['delete_existing']:
+            self._log('Deleting existing records...')
+            self.delete_existing_records(AgendaPage, **options)
+
+    @staticmethod
```
ccostino: Out of curiosity, why make these (and several other methods) `staticmethod`s?  There's nothing wrong with it and no need to change anything.  I am just curious of the thought process/intention behind it!

fec/home/models.py
```diff
@@ -688,14 +688,23 @@ def content_section(self):
...
         return 'services'
 
+
 class AgendaPage(Page):
     mtg_date = models.DateTimeField(default=datetime.date.today)
-    mtg_time  = models.TimeField(default=datetime.time(10, 00))
+    mtg_time = models.TimeField(default=datetime.time(10, 00))
+
+    imported_html = StreamField(
+        [('html_block', blocks.RawHTMLBlock())],
```
ccostino: Do we want this as the raw HTML or in a rich text block?  @noahmanger, any additional insight?

fec/data_loader/management/commands/import_meeting_agendas.py
```diff
@@ -0,0 +1,144 @@
+from dateutil import parser
+from io import TextIOWrapper
+import json
+from typing import Any, Dict, Optional
```
ccostino: I see that the tests have passed, but the `typing` module is in Python 3.5+, and the CMS is still running in 3.4.  Will this cause an issue?  Kind of odd that the tests would pass if it can't import the module, so I'm guessing I'm missing something.

dogweather: That's a good catch, and weird that the tests still pass. I think that the typing module was sort of stealth included in earlier versions of Python, mostly as an internal api for testing before being announced.

I'll remove the import and adjust the code.

fec/home/models.py
dogweather: Noah and I talked about this, and I'm pretty sure this is what we landed on: a catch-all field which will only be populated for the legacy imported pages. And implemented as raw html. (?)

fec/data_loader/management/commands/import_meeting_agendas.py
dogweather: My idea is that the ideal is having as many pure functions as possible, relying on state (`self`) only when abs. necessary. This should simplify the code, making it easier to understand and debug. The irony is that the `@staticmethod` annotation is somewhat noisy, as well as the invocation sites which still must be prefixed with `self.`. I believe that the same thing might be able to be achieved, in a quieter manner, with functions (not methods) on the module. Here, though, I stayed with methods everywhere because that was the current coding style.

fec/data_loader/management/commands/import_meeting_agendas.py
dogweather: Thanks!

dogweather:
About testing:

I verified the code is working by editing the pages in the admin ui. I wasn't able to view them in their normal public view, probably because of the state of my database, and my not using the correct parent page path.

fec/data_loader/management/commands/import_meeting_agendas.py
ccostino: Ah okay, that makes sense regarding checking state vs. not.  Thanks for explaining!

ccostino:
Okay - @noahmanger, what is the proper path for these?  Is it it the same as all of the others, `/home/updates/`?

noahmanger:
Yeah, that's the path for them, though the `updates` view in `view.py` will need to be updated to actually query them.

ccostino:
Oh, I think I see where you mean, [here](https://github.com/18F/fec-cms/blob/develop/fec/home/views.py#L56)?  Is that something that should be included with this PR, then?

dogweather:
I was just reading through that code, but don't quite understand what this function does, and how it needs to be modified.

noahmanger:
Yes, but sorry, I meant to say that should be a follow up task. (And yeah, in the process that function should be cleaned up).

ccostino:
Ah okay, thanks!  It sounds like we're good to go as is here with moving forward for importing?

noahmanger:
Yep! If it's working correctly.

fec/home/templates/home/agenda_page.html
```diff
@@ -27,6 +27,10 @@ <h1 class="heading__left">{% formatted_title self %}</h1>
         </div>
       </header>
    <div class="main__content">
+       <div>
+           {% self.imported_html  %}
```
ccostino: I was testing the importing and discovered that this requires the use of a Wagtail template tag: [`include_block`](http://docs.wagtail.io/en/v1.9/topics/streamfield.html#template-rendering)

If this line is changed to this:

```python
{% include_block self.imported_html %}
```

The page will render in both the preview mode and the live published page. :-)

ccostino:
Okay!  Aside from a very minor template change we should be good to go with this. :-)

@dogweather I made a note of the change to the template; @noahmanger should I import the content into production as soon as this is merged or wait until the next release?

ccostino:
One thing I'm noticing with a random sampling of the pages imported is that none of them have video links it appears.  I've only looked at a handful or so, but is this to be expected or is there a bit of a content audit we should do to make sure all pieces of the content are in place as expected?

ccostino:
I was able to do a bit of testing and came up with this list of meetings that have meeting videos associated with them:

```python
('March 10, 2015', 'https://fec.adobeconnect.com/p5a22y70wqp/'),
('March 19, 2015', 'https://fec.adobeconnect.com/p14tqjhban5/'),
('July 16, 2015', 'https://fec.adobeconnect.com/p3432gocxrd'),
('August 11, 2015', 'https://fec.adobeconnect.com/p1nz839nmjj/'),
('September 17, 2015', 'https://fec.adobeconnect.com/p14he4ftpo5/'),
('October 1, 2015', 'https://fec.adobeconnect.com/p97uvyzdpga/'),
('October 29, 2015', 'https://fec.adobeconnect.com/p8e2zhbm4az/'),
('November 10 and 17, 2015', 'https://fec.adobeconnect.com/p9ihqli5apz/'),
('December 17, 2015', 'https://fec.adobeconnect.com/p9k9licpypf/'),
('January 14, 2016', 'https://fec.adobeconnect.com/p9ihsorpnwc/'),
('February 11, 2016', 'https://fec.adobeconnect.com/p55tgwpu7fd/'),
('April 14, 2016', 'https://fec.adobeconnect.com/p42ts63hq84/'),
('June 30, 2016', 'https://fec.adobeconnect.com/p1oc509ado5/'),
('July 14, 2016', 'https://fec.adobeconnect.com/p8uqv4qpml2/'),
('August 16, 2016', 'https://fec.adobeconnect.com/p3c8r38lwtq/'),
('September 15, 2016', 'https://fec.adobeconnect.com/p24zxihtv1h/'),
('September 29, 2016',
  'https://fec.adobeconnect.com/\nhttps://fec.adobeconnect.com/p7qowpuxrpk/'),
('October 27, 2016', 'https://fec.adobeconnect.com/p460p8k0ba7/'),
('November 17, 2016', 'https://fec.adobeconnect.com/p31lytx2ugh/'),
('December 1, 2016', 'https://fec.adobeconnect.com/p5lrtcwxg4k/'),
('December 6, 2016\n        (Audit Hearing)',
 'https://fec.adobeconnect.com/p9iuo3dvo1p/'),
('December 8, 2016', 'https://fec.adobeconnect.com/p84im8p2sop/'),
('January 12, 2017', 'https://fec.adobeconnect.com/p9thiepr0jr/'),
('January 25, 2017', 'https://fec.adobeconnect.com/p5as91xpke3/'),
('March 9, 2017', 'https://fec.adobeconnect.com/p3l3vfr7jma/')
```

That's 25 agenda pages out of 431 that were imported (and from just 2015 on).  If this seems right or is expected, then so be it!  I just wanted to double check.

ccostino:
One other quick thing that comes to mind - if this is all okay with the videos, should we just wrap that section of the template in a conditional to check for the presence of one, and if it's not there, not display that section on the page?

ccostino:
Merging!

PR #888, opened 2017-03-15
Feature/efiling report

noahmanger:
This adds support for including `ResourcePages` as children of `DocumentFeedPages`, which allows something like the efiling study to be included in "Reports about the FEC".

image

The remaining to-do is to add the correct category for this report. Do you know what that should be @emileighoutlaw / @AmyKort ?


noahmanger:
Added the "Study" category as an option for "Agency operations" reports so this is ready to go.

fec/home/templatetags/document_feed.py
```diff
@@ -1,9 +1,29 @@
...
+from home.models import DocumentPage, ResourcePage
+from wagtail.wagtailcore.models import Page, Orderable
 
 register = template.Library()
 
+def get_documents(page, year='', category=''):
+    documents = DocumentPage.objects.child_of(page).live().order_by('-date')
+
+    if year:
+      documents = documents.filter(date__year=year)
```
xtine: check indent

noahmanger: Looks right to me?

xtine: It looks like two spaces vs four spaces.

noahmanger: Oh oops. Fixing!

LindsayYoung:
Thank you for doing this!

PR #889, opened 2017-03-15
Enable 2FA [WIP]

jontours:
Currently working on getting Django to redirect to CF login.  Once that done should be close to being done.

### DIFF
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -235,3 +235,9 @@ if FEC_CMS_ENVIRONMENT != 'LOCAL':
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_LOCATION = 'cms-content'
     AWS_S3_REGION_NAME = env.get_credential('CMS_AWS_DEFAULT_REGION')
+
+UAA_AUTH_URL = 'http://localhost:8080/oauth/authorize'
+UAA_TOKEN_URL = 'http://localhost:8080/oauth/token'
+
+AUTHENTICATION_BACKENDS = \
+    ['uaa_client.authentication.UaaBackend']
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -62,6 +62,8 @@ INSTALLED_APPS = (
     'search',
     'home',
     'data_loader',
+
+    'uaa_client',
      
 )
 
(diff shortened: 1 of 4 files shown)


fec/fec/settings/base.py
```diff
@@ -235,3 +237,13 @@
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_LOCATION = 'cms-content'
     AWS_S3_REGION_NAME = env.get_credential('CMS_AWS_DEFAULT_REGION')
+
+UAA_CLIENT_ID = env.get_credential('CLIENT_ID', 'my-client-id')
+UAA_CLIENT_SECRET = env.get_credential('CLIENT_SECRET', 'my-client-secret')
+UAA_AUTH_URL = 'http://localhost:8080/oauth/authorize'
+UAA_TOKEN_URL = 'http://localhost:8080/oauth/token'
```
ccostino: Should we move the `UAA_AUTH_URL` and `UAA_TOKEN_URL` values here to be in the credential service as well?

fec/fec/settings/base.py
```diff
@@ -235,3 +237,13 @@
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_LOCATION = 'cms-content'
     AWS_S3_REGION_NAME = env.get_credential('CMS_AWS_DEFAULT_REGION')
+
+UAA_CLIENT_ID = env.get_credential('CLIENT_ID', 'my-client-id')
+UAA_CLIENT_SECRET = env.get_credential('CLIENT_SECRET', 'my-client-secret')
```
ccostino: I noticed that these two were set as `CMS_LOGIN_CLIENT_ID` and `CMS_LOGIN_CLIENT_SECRET` in the credential service for the CMS, will these need to change?

fec/fec/templates/uaa_client/login_error.html
```diff
@@ -0,0 +1,10 @@
+<h1>Login Error</h1>
+{% if error_code == 'authenticate_failed' %}
+  <p>
+    Either you don't have permission to log in to this system, or
+    an error occurred when trying to log you in.
+  </p>
+{% else %}
+  <p>Unfortunately, an error occurred when trying to log you in.</p>
+  <p>Error code: <code>{{ error_code }}</code></p>
+{% endif %}
```
ccostino: Nice to have this so folks know when something didn't go right!

fec/fec/urls.py
```diff
@@ -22,8 +25,13 @@
     url(r'^registration-and-reporting/question-RAD/$', home_views.contact_rad),
     url(r'^search/$', search_views.search, name='search'),
     url(r'^updates/$', home_views.updates),
+    #uaa_client/login_error.html
+
 
     url(r'', include(wagtail_urls)),
+
+    url(r'^auth/', include('uaa_client.urls')),
```
ccostino: This URL pattern will never be matched due to the one defined up above on line 18 - Django URL patterns [are read straight down](https://docs.djangoproject.com/en/1.10/topics/http/urls/#how-django-processes-a-request) and the first pattern to match a given route is going to be processed.  I'm not sure if this was intentional or not, but if you're seeing weird behavior and/or not going to where you expect, this could be why.

fec/fec/settings/base.py
ccostino: This was in the `dev` GovCloud space, so if you were also working elsewhere, that could by why too.

jontours: Good catch!, in the branch that this was merged into (move-dev-to-gov-cloud or something like that), these were changed to the right keys.

fec/fec/settings/base.py
jontours: That makes sense to me, I'll take care of that.

fec/fec/urls.py
jontours: Ah, that left in by accident (and I'm glad Django operates exactly how I'd expect in this instance)

PR #890, opened 2017-03-15
Fix/debugging form

noahmanger:
This fixes the contact form by using `env.get_credential()` instead of `os.getenv()` to get the necessary variables.


LindsayYoung:
This looks right, thanks @noahmanger

PR #892, opened 2017-03-16
Home page: Directs Mission link to beta site instead of old site

jenniferthibault:
Since we've released the Mission and history page, this link no longer needs to go to the old FEC site. 

screen shot 2017-03-16 at 10 58 35 am

cc @emileighoutlaw


emileighoutlaw:
Awesome! Good catch, Jen!

Issue #813, opened 2017-02-11
Make data table block in Wagtail


noahmanger:
So that the content managers can present stats in posts cleanly, implement a new data table component.

---
Carrying over from https://github.com/18F/fec-cms/issues/550, the task here is to implement a custom table component that can be used to show tables of data like:

![image](https://cloud.githubusercontent.com/assets/1696495/22849046/3924d992-efae-11e6-8d6c-7cbc751bab64.png)

**Completion criteria:**
- [x] New custom table block in `blocks.py`. This should be an instance of the Wagtail `StructBlock` that supports two sub-blocks: a title and a table. 
- [x] Table block can be added to the `stream_factory` in `models.py` so it can be available to all page types that use that
- [x] The block should render with a custom template that contains special classes so it can be styled uniquely
- [x] The header row and header column should be bold
- [x] Every non-header cell should be right-aligned for easy comparison of numbers

jenniferthibault:
Pulling from https://github.com/18F/fec-cms/issues/550#issuecomment-277839357 , the image below shows how the current alternating `paragraph block` + `table block` + `paragraph block` + `table block` (left) compares to the changes we're proposing on the right, which would be a `table block` with some built in surrounding text fields (right)

![screen shot 2017-02-06 at 5 39 02 pm](https://cloud.githubusercontent.com/assets/11636908/22669601/6148369e-ec93-11e6-9614-83656c2b5351.png)

Here's a few things that I'd add to @noahmanger's criteria above  (pulled from the pink in-line notes). 

- [ ] Need ability to have complex Paragraph style headers close to table—optional ability to give the table a title within that block

- [ ] Set minimum top/bottom margin between text and row edge

- [ ] A field in the table block for footnotes about the table (using the `t-note` class)

AmyKort:
Just checking in on this. Because this allows us to have the complete set of our press releases up, it's a pre-flip task.

noahmanger:
@AmyKort in progress and due to be released next week. We're just sussing out some details.

AmyKort:
thank you!!

noahmanger:
Resolved by https://github.com/18F/fec-cms/issues/901

PR #901, opened 2017-03-20
blocks.py, models.py, custom_table.html

johnnyporkchops:
Created a custom table block so a content editor can add statistical press release tables and also use for creating other table and text sections as needed. Issue : 813
#813

This follows [this design](https://cloud.githubusercontent.com/assets/11636908/22669601/6148369e-ec93-11e6-9614-83656c2b5351.png) as closely as possible while also balancing the concern that:
1) The exact structure of these tables will vary
2) It should be useful for creating other similar content sections as needed
3) It will be an available block (within the ubiquitous streamfactory) on several templates and if content editors use it-- it should be specific enough to create the statistical press releases but not so rigid that it is cumbersome to use it for other table/text combined content use-cases.

It is setup as a streamblock with the option to add more that one table for each "custom_table section", allowing the ability to stack tables closely within one section or put them into separate horizontally- spaced sections. (They can even stack adjacent tables with no space between to achieve the "mid-table header" format that some of the legacy site tables seem to have, although @jenniferthibault's  [design ](https://cloud.githubusercontent.com/assets/11636908/22669601/6148369e-ec93-11e6-9614-83656c2b5351.png)departs from that, thankfully.)

Adding corresponding styles to:
/scss/elements/_tables.scss
https://github.com/18F/fec-style/pull/649

Screenshot of wagtail page:
scrnsht


fec/home/blocks.py
```diff
@@ -147,3 +147,25 @@ class DocumentFeedBlurb(blocks.StructBlock):
     """For generating a box with a description that links to a document feed page"""
     page = blocks.PageChooserBlock()
     description = blocks.CharBlock()
+
+class CustomTableBlock(blocks.StructBlock):
+    """A custom table"""
```
xtine: Can you expand on this comment, like for example what kind of pages or data this table might be used for.

fec/home/templates/blocks/custom_table.html
```diff
@@ -0,0 +1,23 @@
+{% load wagtailcore_tags %}
+
+<div class="content__section--extra custom_table">
+{% for block in self.custom_table %}
+
+  {% if block.block_type == "title" %}
+    <h4>{{ block }}</h4>
+  
+  {% elif block.block_type == "table" %}
```
xtine: Can you combine the `block.block_type` check for both `table` and `table_intro` into one line since they give the same output?

fec/home/blocks.py
johnnyporkchops: Will expanp on comment

fec/home/templates/blocks/custom_table.html
johnnyporkchops: Sorry, I am not sure what you mean by the above comment, can you please elaborate?

xtine: `{% elif block.blocktype == "title" or block.blocktype == "table" %}`

johnnyporkchops: OK, now I see what you mean, they don't give the same output, they both use the same tag syntax to generate their output. Yes can I change that as now that I am fairly sure that I will that edits will not require wrapping either in an html block .

fec/home/blocks.py
```diff
@@ -166,3 +187,24 @@ class ExampleForms(blocks.StructBlock):
     class Meta:
         template = 'blocks/example-forms.html'
         icon = 'doc-empty'
+
+class CustomTableBlock(blocks.StructBlock):
```
noahmanger: Is this an accident having this in here twice?

johnnyporkchops: Yes that is a mistake, I thought I was starting with a fresh checkout instead of a merge so i just went to the bottom and added by block no realizing that it was merged above the new block

johnnyporkchops:
**home/templates/blocks/custom_table.html**: Removed extraneous markup from this template by removing a hardcoded wrapper div that I had originally added--Wagtail appends the class ```block-custom_table``` to the body div and I use that as the wrapper class instead.

**home/models.py** had ``` ContactInfoBlock``` listed twice in the import statements at the top -- I removed that.

**home/blocks.py** : Removed the extra block definition for CustomTableBlock

PR #902, opened 2017-03-20
Removing the max-length on the committee name field

noahmanger:
Removes all of the `max_length` attributes from the RAD contact form.


noahmanger:
This also adds in the contact info items:
image

PR #914, opened 2017-03-23
Release/public beta 2017 03 24

ccostino:
Secondary release branch being merged into `master` as a part of our mini-release today.


package.json
```diff
@@ -11,11 +11,10 @@
     "aria-accordion": "0.1.1",
     "component-sticky": "1.0.0",
     "es6-weak-map": "2.0.1",
-    "fec-style": "7.6.0",
+    "fec-style": "7.7.0",
```
noahmanger: Can we just change this to pin to major only?

ccostino: We've run into some issues with this in the past, I'd like to spend just a bit more time making sure we have it set correctly when we're not pushing things out immediately. :-)

noahmanger: 👍

PR #920, opened 2017-03-27
Disables redirect in case environment is local

jontours:
@ccostino @LindsayYoung Well, turns out we did have people doing local environment!!  Hooray for rapid iterations.  This disables the redirect for local dev.

### DIFF
--- a/fec/fec/urls.py
+++ b/fec/fec/urls.py
@@ -16,7 +16,6 @@ from search import views as search_views
 urlpatterns = [
     url(r'^django-admin/', include(admin.site.urls)),
     url(r'^auth/', include(uaa_urls)),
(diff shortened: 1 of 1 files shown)


fec/fec/urls.py
```diff
@@ -28,6 +28,9 @@
     url(r'', include(wagtail_urls)),
 ]
 
+if settings.FEC_CMS_ENVIRONMENT != 'LOCAL':
+    urlpatterns.insert(2,url(r'^admin/login', uaa_views.login, name='login'))
+
```
ccostino: Does the original line need to be removed in the `urlpatterns` set above?

jontours: Yup!  snafu on my part, change coming!  thanks.

PR #937, opened 2017-03-28
Feature/gov cloud support

LindsayYoung:
Adds Bandit
removes Sentry


ccostino:
@LindsayYoung, while merging and taking care of the `cf` CLI update I went ahead and updated a few other things as well to bring the CMS up to Python 3.5.3 too.  I figured I'd hold off on merging just in case you wanted to take a quick glance at it, but the updates are largely the same as the web app and API, nothing different!

PR #943, opened 2017-03-29
Feature/wagtail tweaks

noahmanger:
This fixes a few small things that @emileighoutlaw brought to my attention, including:

- The need to have more than just `CollectionPages` nested directly under the `ServicesLandingPage`
- A bug where if a section title began with a number (like "2018 Reporting dates"), it would result in an invalid HTML `id` because those can't begin with digits. So I added a filter to strip these out. This will fix the issue with the sticky side nav not working on certain pages
- Adds the ability to make a "related pages" sidebar on the `ResourcePage`, like:

image


xtine:
Is it necessary to have an extra tag to remove the digits, or could something be appended in the HTML like `href="#id-{{section.value.title | slugify }}"` ?

noahmanger:
I thought about that, but then thought this would be better because it makes the URLs prettier when the fragment is in there. But down to go the simpler route if you think it's better.

xtine:
@noahmanger: Ok, that makes sense. I was just wondering. Potentially can reuse the tag for something else as well.

fec/home/templates/partials/section-nav.html
```diff
@@ -1,11 +1,12 @@
 {% load wagtailcore_tags %}
+{% load filters %}
 
 <div class="sidebar-container sidebar-container--left">
   <div class="js-sticky-side" data-sticky-container="{{ container }}">
     <nav class="sidebar sidebar--neutral sidebar--left side-nav js-toc">
       <ul class="sidebar__content">
           {% for section in sections %}
-            <li class="side-nav__item"><a class="side-nav__link" href="#{{section.value.title | slugify }}">{{ section.value.title }}</a></li>
+            <li class="side-nav__item"><a class="side-nav__link" href="#{% filter remove_digits|slugify %}{{ section.value.title }}{% endfilter %}">{{ section.value.title }}</a></li>
```
xtine: Is it possible to do `{% section.value.title|remove_digits|slugify %}` ?

noahmanger: I tried that and it didn't seem to work. This seems to be Django's solution. https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#filter

PR #948, opened 2017-03-30
Update Agenda model time

xtine:
This updates the `AgendaPage` model. The date and time fields are renamed to `date` and `time` accordingly to be consistent with the other Page models. This is important for the Latest Updates page in particular, as the page querysets are combined and sorted by `date` and any unique naming of `date` will throw this off.

The `date` field is also changed from DateTimeField to just DateField to remove ambiguity as DateTimeField stores both date and time. Switching to just DateField also makes it consistent with all other date fields and more separates the date and time fields for Wagtail editing.

In addition, the import meeting agendas scraper is updated to store just date for now as we figured out time isn't scraped correctly as of yet, which may be addressed in another issue.

/cc @johnnyporkchops FYI on model changes.


fec/home/models.py
```diff
@@ -692,8 +692,8 @@ def hero_class(self):
 
 
 class AgendaPage(Page):
-    mtg_date = models.DateTimeField(default=datetime.date.today)
-    mtg_time = models.TimeField(default=datetime.time(10, 00))
+    date = models.DateField(default=datetime.date.today)
+    time = models.TimeField(default=datetime.time(10, 00))
```
noahmanger: Can we make `time` not required and without a default, that way imported agendas won't have it set and we can then exclude it from the template?

ccostino: Hmm, yeah, we could do that - it'd default to `None` in the case of being not required, but definitely fine.

PR #952, opened 2017-03-31
Feature/transition urls

noahmanger:
This just does a quick find-and-replace of all the hard-coded links to fec.gov in the CMS and introduces config variables that can be set to point to given URLs. This way, when @patphongs has the new transition pages up, we can just set the variable for transition links to go to those. 

Both the `FEC_TRANSITION_URL` and `FEC_CLASSIC_URL` default to `www.fec.gov` but in the future they'll point to `transition.fec.gov` and `classic.fec.gov` respectively.


noahmanger:
Closing in favor of incorporating these changes in https://github.com/18F/fec-cms/pull/959

PR #963, opened 2017-04-04
Vacant commissioners

johnnyporkchops:
Automatically show vacant seats on current commissioner's block including Vacant Chair or Vacant Vice Chair

### DIFF
--- a/fec/home/templates/partials/current-commissioners.html
+++ b/fec/home/templates/partials/current-commissioners.html
@@ -1,11 +1,48 @@
+{% load wagtailimages_tags %}
+{% load staticfiles %}
 {% comment %}
   This partial is used by the commissioners inclusion tag in commissioners.py
-  Show the chair, then the vice chair, then everyone else
+  Show the chair, then the vice chair, everyone else and vacant seats
 {% endcomment %}
 <div class="grid grid--2-wide">
+{% if chair_commissioner %}
   {% include 'partials/commissioner.html' with commissioner=chair_commissioner %}
+{% else %}
+  <div class="grid__item commissioner-height">
+  <div class="icon-heading">
+    <img class="icon-heading__image" src="{% static "img/headshot--no-photo.png" %}" alt="Headshot of vacant seat">
+    <div class="icon-heading__content">
+      <div class="t-lead">Vacant Chair Seat</div>
+    </div>
+  </div>
+</div>
+{% endif %}
+
+{% if vice_commissioner %}
   {% include 'partials/commissioner.html' with commissioner=vice_commissioner %}
+{% else %}
+  <div class="grid__item commissioner-height">
+  <div class="icon-heading">
+    <img class="icon-heading__image" src="{% static "img/headshot--no-photo.png" %}" alt="Headshot of vacant seat">
+    <div class="icon-heading__content">
+      <div class="t-lead">Vacant Vice-Chair Seat</div>
+    </div>
+  </div>
+</div>
+{% endif %}
+
 {% for commissioner in commissioners %}
   {% include 'partials/commissioner.html' with commissioner=commissioner %}
+{% endfor %}
+
+{% for vacant in vacant_seats %}
+  <div class="grid__item commissioner-height">
+  <div class="icon-heading">
+    <img class="icon-heading__image" src="{% static "img/headshot--no-photo.png" %}" alt="Headshot of vacant seat">
+    <div class="icon-heading__content">
+      <div class="t-lead">Vacant Seat</div>
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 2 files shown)


LindsayYoung:
There are not any vacant seats, right now, correct? If we don't need this code right away, we will probably just want this merged in the next release or the release after if we want to merge it into dev.

johnnyporkchops:
Yes there are not vacant chair seats now but the original request for this was to handle the vacant non-chair seat wherever this current-commissioner block is used (not on homepage-- that is commissioner section is hardcoded) So if that block is not being used anywhere, then this can wait--and it can probably wait even if that block is being used somewhere.

fec/home/templatetags/commissioners.py
```diff
@@ -2,19 +2,26 @@
...
 from django import template
 from home.models import CommissionerPage
+from django.db.models import Q
 
 register = template.Library()
 
 @register.inclusion_tag('partials/current-commissioners.html')
 def current_commissioners():
-    chair_commissioner = CommissionerPage.objects.filter(commissioner_title__contains='Chair') \
+    #vacant_chair = ""
```
noahmanger: Remove comment

fec/home/templatetags/commissioners.py
```diff
@@ -2,19 +2,26 @@
...
+    chair_commissioner = CommissionerPage.objects.filter(commissioner_title__startswith='Chair') \
         .exclude(commissioner_title__contains='Vice').first()
     vice_commissioner = CommissionerPage.objects.filter(commissioner_title__startswith='Vice').first()
-    commissioners = CommissionerPage.objects.filter(commissioner_title__exact='', \
-        term_expiration__isnull=True).order_by('last_name')
+    commissioners = CommissionerPage.objects.filter(term_expiration__isnull=True) \
+        .exclude(Q(commissioner_title__startswith='Chair') \
```
noahmanger: Does this `exclude(Q(...))` need to be here? Isn't that what `commissioner_title_exact=''` does?

fec/home/templates/partials/current-commissioners.html
```diff
@@ -1,11 +1,48 @@
...
 {% endcomment %}
 <div class="grid grid--2-wide">
+{% if chair_commissioner %}
   {% include 'partials/commissioner.html' with commissioner=chair_commissioner %}
+{% else %}
+  <div class="grid__item commissioner-height">
+  <div class="icon-heading">
+    <img class="icon-heading__image" src="{% static "img/headshot--no-photo.png" %}" alt="Headshot of vacant seat">
+    <div class="icon-heading__content">
+      <div class="t-lead">Vacant Chair Seat</div>
```
noahmanger: "Seat" should be lowercase.

fec/home/templates/partials/current-commissioners.html
```diff
@@ -1,11 +1,48 @@
...
+{% endif %}
+
+{% if vice_commissioner %}
   {% include 'partials/commissioner.html' with commissioner=vice_commissioner %}
+{% else %}
+  <div class="grid__item commissioner-height">
+  <div class="icon-heading">
+    <img class="icon-heading__image" src="{% static "img/headshot--no-photo.png" %}" alt="Headshot of vacant seat">
+    <div class="icon-heading__content">
+      <div class="t-lead">Vacant Vice-Chair Seat</div>
```
noahmanger: "Seat" should be lowercase.

fec/home/templates/partials/current-commissioners.html
```diff
@@ -1,11 +1,48 @@
...
 {% for commissioner in commissioners %}
   {% include 'partials/commissioner.html' with commissioner=commissioner %}
 {% endfor %}
+
+{% for vacant in vacant_seats %}
+  <div class="grid__item commissioner-height">
+  <div class="icon-heading">
+    <img class="icon-heading__image" src="{% static "img/headshot--no-photo.png" %}" alt="Headshot of vacant seat">
+    <div class="icon-heading__content">
+      <div class="t-lead">Vacant Seat</div>
```
noahmanger: "Seat" should be lowercase.

fec/home/templatetags/commissioners.py
johnnyporkchops: I just tried to look at it and remember why I added those q values and I know there was a reason.. I think it had to do with the fact that : if term-expiration is null for Vice and Chair , then they would show up in that list, but I could be getting the logic wrong, it can be confusing. If you think it needs to come out, thats fine, and I will test it later and if it breaks add it back in.

noahmanger: Ok cool. Just checking.

PR #972, opened 2017-04-10
Feature/search

noahmanger:
This implements a prototype of a combined candidate/committee/site content search using the DigitalGov Search API. 
image

There's a little bit to explain, and I'd like to discuss it all with folks, so here it is:

## DigitalGov Search
For starters, this makes use of DigitalGov Search's custom indexes, which is basically an ElasticSearch instance that's accessible with the [I14Y indexing API](https://search.digitalgov.gov/developer/i14y.html). This brings a few advantages over relying on a commercial search engine like Bing:

- Real-time, on-demand updating of indexes, rather than waiting for your site to be re-crawled. This is especially useful around a major domain switch.
- Results are accessible over an API, allowing us to customize the display as well as integrate with other tools (like the data API)
- The index accepts optional `description` and `tags` fields for adding custom meta-data for pages
- We can add whatever we want to the index, whether that's a CMS page, a page from the web app, or a static page from transition.fec.gov

## How it works
The [I14Y API is documented here](http://gsa.github.io/slate/). It's possible to use Wagtail's content publishing hooks to push updates directly to this index in the future, but for now I went with a different solution for the sake of doing a bulk update. 

I added two management commands: `scrape_pages` and `index_pages`. `scrape_pages` can scrape either a single page (e.g. the About page: `scrape_pages --page /home/about/`) or all descendants of a page (e.g. all pages in the Help section: `scrape_pages --descendants_of /home/help-candidates-and-committees`). It then goes and scrapes the content within the main content area on each page (as well as whatever data from the CMS database we want) and spits out JSON in this format in `dump.json`:

```
    {
        "title": "Court cases",
        "content": "lorem ipsum",
        "path": "https://beta.fec.gov/legal-resources/court-cases/",
        "language": "en",
        "promote": "false",
        "created": "2017-03-08-185258",
        "document_id": 6667
    },
```

The scraping is a little crude, just grabbing all the text content of the main area and replacing newlines with spaces. Maybe this could be cleaned up, though it seems to work pretty well.

The `index_pages` command takes this JSON and posts it to I14Y to add them. Because each page has a unique `document_id` (which corresponds to the page's `id` in Wagtail), they can also be updated with new info or deleted easily.

## The results display
Once the indexes are set up, it's pretty straightforward. The best way to search for candidates and committees is to hit our API directly, so in the search view, I simply added calls to those endpoints. If there's candidate or committee results, they appear on the top, followed by results from the search API. There's probably a better way to do this, but it gets the point across.

Also, with DigitalGov you can make "Best bets", which are basically suggestions that you associate keywords with. Those appear at the top, like "Calendar" in the screenshot above.

And because not _everything_ is searchable through here (we can constrain or expand as much as we want), I added links to other tools like Legal Resources and Individual Contributions, with the search term passed through. 

Last, just for kicks, I added a specific Latest Updates search term filter, since I don't think we'll want those in the main index because of the number of false positives it will return:
image

Ok! Thanks for reading my first novel. If you've gotten this far and would like to discuss, please do.


noahmanger:
Alright, after hashing out what we think is a solid foundation for the UX and design of this, I think this is in good place for an initial review. 

I'll still add tests before merging, but probably good for other eyes on this sooner than later.

search

noahmanger:
You'll also need https://github.com/18F/fec-style/pull/673 when reviewing.

fec/fec/settings/base.py
```diff
@@ -188,10 +189,12 @@
 CONTACT_EMAIL = 'betafeedback@fec.gov';
 CONSTANTS = constants
 
+
 # Config for the ServiceNow API for contacting RAD
 FEC_SERVICE_NOW_API = env.get_credential('FEC_SERVICE_NOW_API')
 FEC_SERVICE_NOW_USERNAME = env.get_credential('FEC_SERVICE_NOW_USERNAME')
 FEC_SERVICE_NOW_PASSWORD = env.get_credential('FEC_SERVICE_NOW_PASSWORD')
+FEC_DIGITALGOV_KEY = env.get_credential('FEC_DIGITALGOV_KEY')
```
ccostino: This environment variable has been added to the credential service, right?

fec/home/views.py
```diff
@@ -63,47 +84,46 @@ def updates(request):
...
       digests = DigestPage.objects.live()
       press_releases = PressReleasePage.objects.live()
-
-      # Hide behind feature flag unless explicitly requested
-      # Only authenticated users will be able to explicitly request them for now
-      if settings.FEATURES['record']:
-        records = RecordPage.objects.live()
-
-      if settings.FEATURES['tips']:
-        tips = TipsForTreasurersPage.objects.live()
```
ccostino: Nice, thanks for cleaning these out if they're no longer needed!

fec/search/management/commands/index_pages.py
```diff
@@ -0,0 +1,64 @@
...
+import requests
+from bs4 import BeautifulSoup
+
+from django.core.management import BaseCommand
+
+from home.models import Page
+
+
+DIGITALGOV_DRAWER = 'main'
+DIGITALGOV_DRAWER_KEY = os.getenv('DIGITALGOV_DRAWER_KEY')
```
ccostino: This has also been added to the credentials service?

fec/search/management/commands/index_pages.py
```diff
@@ -0,0 +1,64 @@
...
+        self.stdout.write(self.style.WARNING('Indexing pages'))
+
+        with open(options['json_file_path'], 'r') as json_contents:
+            if options['verbosity'] > 1:
+                self.stdout.write((options['json_file_path']))
+
+            pages = json.load(json_contents)
+            for page in pages:
+                self.add(page)
+                # if 'delete-existing' in options:
```
ccostino: If the `delete-existing` option isn't supported, should it be taken out entirely or should this remain so it works?  Or does it matter if this is only intended to be used once?

fec/search/management/commands/scrape_pages.py
```diff
@@ -0,0 +1,117 @@
...
+            default=False,
+            help='Get Tips for Treasurers only'
+        )
+
+    def handle(self, *args, **options):
+        self.stdout.write(self.style.WARNING('Getting pages...'))
+        if options['page']:
+            page = Page.objects.get(url_path=options['page'])
+            # Hack so the loop below works
+            pages = [page]
```
ccostino: Totally fine hack in this case, common pattern in Python. :-)

fec/search/views.py
```diff
@@ -1,40 +1,94 @@
...
+    if 'site' in search_type and search_query:
+        params = {
+            'affiliate': 'betafec_api',
+            'access_key': settings.FEC_DIGITALGOV_KEY,
+            'query': search_query,
+            'limit': 10,
+            'offset': offset
+        }
+        r = requests.get('https://search.usa.gov/api/v2/search/i14y', params=params)
+        web_results = r.json()['web']
```
ccostino: We might want to add just a bit of handling (or change the way the dictionary key is accessed) just in case the response isn't what we expect.

fec/search/views.py
```diff
@@ -1,40 +1,94 @@
...
+            'best_bets': {
+                'results': r.json()['text_best_bets'],
+                'count': len(r.json()['text_best_bets'])
+            },
+            'meta': {
+                'count': web_results['total'],
+                'next_offset': web_results['next_offset'],
+                'prev_offset': prev_offset(limit, int(offset))
+            }
+        }
```
ccostino: Same note as above for these properties; e.g., [`my_dict.get('key_name', <default_value/type>)`](https://docs.python.org/3/library/stdtypes.html#dict.get).

package.json
```diff
@@ -11,7 +11,7 @@
     "aria-accordion": "0.1.1",
     "component-sticky": "1.0.0",
     "es6-weak-map": "2.0.1",
-    "fec-style": "9.0.1",
+    "fec-style": "git+https://github.com/18f/fec-style#header-search",
```
ccostino: We'll have to modify this as well once the new `fec-style` version goes in, before this is merged.

ccostino:
@noahmanger, I tried to get this to work but have a couple of questions:

* One, can I use the credentials we (I'm assuming) now have from DigitalGov for i14y or should I get my own?
* I tried running `scrape_pages --page /home/about/` but it complained about no `fec/search/management/dump.json` file existing - do I need to do something else or should it be writing this file for me?

noahmanger:
@ccostino yeah there's only one set. I'll add them to the credential service. And no, it doesn't write a new file. I always just have one but that might be nice.

fec/search/management/commands/index_pages.py
ccostino: Since this is in the credential service, this will need to change slightly.  You'll have to import the `env` module we have first:

```python
from .env import env
```

Then this line will change to this:

```python
DIGITALGOV_DRAWER_KEY = env.get_credential('DIGITALGOV_DRAWER_KEY')
```

This will check the credential service first and if it's not found there it'll fall back to looking in the local environment variables, this way the statement still works locally and in cloud.gov. :-)

Alternatively, you could also retrieve and set this constant in the settings file like is done with `FEC_DIGITALGOV_KEY` and just retrieve it from the settings.  Just remember to import the settings in that case:

```python
from django.conf import settings
```

fec/search/management/commands/scrape_pages.py
```diff
@@ -0,0 +1,117 @@
...
+        content = article if article else main
+
+        for tag in content:
+            text = text + tag.get_text().replace('\n', ' ')
+
+        return text
+
+    def write_articles(self, pages, **options):
+        self.stdout.write('Writing to file')
+        fname = os.path.join('fec/search/management/dump.json')
```
ccostino: After testing this some, I wonder if this file path is better expressed this way:

```python
os.path.join(settings.REPO_DIR, 'fec/search/management/dump.json')
```

The reason being is that this will account for where the app is currently located so that the sub-directories and file can always be found.  This way the scraping will always be able to write to the file regardless of where the management command is executed.

In order to do this, you'll also need to import the project settings up top:

```python
from django.conf import settings
```

That'll give you access to `settings.REPO_DIR`.

fec/search/management/commands/index_pages.py
noahmanger: Ah yeah I'll remove it.

fec/search/management/commands/index_pages.py
ccostino: This can be disregarded, @noahmanger and I spoke and the intention is that these would just be run and managed locally.

ccostino:
Very awesome to see come together, @noahmanger, well done!  🎉 👏  As soon as we can get the tests fixed I say this is ready and good to go. :-)

PR #986, opened 2017-04-12
Logging feature for ATO

jontours:
The receiver as defined as been tested and currently successfully receives changes to a user, pages I'm not too sure (as there are multiple page models so maybe just figure out which are needed).  Logging statement needs to be formalized too (time/date, specific action).


jontours:
This feature is a good start for logging.  The following is logged: 

- user login
- user change
- page modification w/ username making change
- any time a user is added/removed from Group; username and group name logged.

LindsayYoung:
Thanks for all your work on this!

LindsayYoung:
I think there should be some migrations in here, since there were model changes.

PR #991, opened 2017-04-14
Removing old registration and reporting models

noahmanger:
Now that we no longer have the Registration and Reporting section, this is a bit of housekeeping to remove the old page models for that section, along with the templates.

Note: we're still in the process of moving the children pages from this section to the new Help section, so this shouldn't be merged in until that has happened.


noahmanger:
Ok this can be merged in, now that we've moved those old pages. @xtine can you take a look?

xtine:
Migration needs to be redone.

noahmanger:
Merged the migrations. I noticed that a bunch of the builds are failing because of unmerged migrations, so hopefully this resolves it.

PR #1004, opened 2017-04-20
Adding calculator icon to section aside

noahmanger:
image

For the admin fine calculator link.

cc @jenniferthibault 

---

Update, this also includes a change adding the raw HTML block to resource page templates for @emileighoutlaw .


fec/home/models.py
```diff
@@ -523,6 +523,7 @@ class DocumentPage(ContentPage):
                                 choices=constants.report_child_categories.items(), null=True)
     content_panels = Page.content_panels + [
         FieldPanel('date'),
+        FieldPanel('year_only'),
```
jenniferthibault: What is the use case for this?

noahmanger: Certain documents (like [FOIA reports](https://beta.fec.gov/about/reports-about-fec/foia-reports/)) don't make sense to show the month. This was already a setting we have enabled behind the scenes that we enabled when we imported the first batch, so this just adds a setting in the editor.

jenniferthibault: Cool, thanks!

PR #1006, opened 2017-04-21
Feature/rad lookup

noahmanger:
This adds the RAD lookup tool to the contact form page. 

image

The lookup tool uses the typeahead, but on `typeahead:select` it calls the `/rad-analyst/` endpoint and shows the name below. 

image

When you clear the input, the name goes away.

On successful submit of the contact form, the lookup stays there:

Last, I made it so that if the contact from feature flag is disabled, this page is still accessible, but the form just doesn't show:

image

So this also enables the card to this page from the Help landing page, and then once the form can go live, we can just turn it on.

Also FYI, it's possible to show more info in here, such as the analyst's extension, email address and title if we want to (it's all in the API). cc @llienfec 

cc @jenniferthibault @xtine @emileighoutlaw


noahmanger:
Updated the dependencies, so this is all good to go.

xtine:
LGTM, but @ccostino can you also 👀 to make sure I didn't miss anything?

ccostino:
Sure thing!

fec/fec/tests/js/analyst-lookup.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    it('does not clear the input if there is a value', function() {
+      var $input = this.lookup.$input;
+      $input.val('Some value');
+      this.lookup.handleChange({target: $input});
+      expect(this.hide).to.not.have.been.called;
+    });
+  });
+});
```
ccostino: Nice tests!

fec/home/views.py
```diff
@@ -211,37 +211,37 @@ def commissioners(request):
...
-  # If it's a POST, post to the ServiceNow API
-  if request.method == 'POST':
-    form = ContactRAD(request.POST)
-    response = form.post_to_service_now()
-    if response == 201:
-      return render(request, 'home/contact-form.html', {
-        'self': page_context,
-        'success': True
-      })
+    if settings.FEATURES['radform']:
```
ccostino: We're still making use of this flag for now, right?

PR #1021, opened 2017-04-26
Updating jquery and fullcalendar

noahmanger:
This updates jQuery to the latest version. In my testing, however, I quickly discovered the calendar was broken, and learned that the library we use for it added support for jQuery 3 in [2.8.0](https://github.com/fullcalendar/fullcalendar/blob/master/CHANGELOG.md#v280-2016-06-19), so I updated that library as well.

Everything seems to work now, but again, another set of eyes wouldn't hurt.


ccostino:
D'oh, looks like we still have a test failure, but that's because of the `fec-style` update needed first with https://github.com/18F/fec-style/pull/686, right @noahmanger?

noahmanger:
That's my best bet, because I can't think of any other reason this would fail. The tests pass locally for me.

PR #1023, opened 2017-04-26
Removing links to citizens info

noahmanger:
Small thing I noticed: since we're not launching the public info pages, we need to remove the links on the home page.

### DIFF
--- a/fec/home/templates/home/home_page.html
+++ b/fec/home/templates/home/home_page.html
@@ -110,9 +110,11 @@
       <h2 class="heading--section">Introduction to campaign finance and elections</h2>
       <div class="billboard">
         <div class="billboard__image__container">
-          <!-- <span class="t-block u-padding--bottom">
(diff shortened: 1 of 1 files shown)


fec/home/templates/home/home_page.html
```diff
@@ -110,9 +110,9 @@ <h3 class="heading__left">Upcoming events</h3>
       <h2 class="heading--section">Introduction to campaign finance and elections</h2>
       <div class="billboard">
         <div class="billboard__image__container">
-          <span class="t-block u-padding--bottom">
+          <!-- <span class="t-block u-padding--bottom">
```
xtine: Might be better to use the Django commenting so it doesn't show in source: `{% comment %} {% endcomment %}`

noahmanger: Good idea.

PR #1026, opened 2017-04-26
Add missing transition URL env var

ccostino:
This changeset adds a missing env var to our base manifest to ensure that the transition site URL does not get lost on subsequent deploys.


noahmanger:
Just checking: this will work with the way that we get the variable? https://github.com/18F/fec-cms/blob/develop/fec/fec/settings/base.py#L206

ccostino:
Yup!  That method will check the credential service first and if it doesn't find the env var there, it'll fall back to the app-specific env vars that are set either via the manifest or manually.

PR #1030, opened 2017-04-27
Update remaining env var references

ccostino:
This changeset fixes a few more env var references and makes use of the `env.get_credential()` method to ensure that an env var is found.  To recap, that method checks the credential service in Cloud Foundry first for an env var, and if it cannot find it then it falls back to checking the host environment the app is running in.  This method works in both the deployed apps and locally for that reason.


fec/fec/settings/base.py
```diff
@@ -202,6 +202,7 @@
 FEC_SERVICE_NOW_USERNAME = env.get_credential('FEC_SERVICE_NOW_USERNAME')
 FEC_SERVICE_NOW_PASSWORD = env.get_credential('FEC_SERVICE_NOW_PASSWORD')
 FEC_DIGITALGOV_KEY = env.get_credential('FEC_DIGITALGOV_KEY')
+FEC_DIGITALGOV_DRAWER_KEY - env.get_credential('DIGITALGOV_DRAWER_KEY')
```
noahmanger: This is a `-` not `=`

ccostino: D'OH!

LindsayYoung:
Thanks!

PR #1032, opened 2017-05-01
Link to calendar from cards

noahmanger:
Per the discussion in https://github.com/18F/fec-cms/issues/1011, this replaces the "upcoming date" on the Help section of the home page with a link to deadlines on the calendar, since the logic of which deadline to show is too complicated (and not currently supported by the API) for now.

image


xtine:
Looks good. I'll make the PR to remove the JavaScript logic in `fec-style`.

xtine:
Oops, typed too quickly and realized that the JS that controls this is also in this repo. Can you remove `new upcomingEvents.UpcomingDeadlines();` and the corresponding function code in `upcoming-events.js`?

noahmanger:
Oh yep. Good catch.

xtine:
👍

PR #1038, opened 2017-05-04
Update strategy and budget feed drop-down options

emileighoutlaw:
We aren't using the `IT strategic plan` or `Annual performance report` tags in our "Performance, strategy and budget" section, so we don't need them in the dropdown option. This PR removes them.

I want to triple-check with @AmyKort that this is correct before a merge, though.


AmyKort:
Sounds right to me. Thanks!

PR #1041, opened 2017-05-05
Updates "donating"

emileighoutlaw:
This changes the word "donating" to "giving," which is more technically precise. The edit is a request from FEC, and I agree that it's a good one. :)


noahmanger:
Replaced by https://github.com/18F/fec-cms/pull/1049

PR #1050, opened 2017-05-10
Adding mid-year and year-end filters

noahmanger:
It turns out the calendar data was correctly coded all along, we just weren't exposing the filters on the front-end. 

This adds them here:
image

And also adds them in the URLs from the home page.

cc @llienfec @AmyKort @anthonygarvan


emileighoutlaw:
Capitalized `Mid-Year` and `Year-End` because @llienfec recently taught me that's FEC's preferred style :)

PR #1053, opened 2017-05-11
Feature/social images

noahmanger:
Adds custom social images for the different sections of the site, with fallback to a generic one. 

To do this, I had to:

- Make a function for `CustomPage`, `ResourcePage` and `CollectionPage` instances to figure out which section they're in, since they can be used in multiple places.
- Make template filters for generating static paths based on the parent section

It works, but let me know if there's a better way to do this.

Requires https://github.com/18F/fec-style/pull/699

### DIFF
--- a/fec/home/models.py
+++ b/fec/home/models.py
@@ -88,7 +88,7 @@ class ContentPage(Page):
     # Default content section for determining the active nav
     @property
     def content_section(self):
-        return 'help'
+        return ''
 '''
 class Person(User):
     objects = User()
@@ -760,7 +760,7 @@ class LegalResourcesLandingPage(ContentPage, UniqueModel):
 
     @property
     def content_section(self):
-        return 'legal-resources'
+        return 'legal'
 
 
 class ServicesLandingPage(ContentPage, UniqueModel):
@@ -789,7 +789,7 @@ class ServicesLandingPage(ContentPage, UniqueModel):
 
     @property
     def content_section(self):
-        return 'candidate-and-committee-services'
+        return 'help'
 
     @property
     def hero_class(self):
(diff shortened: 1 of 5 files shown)


fec/home/models.py
```diff
@@ -168,10 +187,6 @@ def content_section(self):
         return ''
 
 
-class LandingPage(ContentPage):
-    template = 'home/registration-and-reporting/landing_page.html'
```
ccostino: Ah, we're getting rid of this model?

fec/home/models.py
```diff
@@ -54,6 +54,25 @@
...
+    }
+    ancestors = page.get_ancestors()
+    content_sections = [
+        slugs.get(ancestor.slug) for ancestor in ancestors
+        if slugs.get(ancestor.slug) != None
+    ]
+    if len(content_sections):
+        return content_sections[0]
+    else:
+        return ''
```
ccostino: Nice!

fec/home/templatetags/filters.py
```diff
@@ -87,3 +88,25 @@ def splitlines(value):
...
+        return static('img/social/general.png')
+
+
+@register.filter(name='get_touch_icon')
+def get_touch_icon(content_section, dimension):
+    """
+    Returns a path to a touch icon for the given dimension and content_section
+    """
+    print('====')
+    print(content_section)
```
ccostino: Do we need these `print` statements in here?

fec/home/models.py
noahmanger: Yep! We don't need it anymore because that section doesn't exist.

fec/home/templatetags/filters.py
noahmanger: Oops! Fixed.

noahmanger:
Added the fec-style bump here.

PR #1055, opened 2017-05-11
[MERGE WITH GIT FLOW] Fix meeting agenda import script

ccostino:
**MERGE WITH GIT FLOW**

This is a hotfix to change the model references in the meeting agenda import script to reflect the recent model changes.  The model name had changed to `MeetingPage` from `AgendaPage`, but the import script was not updated to account for this.


noahmanger:
Does this need to be hotfixed? Or can we just merge into develop?

ccostino:
At this point it can just go into `develop` I think.  Thanks!

PR #1056, opened 2017-05-12
Feature/search indexing improvements

noahmanger:
This cleans up and simplifies the search engine scraping / indexing scripts for easier maintenance (before it's all automated). I also wrote up some [detailed instructions](https://github.com/18F/fec-cms/blob/feature/search-indexing-improvements/fec/search/management/instructions.md) in `instructions.md`. 

Even when the CMS is automatically pushing updates to the search engine, we will still have static app pages and transition pages that need to be manually maintained. I've had a [spreadsheet](https://docs.google.com/spreadsheets/d/1IN6bOtMjAg92mMGcgKhtss5G7vJz0OXXggOM9FjtdAw/edit#gid=0) 🔒  going for this, but for the sake of keeping all the data in one place, I checked in the data in JSON format here as well. 

This needs to get out of just my brain, so would appreciate eyes from @patphongs @ccostino @xtine and anyone else.

This feels a little cumbersome, so if anyone has ideas for improvement (or wants me to clarify anything), lemme know.


fec/fec/settings/base.py
```diff
@@ -202,7 +202,8 @@
 FEC_SERVICE_NOW_USERNAME = env.get_credential('FEC_SERVICE_NOW_USERNAME')
 FEC_SERVICE_NOW_PASSWORD = env.get_credential('FEC_SERVICE_NOW_PASSWORD')
 FEC_DIGITALGOV_KEY = env.get_credential('FEC_DIGITALGOV_KEY')
-FEC_DIGITALGOV_DRAWER_KEY = env.get_credential('DIGITALGOV_DRAWER_KEY', '')
+FEC_DIGITALGOV_DRAWER_KEY_MAIN = env.get_credential('DIGITALGOV_DRAWER_KEY_MAIN', '')
+FEC_DIGITALGOV_DRAWER_KEY_TRANSITION = env.get_credential('DIGITALGOV_DRAWER_KEY_TRANSITION', '')
```
ccostino: These changes have been accounted for in the credential service as well, right?

fec/search/management/commands/scrape_cms_pages.py
```diff
@@ -65,15 +73,25 @@ def handle(self, *args, **options):
...
+                pages += Page.objects.child_of(parent).live()
+
         extracted = []
 
         for page in pages:
             p = {
               "document_id": page.id,
               "title": page.title,
-              "path": "https://beta.fec.gov" + page.url,
+              "path": BASE_URL + page.url,
```
ccostino: Nice catch!

fec/search/management/commands/scrape_transition_pages.py
```diff
@@ -0,0 +1,92 @@
...
+            help='Path to JSON file to load'
+        )
+
+    def handle(self, *args, **options):
+        self.stdout.write(self.style.WARNING('Getting pages...'))
+        extracted = []
+        if options['json_file_path']:
+            file_name = options['json_file_path']
+        else:
+            file_name = os.path.join(settings.REPO_DIR, 'fec/search/management/data/transition_pages.json')
```
ccostino: I noticed the `.gitignore` update to ignore the CMS output file, should we ignore this JSON file too?

fec/search/management/commands/scrape_transition_pages.py
```diff
@@ -0,0 +1,92 @@
...
+
+        for tag in content:
+            text = text + tag.get_text().replace('\n', ' ')
+
+        return text
+
+    def write_articles(self, pages, **options):
+        """Write the extracted data to a file"""
+        self.stdout.write('Writing to file')
+        fname = os.path.join(settings.REPO_DIR, 'fec/search/management/data/output.json')
```
ccostino: Is the intention to overwrite the `fec/search/management/data/output.json` or should this be its own file?

fec/search/management/commands/scrape_web_app_pages.py
```diff
@@ -0,0 +1,92 @@
...
+            help='Path to JSON file to load'
+        )
+
+    def handle(self, *args, **options):
+        self.stdout.write(self.style.WARNING('Getting pages...'))
+        extracted = []
+        if options['json_file_path']:
+            file_name = options['json_file_path']
+        else:
+            file_name = os.path.join(settings.REPO_DIR, 'fec/search/management/data/web_app_pages.json')
```
ccostino: Same ignore note on this file as above, should it be added?

fec/search/management/commands/scrape_web_app_pages.py
```diff
@@ -0,0 +1,92 @@
...
+
+        for tag in main:
+            text = text + tag.get_text().replace('\n', ' ')
+
+        return text
+
+    def write_articles(self, pages, **options):
+        """Write the extracted data to a file"""
+        self.stdout.write('Writing to file')
+        fname = os.path.join(settings.REPO_DIR, 'fec/search/management/data/output.json')
```
ccostino: Same note about the overwriting of the `fec/search/management/data/output.json` file as above - is this intentional?

fec/search/management/instructions.md
```diff
@@ -0,0 +1,109 @@
...
+3. **Index the pages:** Run `fec/manage.py index_pages -transition`. This command works the same as it does for adding CMS pages, but with the `-transition` flag it will put them in the transition drawer.  
+
+## Additional DigitalGov configuration
+**Best bets:** of the really great features of DigitalGov Search is what's called "Best bets". These are basically search suggestions that you can manually add (or add in bulk by uploading a spreadsheet) which map a URL to a specific set of keywords. Any Best Bet will be returned at the top of the search results. 
+
+**Deleting pages:** To remove pages from the index, you'll need to make a DELETE request with the `document_id` you want to delete. [More info in the docs](http://gsa.github.io/slate/#delete-a-document).
```
ccostino: This whole set of instruction is _awesome_ to see! :-)

fec/search/management/commands/scrape_transition_pages.py
noahmanger: Yeah, my thought was that output.json would just always be a temporary file. Does that make sense?

fec/search/management/commands/scrape_transition_pages.py
noahmanger: That was intentional, but open to changing. My thought was that we could check in these files for all of them, so that way it at least always lives somewhere public.

fec/search/management/instructions.md
noahmanger: Thanks! Yeah, kind of a convoluted system, so feels good to have it all written out.

fec/fec/settings/base.py
noahmanger: These are only used when running the script, which happens locally, and they're accessible from the digitalgov admin panel, so I was thinking they don't need to be included in the service. What do you think?

ccostino: Oh that's right I forgot that this was a manual process and run locally.  If that's the case then I don't think they need to be there.  I was operating on auto-pilot when I saw the calls to retrieve them and assumed they'd have to be there.

fec/search/management/commands/scrape_transition_pages.py
ccostino: As I reviewed this more I picked up on that intention, which makes sense to me!

fec/search/management/commands/scrape_transition_pages.py
ccostino: Yup, I just wanted to double check, thanks! :-)

PR #1060, opened 2017-05-12
Feature/meeting template updates

xtine:
Updated design for meeting templates.

screencapture-127-0-0-1-8000-updates-april-20-2017-open-meeting-1494651825098
screencapture-127-0-0-1-8000-updates-april-20-2017-open-meeting-1494651773391

TODO:
- styling for agenda item audio
- prevent auto-downloading of uploaded documents on wagtail


noahmanger:
Alright, I think I finished this up since Christine is out this week. A bunch of small changes:

Open meeting, when only a sunshine act notice is posted:
image

Executive session:
image

Meeting with live video and captions:
image

Agenda:
image

Minutes, recordings and notices:
image

Old agendas will now show minutes and notices at the bottom after the agenda:
image

Last, I tidied up the Wagtail interface, grouping similar fields in collapsible panels:
image

fec/home/models.py
```diff
@@ -819,49 +819,68 @@ class MeetingPage(Page):
...
     )
 
-    mtg_media = StreamField([
-        ('full_video_url', blocks.TextBlock(required=False)),    # 'video_link'
-        ('full_audio_url', blocks.TextBlock(required=False)),    # 'primary_audio_link'
-        ('mtg_transcript_url', blocks.TextBlock(required=False))  # 'closed_captioning_link'
-    ])
+    full_video_url = models.URLField(blank=True)
+    full_audio_url = models.URLField(blank=True)
+    mtg_transcript_url = models.URLField(blank=True)
```
ccostino: Ah, nice to see this could be simplified a bit!  I know we weren't sure about that before.

fec/home/models.py
```diff
@@ -819,49 +819,68 @@ class MeetingPage(Page):
...
 
     agenda = StreamField([
-        ('agenda_item', blocks.StreamBlock([
+        ('agenda_item', blocks.StructBlock([
```
ccostino: I'm not sure I understand the switch from a `StreamBlock` to a `StructBlock` here, but I'm also not seeing anything wrong with it based on the [documentation about blocks](http://docs.wagtail.io/en/v1.10/topics/streamfield.html#structural-block-types).

Definitely not a showstopper, but I am curious!

ccostino:
Also noting here that like some of the other open PRs, we need to figure out which one to incorporate the `fec-style` update in to get the supporting styles/updates in place too.

fec/home/models.py
noahmanger: `StreamBlock` is for adding a flexible number of sub-blocks, whereas `StructBlock` gives you a predefined set of sub-blocks. Basically it makes the editing interface a little nicer and the template code a little cleaner.

ccostino: Okay, that makes sense, thanks!

PR #1069, opened 2017-05-18
Meeting feed fix

noahmanger:
This makes a few final fixes for the meetings to be released tomorrow.

1. It adds the open meeting agendas to the "What's happening" section:
image

2. It removes the check on user authentication for showing meetings on latest updates, thus adding it as an option (I also arranged these alphabetically like we discussed @jameshupp ):
image

3. It re-arranges the content in the "About latest updates" box to more closely match the order in the select:
image

4. It fixes a bug that @AmyKort noticed where certain events weren't showing up in the "upcoming events" feed. I figured out that this is because in a case like `category=Open+meeting`, the `+` was being url encoded. I changed the calendar URL builder to decode the URL, but maybe there's a better way?

### DIFF
--- a/fec/fec/static/js/calendar-helpers.js
+++ b/fec/fec/static/js/calendar-helpers.js
@@ -23,7 +23,7 @@ function getGoogleUrl(event) {
 }
 
 function getUrl(path, params) {
-  return URI(window.API_LOCATION)
+  var url = URI(window.API_LOCATION)
     .path(Array.prototype.concat(window.API_VERSION, path || [], '').join('/'))
     .addQuery({
       api_key: window.API_KEY,
@@ -31,6 +31,9 @@ function getUrl(path, params) {
     })
     .addQuery(params || {})
     .toString();
+
+  // Decode in order to preserve + signs
+  return URI.decode(url);
 }
 
 function className(event) {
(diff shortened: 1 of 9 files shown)


fec/fec/static/js/calendar-helpers.js
```diff
@@ -23,14 +23,17 @@ function getGoogleUrl(event) {
...
     .path(Array.prototype.concat(window.API_VERSION, path || [], '').join('/'))
     .addQuery({
       api_key: window.API_KEY,
       per_page: 500
     })
     .addQuery(params || {})
     .toString();
+
+  // Decode in order to preserve + signs
+  return URI.decode(url);
```
ccostino: I think you're good here for preserving the URL in the format you need, that's how I would've done it!

fec/gulpfile.js
```diff
@@ -25,7 +28,8 @@ gulp.task('build-js', function () {
       .bundle()
       .pipe(source(entry))
       .pipe(buffer())
-      .pipe(uglify())
+      .pipe(gulpif(production, uglify()))
```
ccostino: Oh nice, this fixes the minify issue as well?

PR #1074, opened 2017-05-22
Feature/search index fix

noahmanger:
This makes a few tweaks to the site search system:

- Fixes the `update` command to use the correct URL for the request
- Changes the `scrape_cms_pages` command to only get live *and* public pages (leaving out any that are password protected)
- Replaces the URL of site search results with whatever the `CANONICAL_BASE` is, so that once that's fec.gov rather than beta.fec.gov, all the links will point to the right place
- Highlights text matches in best bet titles

Resolves https://github.com/18F/fec-cms/issues/1025

### DIFF
--- a/fec/search/management/commands/index_pages.py
+++ b/fec/search/management/commands/index_pages.py
@@ -67,8 +67,13 @@ class Command(BaseCommand):
             print(r.__dict__)
 
     def update(self, page):
-        self.stdout.write('Updating {}'.format(page['document_id']))
-        requests.put("https://i14y.usa.gov/api/v1/documents", auth=(drawer, key), data=page)
+        url = "https://i14y.usa.gov/api/v1/documents/" + page.get('document_id')
+        r = requests.put(url, auth=(drawer, key), data=page)
+        if r.status_code == 200:
+            self.stdout.write('Updated {}'.format(page['document_id']))
+        else:
+            self.stdout.write('Could not update {}'.format(page['document_id']))
+            print(r.__dict__)
 
     def delete(self, page):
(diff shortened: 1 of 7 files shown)


fec/search/management/commands/index_pages.py
```diff
@@ -67,8 +67,13 @@ def add(self, page):
...
     def update(self, page):
-        self.stdout.write('Updating {}'.format(page['document_id']))
-        requests.put("https://i14y.usa.gov/api/v1/documents", auth=(drawer, key), data=page)
+        url = "https://i14y.usa.gov/api/v1/documents/{}".format(page.get('document_id'))
+        r = requests.put(url, auth=(drawer, key), data=page)
+        if r.status_code == 200:
+            self.stdout.write('Updated {}'.format(page['document_id']))
+        else:
+            self.stdout.write('Could not update {}'.format(page['document_id']))
+            print(r.__dict__)
```
xtine: rm print statement?

fec/search/templates/search/search.html
```diff
@@ -4,8 +4,6 @@
 
 {% block body_class %}template-searchresults{% endblock %}
 
-{% block title %}Search{% endblock %}
```
xtine: What will the title be instead here?

noahmanger: There's actually a default block for this in `base.html` that just uses the title from the view.

fec/search/management/commands/index_pages.py
noahmanger: This is only run locally and it's helpful for debugging, so I think it makes sense to keep.

xtine:
👍

PR #1079, opened 2017-05-24
Use absolute url for twitter card images

noahmanger:
Currently Twitter cards for CMS pages aren't showing images because they're using relative URLs rather than absolute. This fixes that.


noahmanger:
Closing in favor of https://github.com/18F/fec-cms/pull/1080

PR #1085, opened 2017-05-25
Adding support for automatic search indexing

noahmanger:
This adds Wagtail hooks for automatically adding, updating and deleting pages from the DigitalGov search engine. I basically took the same methods used for manually updating the index, refactored slightly and then added to Wagtail hooks on page creation, editing and deletion.

Still to do:
- [x] Add tests
- [x] Test with a dummy search engine (still waiting for a sandbox account to use)
- [x] Remove print statements

### DIFF
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -204,6 +204,8 @@ FEC_SERVICE_NOW_PASSWORD = env.get_credential('FEC_SERVICE_NOW_PASSWORD')
 FEC_DIGITALGOV_KEY = env.get_credential('FEC_DIGITALGOV_KEY')
 FEC_DIGITALGOV_DRAWER_KEY_MAIN = env.get_credential('DIGITALGOV_DRAWER_KEY_MAIN', '')
 FEC_DIGITALGOV_DRAWER_KEY_TRANSITION = env.get_credential('DIGITALGOV_DRAWER_KEY_TRANSITION', '')
+DIGITALGOV_BASE_API_URL = 'https://i14y.usa.gov/api/v1'
+DIGITALGOV_DRAWER_HANDLE = 'main'
 
 FEC_TRANSITION_URL = env.get_credential('FEC_TRANSITION_URL', 'http://www.fec.gov')
 FEC_CLASSIC_URL = env.get_credential('FEC_CLASSIC_URL', 'http://www.fec.gov')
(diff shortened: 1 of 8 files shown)


noahmanger:
Updates: I just added tests and removed some of the print statements, though seems like it might be helpful to keep some of them in, unless there's a better way to log errors?

Also, still waiting on a sandbox account to test it with.

noahmanger:
And it doesn't seem possible to include status messages in the Wagtail editor on success or error, unfortunately.

noahmanger:
Alright this is all done. I was able to verify it works correctly with a sandbox search engine.

fec/fec/settings/base.py
```diff
@@ -204,6 +204,8 @@
 FEC_DIGITALGOV_KEY = env.get_credential('FEC_DIGITALGOV_KEY')
 FEC_DIGITALGOV_DRAWER_KEY_MAIN = env.get_credential('DIGITALGOV_DRAWER_KEY_MAIN', '')
 FEC_DIGITALGOV_DRAWER_KEY_TRANSITION = env.get_credential('DIGITALGOV_DRAWER_KEY_TRANSITION', '')
+DIGITALGOV_BASE_API_URL = 'https://i14y.usa.gov/api/v1'
+DIGITALGOV_DRAWER_HANDLE = 'main'
```
ccostino: Not a showstopper, but I'm wondering if these should also be pulled from environment variables too?

fec/search/tests/test_indexing.py
```diff
@@ -0,0 +1,193 @@
...
+    def test_update_document(self, scrape, m):
+        # It makes a PUT to the update endpoint
+        m.register_uri('PUT', 'http://localhost:3000/documents/123', status_code=200)
+        search.update_document(self.page)
+        self.assertTrue(m.called)
+
+
+    @requests_mock.Mocker()
+    def test_delete_document_in_prod(self, scrape, m):
+        # It makes a PUT to the update endpoint in production
```
ccostino: "It makes a `DELETE`..." :-)

fec/fec/settings/base.py
noahmanger: Yeah, probably makes sense, that way they're shared by the manual script.

ccostino: I see the change below, but I meant these two values here specifically coming from environment variables too. :-)

noahmanger: Ah I was confused. I think these can actually be hard-coded.

PR #1086, opened 2017-05-25
Hotfix/meeting template item audio

johnnyporkchops:
Item audio file block was just returning name of file:
{{ block.value.item_audio }}, so I changed them to{{ block.value.item_audio.file.url }}, which targets the URL value AND forces it to play in browser instead if just download.

Meeting transcript block was just set to block:value which does not work here. So I changed it to: 
{{ self.mtg_transcript_url }}

Edited some language in the streaming media panel per request from Greg Scott:
"Watch the meeting live at 10:00 a.m. (EDT)"
We could incorporate the meeting time block into this b/c the meeting times do change, but I did not do this yet.

### DIFF
--- a/fec/home/templates/home/meeting_page.html
+++ b/fec/home/templates/home/meeting_page.html
@@ -49,7 +49,7 @@
           <a class="u-no-border" href="{{ self.live_video_url }}"><img src={% static 'img/thumbnail--video.png' %} alt="Video"></a>
           </div>
         <div class="usa-width-two-thirds">
-          <h3 class="u-no-margin"><a href="{{ self.live_video_url }}">Watch the live meeting now</a></h3>
+          <h3 class="u-no-margin"><a href="{{ self.live_video_url }}">Watch the meeting live at 10:00 a.m. (EDT)</a></h3>
           <p class="u-no-margin">This meeting is currently being streamed{% if self.live_video_captions %} with <a href="{{ self.live_video_captions }}">live captions</a>{% endif %}. The recorded video, audio and captions will be posted on this page after the meeting ends.</p>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


fec/home/templates/home/meeting_page.html
```diff
@@ -49,7 +49,7 @@ <h1 class="heading__left">{% formatted_title self %}</h1>
           <a class="u-no-border" href="{{ self.live_video_url }}"><img src={% static 'img/thumbnail--video.png' %} alt="Video"></a>
           </div>
         <div class="usa-width-two-thirds">
-          <h3 class="u-no-margin"><a href="{{ self.live_video_url }}">Watch the live meeting now</a></h3>
+          <h3 class="u-no-margin"><a href="{{ self.live_video_url }}">Watch the meeting live at 10:00 a.m. (EDT)</a></h3>
           <p class="u-no-margin">This meeting is currently being streamed{% if self.live_video_captions %} with <a href="{{ self.live_video_captions }}">live captions</a>{% endif %}. The recorded video, audio and captions will be posted on this page after the meeting ends.</p>
```
noahmanger: Should the "currently" phrasing also change?

johnnyporkchops:
Good point @manger re the current language. Additionally, maybe I Should go back to Greg and just get an ok to change the wording so there's no need to hard code the time or try to pull it from the field since its listed below that?

noahmanger:
I guess it depends how you see yourselves using this. Do you think you'll only add the video links while the meeting is running?

johnnyporkchops:
We add them in the morning a few hours before the meeting usually, plus the EDT changes too so it's probably better that way, I will check eeith Greg

johnnyporkchops:
I corrected the streaming media language.

PR #1105, opened 2017-06-12
Feature/indent outdent

noahmanger:
This adds support for indenting/outdenting lists in Wagtail so that you can create nested lists:

image

Here's how it looks in the editor:
image

There's no built-in icons for indent/outdent, so I just used the arrow icons (the ones on the far right of the toolbar). They have hover title text that say "indent" or "outdent".

Resolves https://github.com/18F/fec-style/issues/693

### DIFF
--- a/fec/fec/static/css/customize-editor.css
+++ b/fec/fec/static/css/customize-editor.css
@@ -24,3 +24,46 @@ blockquote {
 	 margin-left: 0;
 	 font-style: italic;
 }
+
+.richtext ul ul,
+.richtext ul ol,
+.richtext ol ol,
+.richtext ol ul {
+  margin-left: 15px;
+}
+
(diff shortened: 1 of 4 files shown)


jenniferthibault:
I'm not sure what's possible here, but do we have control over the order that these buttons appear in? It would be helpful to group them with the bullet/list icons, since they control their positioning, instead of separating the two to opposite ends of the toolbar.

patphongs:
@noahmanger thanks for taking this on. It looks really good, just reviewed and added a minor commit to trigger the outdent for ordered lists also.

noahmanger:
Thanks @patphongs . Good catch!

And yeah, I don't think we really have control over where the buttons are placed, since we're just adding them onto the default menu bar.

PR #1108, opened 2017-06-14
Feature/updates-search

noahmanger:
This adds the correct search functionality to all latest updates by adding the necessary body fields to the built-in Wagtail search index. Previously only titles were searched.

In the case of no results, it also adds a no search results message.

image

I added a line about how Record articles before 2011 aren't included in these results since they're only in PDF form, but could use @jameshupp eyes on microcopy.

This also fixes an issue where searching by a combination of filters would result in an error (e.g. year and keyword). 

Resolves https://github.com/18F/fec-cms/issues/1102

### DIFF
--- a/fec/home/views.py
+++ b/fec/home/views.py
@@ -89,6 +89,10 @@ def get_meetings(category_list=None, year=False, search=None):
 
     if year:
         meetings = meetings.filter(date__year=year)
+
+    if search:
+        meetings = meetings.search(search)
+
     return meetings
 
 
--- a/fec/home/views.py
+++ b/fec/home/views.py
@@ -140,6 +140,8 @@ def updates(request):
         meetings = MeetingPage.objects.live()
 
         if year:
+            # Trying to filter using the built-in date__year parameter doesn't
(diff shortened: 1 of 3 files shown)


fec/home/views.py
```diff
@@ -33,7 +34,8 @@ def get_records(category_list=None, year=None, search=None):
             records = records.filter(category=category)
 
     if year:
-        records = records.filter(date__year=year)
+        year = int(year)
+        records = records.filter(date__gte=datetime(year, 1, 1)).filter(date__lte=datetime(year, 12, 31))
```
noahmanger: For some reason, using the built-in `__year` filter was resulting in errors when combined with calling `.search()`. Switching to `__gte` and `__lte` solves it. Uglier but it works.

jameshupp:
@noahmanger Just made a few small tweaks, but it looks good to me!

fec/home/views.py
xtine: Can you make a comment about this in case a later refactor tries to go back to using `__year`?

noahmanger: Good call.

xtine:
👍

PR #1118, opened 2017-06-26
WIP- contribution limits snippet block

johnnyporkchops:
## This PR is replaced by [1193](https://github.com/18F/fec-cms/pull/1193/files)
This creates the contribution limits chart as a Snippet in Wagtail. 
https://github.com/18F/fec-cms/issues/1093 This allows the page editor to embed the contribution limits chart as as snippet within in the sections of a ResourcePage template.
The Snippet can be edited in one place and changes will propagate throughout the site wherever it is included.
This is currently included as a block in the ResourceBlock section of the ResourcePage template. 

This is the first use of Snippets in the CMS. Since this is a WIP PR, I will mention other ways to add snippets to a page template:
**Snippet chooser block:**

``` ('choose_snippet', SnippetChooserBlock('home.ContributionLimitsChart')),```

**Or snippet chooser panel:**

```snippet_choose = models.ForeignKey(
        'FECSnippets',
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+'
    ) 

Then in  ResourcePageModel:
 SnippetChooserPanel('snippet_choose'), ```
``` ('choose_snippet', SnippetChooserBlock('home.ContributionLimitsChart')),```
**Or snippet chooser panel:**
```snippet_choose = models.ForeignKey(
        'FECSnippets',
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+'
    ) 

Then in  ResourcePageModel:
 SnippetChooserPanel('snippet_choose'), ```
```

The snippet chooser panel would not work for the ResourcePage because the issue for this PR requires the page editor to be able to  intersperse the chart within sections of the ResourceBlock and a panel cannot be put inside the ResourceBlock, but a snippetChooserBlock can be put there.

Each registered snippet class has its own model fields and editor panels associated with it. So we could make a more generic FECSnippets model that has a more versatile RichText fields etc. and then a user could choose from several snippets in a ChooserBlock (or panel) and we would control what fields are rendered and how, in the template for the page.

Otherwise, we can register snippet models as needed based on the type of fields applicable for a particular piece of content and manage those as separate snippet classes in the Snippets area of the CMS. I personally think this is the most efficient solution.


noahmanger:
Thanks for digging in on this and sharing a few of the options. That all makes sense. I like the idea of using the Snippet Chooser so that we don't need to add a whole bunch of custom `StaticBlocks` on individual page models. 

I also totally see the value in one or more generic snippets. If you think that makes sense to include here, go for it. I know @dorothyyeager has said that it could be useful.

Also, if you think that the contributions limit one would benefit from a more custom snippet that had different fields to make it easier to edit, that could make sense too (I believe you suggested this already). That way the HTML could just live in a template file and content editors wouldn't need to worry about it.

johnnyporkchops:
@noahmanger , I think I can rename the snippet class to FECSnippets and then the contribution limits snippet can be chosen from there. Other basic snippets could be added there as needed.

I think if we want to make a more versatile Snippet class with more field options, that could be done with another class. But I am wondering if this is shoehorning Snippets into something that they are not that suited for. In  the documentation it says "Decide carefully if the content type you would want to build into a snippet might be more suited to a page." I think this is because it is generally designed for mostly basic content in the form of an html "snippet". As of Wagtail 1.9, I understand you can add streamfields to snippets, which would make it more versatile, but that would take some developing.

I think we should register a basic FECSnippets class (title, content fields) to include the contributions limits table and then consider developing the more versatile snippet class going forward. What is your take on this?

noahmanger:
Sounds good to me!

johnnyporkchops:
There is also the second part of this issue which is making the contributions limit table work in pages with sidebar and related links,

johnnyporkchops:
This is not ready for merge, we still need to change the model and blocks to reflect the decisions made in the conversations above.

Issue #1140, opened 2017-07-17
Pretty URLs for documents uploaded through Wagtail


PaulClark2:
Make documents uploaded through Wagtail have pretty FEC.gov URLs instead of AWS S3 bucket URLs.

patphongs:
Django documentation for django-storages stipulates that we can use `AWS_STORAGE_BUCKET_NAME` to specify a custom URL. But not sure if this is directly associated with the bucket name that is set in AWS (right now we can't set our own bucket name in cloud.gov). That's why we set routes for the bucket within the proxy. Would be good to investigate further on what this would do.

johnnyporkchops:
So the documents  uploaded through wagtail and in the documents folder start out with a www.fec.gov/ url but are redirected to the S3 bucket. Documents in the /resources/updates/agendas/2017/ stay as pretty urls and do not redirect to an AWS URL.

patphongs:
I'm hoping that if we specify the `AWS_STORAGE_BUCKET_NAME` in the base.py, it will remove the bucket name string `https://s3-us-gov-west-1.amazonaws.com/cg-47928592-406c-4536-8234-99b896e8d57d` and replace it with this `https://www.fec.gov/resources/`. 

For example, this URL 
https://s3-us-gov-west-1.amazonaws.com/cg-47928592-406c-4536-8234-99b896e8d57d/cms-content/documents/mtgdoc_17-29-a.pdf

would be transformed into 
https://www.fec.gov/resources/cms-content/documents/mtgdoc_17-29-a.pdf

noahmanger:
Resolved by https://github.com/18F/fec-cms/pull/1143

PR #1143, opened 2017-07-20
Implemented Custom Domain Path for CMS Uploads

patphongs:
This resolves issue #1140 

Note that this will require an additional environmental variable added to set the custom path variable `CMS_AWS_CUSTOM_DOMAIN=www.fec.gov/resources`

This PR allows us to maintain user friendly URLs when accessing content uploaded to CMS.


ccostino:
New environment variable is now set in all three environments, merging!

PR #1145, opened 2017-07-21
Multiple numbered, amended Sunshine Act notices

johnnyporkchops:
Issue https://github.com/18F/fec-cms/issues/1144
Show multiple amended Sunshine Act notices with numbers. One change made to the meeting_page.html template is to show the notices in the order they are added to the field rather than reversed (newest first). If I left them reversed it was over-complicated to get the order correct, but adding the numbers solves any confusion with that on the rendered page.
## **New:**
new_multiple_sunshines
new_multiple_sunshines_upates

### DIFF
--- a/fec/fec/templates/partials/update.html
+++ b/fec/fec/templates/partials/update.html
@@ -40,13 +40,9 @@
         {% for link in update.sunshine_act_links|splitlines %}
           <li>
             <a class="pdf-link" href="{{ link }}">
-              {% if forloop.counter == 1 %}
-              Sunshine Act Notice
-              {% else %}
-              Amended Sunshine Act Notice
-              - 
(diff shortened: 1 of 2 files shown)


noahmanger:
This is great! Good catches and solutions. My only suggestion is what about if there's only one amended notice, we leave off the number? I think the number is useful if there's more than one amendment, but without it reads a little confusing to me.

johnnyporkchops:
Good Point! I think I can make that change and update.

johnnyporkchops:
Updated to show numbers for amended notices only if there is more than one.

johnnyporkchops:
## > 1 && < 3 (or 2) ;-)
multiple_sunshines_ 1 3
## >= 3
multiple_sunshines_ 3

PR #1165, opened 2017-07-28
Feature/home page version 2

xtine:
New home page design. https://github.com/18F/fec-cms/issues/1137

Depends on https://github.com/18F/fec-style/pull/749

screencapture-127-0-0-1-8000-1501716757169

screen shot 2017-08-02 at 4 33 00 pm


xtine:
Mobile:
screencapture-127-0-0-1-8000-1501285904799

jenniferthibault:
## Updated banner designs
This style will accommodate more than one banner announcement. The pattern `message--mini` exists and can be seen on candidate profile tables here: https://www.fec.gov/data/candidate/P00003392/?election_full=True&cycle=2016&tab=raising#individual-contribution-transactions

**Desktop single & double**
screen shot 2017-08-02 at 11 02 23 am


**Mobile**
screen shot 2017-08-02 at 11 03 55 am

## Featured/pinned item logic
To start out, I think it might be good to limit to one featured item at a time. (Remembering our principle that if everything is highlighted, nothing is highlighted). Since we're enabling the banner feature for more time-sensitive announcements, I think rotating through one featured publication at a time is good practice to start and should meet known needs now.

**When there isn't a featured item, the categories are fixed as follows:**
- The most recent FEC record article
- The most recent Tips for treasurers article
 - The most recent press release
 - The most recent weekly digest

**When there is a featured item, the weekly digest item gets bumped out of the display**

## Hero image styling
I had gotten some initial feedback from Noah that encouraged us to make the hero photo shorter, as you've implemented here. Seeing it though, I don't think our image works in that space well, and is cropped too far. This is meant to be a fully visual image, not a textural slice like on our landing pages. I'd like to return to a more full-sized image that follows the [USWDS pattern](https://feedbackmvp-demo.app.cloud.gov/) more closely and retains more breathing space around it. The goal is to not crop the seal or the chairs:

screen shot 2017-08-02 at 11 18 15 am

fec/home/templates/partials/home-page-banner-announcement.html
```diff
@@ -0,0 +1,18 @@
+{% load wagtailcore_tags %}
+
+<section class="slab slab--secondary">
```
noahmanger: I think this should have an `if` here so that an empty banner doesn't show if there's nothing there:
image

fec/home/models.py
```diff
@@ -491,17 +491,26 @@ def get_author_office(self):
...
 
     content_panels = Page.content_panels + [
-        FieldPanel('link'),
-        FieldPanel('homepage_expiration'),
+        FieldPanel('description'),
+        FieldPanel('link_title'),
+        FieldPanel('link_url'),
+        FieldPanel('link_title_2'),
+        FieldPanel('link_url_2'),
+        FieldPanel('date_active'),
```
noahmanger: What's the purpose of `date_active` field? Looks like it doesn't matter if it's set in the future or not.

fec/home/templates/home/home_page.html
```diff
@@ -7,98 +7,53 @@
...
-          <div class="grid__item">
-            <h3>Learn who's giving to candidates and committees</h3>
-            <a class="t-sans" href="{{ settings.FEC_APP_URL }}/receipts/individual-contributions">Look up an individual contributor »</a>
-          </div>
-        </div>
+      <ul class="grid grid--4-wide grid--flex t-sans">
+        <li class="grid__item grid__item--event">
+          <div class="t-bold">Next commission meeting:</div>
+          <div class="js-next-commission-meeting"></div>
+          <a class="button button--alt button--updates" href="/calendar/?category=Executive+Sessions&category=Open+Meetings&category=Public+Hearings">All meetings</a>
```
noahmanger: This link should actually go to the meetings _agenda_ page: `/updates/?update_type=meetings`

fec/home/models.py
xtine: I could name it better, but will sort the banner announcements by date for the ordering of more than one banner announcement.

noahmanger: Ah ok. That works for me.

PR #1172, opened 2017-08-03
Adding Folder page type

noahmanger:
This is testing out an idea I had to make Wagtail easier to navigate. The idea is to introduce a new page model called `Folder` that just has a title and a field for a URL to redirect to:
image

Because every Wagtail page has a unique URL, this approach is necessary in order to allow us to nest pages in "folders" without also having those folders be pages. So for instance, making a folder for press releases and then moving all press releases into it would result in URLs like `fec.gov/updates/press-releases/some-press-release/`. But `fec.gov/updates/press-releases/` doesn't exist on its own, so making it as a folder page would allow us to put a URL to redirect to, so that it could, say go to `fec.gov/updates?type=press+releases`. 

The one thing that seems tricky with this is that the URLs to all the updates pages will change. Maybe @patphongs has an idea for how we could deal with this from a redirect perspective?

If we do go this route, it's easy to bulk move pages using the Django shell using `Page.move(new_parent_page, 'first-child')`.


noahmanger:
Pinging @patphongs . Can you take a look and let me know what you think about this idea?

PR #1193, opened 2017-08-14
Snippet to embed contribution limits table

johnnyporkchops:
So  that page editors can embed the current contribution limits table in pages
Maintain the table in one location, where changes would be reflected globally, wherever the table is added
Issue https://github.com/18F/fec-cms/issues/1093

After much discussion and a syc with design and content folks, this PR is ready for review to merge.
History of conversations are in https://github.com/18F/fec-cms/pull/1153 and https://github.com/18F/fec-cms/pull/1118. (This PR replaces 1153 and 1118 to avoid stale migration conflicts)

embed-table-snippet

The reason for the extended discussions is that we wanted to make sure that when adding the first use of [Wagtail Snippets](http://docs.wagtail.io/en/v1.11.1/topics/snippets.html) to the CMS, we were doing it in a way that would be most intuitive for editors and also allow the use of additional snippets in the future.

Responsive styles created and merged to fec-style still apply and do not need to be adjusted. https://github.com/18F/fec-style/pull/746


fec/home/models.py
```diff
@@ -1047,3 +1050,22 @@ class ReportingExamplePage(Page):
     @property
     def content_section(self):
         return 'help'
+
+@register_snippet
+@python_2_unicode_compatible  # provide equivalent __unicode__ and __str__ methods on Python 2
```
noahmanger: What does this decorator do? Is it necessary?

fec/home/models.py
```diff
@@ -1047,3 +1050,22 @@ class ReportingExamplePage(Page):
...
+class EmbedTableSnippet(models.Model):
+    title = models.TextField()
+    description=models.TextField()
```
noahmanger: Small thing, but missing spaces around the `=`.

fec/home/models.py
johnnyporkchops: It is so that the code supports Python 2, I assume it was included in the Snippets docs since you depend on returning the string representation of the fields to show the name of  the snippet in snippets area. I can take it out since we are not using Python 2 and we use the __str__ method elsewhere without using the decorator.https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.encoding.python_2_unicode_compatible

noahmanger: Cool. Yeah I think it's probably ok to take it out.

PR #1200, opened 2017-08-15
Adding custom admin panels for news and updates

noahmanger:
This uses Wagtail's `ModelAdmin` hooks to implement custom admin panels for the latest updates. 

Now, there will be a new menu item for "News and updates", which contains links to each type of update:
image

Each one of those takes you to a view that is default sorted by the original date of the item:
image

I think this is actually a much better approach than the "Folder" concept earlier as it doesn't require any redirects or moving pages around.

### DIFF
--- a/fec/home/wagtail_hooks.py
+++ b/fec/home/wagtail_hooks.py
@@ -1,20 +1,8 @@
-from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
+from wagtail.contrib.modeladmin.options import ModelAdmin, ModelAdminGroup, modeladmin_register
 from wagtail.wagtailcore import hooks
 
-from .models import Author
+from .models import Author, PressReleasePage, DigestPage, TipsForTreasurersPage, RecordPage
 from search.utils.search_indexing import handle_page_edit_or_create, handle_page_delete
-
-class AuthorAdmin(ModelAdmin):
-    model = Author
-    menu_icon = 'user'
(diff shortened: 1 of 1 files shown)


fec/home/wagtail_hooks.py
```diff
@@ -1,32 +1,71 @@
...
+    menu_label = 'News and updates'
+    menu_icon = 'folder-open-inverse'
+    menu_order = 200
+    items = (PressReleaseModelAdmin, DigestPageModelAdmin, TipsForTreasurersPageModelAdmin, RecordPageModelAdmin)
+
+modeladmin_register(AuthorAdmin)
+modeladmin_register(PressReleaseModelAdmin)
+modeladmin_register(DigestPageModelAdmin)
+modeladmin_register(TipsForTreasurersPageModelAdmin)
+modeladmin_register(RecordPageModelAdmin)
```
ccostino: It looks like these are extra `register` statements that aren't needed...  you only need the last one for the menu itself to show up with the sub items in it.  What I found was that with these statements in place, you see them in the main menu too:

screen shot 2017-08-16 at 1 03 18 am

Where as if you just keep the one below, they go away and you have just the one menu item. :-)

noahmanger:
Good catch! I made that fix earlier but apparently forgot to push it. There it is.

dorothyyeager:
Thank you @ccostino @noahmanger

PR #1201, opened 2017-08-15
Fix logging of user info on page delete

noahmanger:
Currently, trying to delete a page in Wagtail can result in a server error if there is not a `user_id` in `kwargs['instance']`. This just updates the logic a to be a little more forgiving if that's the case. 

Can @ccostino or @LindsayYoung double-check this?

Resolves https://github.com/18F/fec-cms/issues/1189

### DIFF
--- a/fec/home/models.py
+++ b/fec/home/models.py
@@ -188,12 +188,15 @@ def log_user_save(sender, **kwargs):
 @receiver(pre_delete, sender=PageRevision)
 @receiver(post_save, sender=PageRevision)
 def log_revisions(sender, **kwargs):
-    try:
-        user_id = int(kwargs.get('instance').user_id)
-        user = User.objects.get(id=user_id)
-    except User.DoesNotExist:
+    user_id = kwargs.get('instance').user_id
+    if user_id:
+        try:
+            user = User.objects.get(id=user_id)
+            logger.info("page was modified: {0} by user {1}".format(kwargs.get('instance'), user.get_username()))
+        except User.DoesNotExist:
+            logger.info("User not found")
+    else:
         logger.info("User not found")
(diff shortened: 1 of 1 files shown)


ccostino:
I'll take a look, thanks @noahmanger!

fec/home/models.py
```diff
@@ -188,12 +188,15 @@ def log_user_save(sender, **kwargs):
...
-        user = User.objects.get(id=user_id)
-    except User.DoesNotExist:
+    user_id = kwargs.get('instance').user_id
+    if user_id:
+        try:
+            user = User.objects.get(id=user_id)
+            logger.info("page was modified: {0} by user {1}".format(kwargs.get('instance'), user.get_username()))
+        except User.DoesNotExist:
+            logger.info("User not found")
+    else:
```
ccostino: This is a really good approach at fixing the immediate issue with this code, thanks @noahmanger!

However, I'm a bit concerned about two underlying issue here:

* The `user` and `user_id` properties of the `instance` can be `None` (presumably if that user was removed from the database prior to any action being taken on a page)
* A copy or delete action of a page results in multiple `page modified...` log statements not associated with the user currently logged in performing the action, but instead with the user who is attached to a given revision; however, the last action logged is the one associated with the user currently performing the action - this is because the action is repeated against every revision of the page, not just the current one

What's interesting is that creating or editing a page results in only one log statement, compared to the note about the copy or delete actions.  In any case, this means we're seeing multiple log statements that aren't exactly accurate since they don't represent the current action being taken, which could be confusing (they do, however, maintain their original date and time, so they're not exactly wrong, either).

I'd also imagine we still want a record of the action too even if a user is not found associated with it.  Wagtail is doing some of its own logging under the hood as it is so we're partially covered with that, but we could probably shore this up a bit too to still capture something else of use, even without a user.

PR #1205, opened 2017-08-16
Feature/boolean search

noahmanger:
This adds the boolean search modal to the legal resources landing page. 

how it works

Users can click the "more keyword options" button which opens a modal window with four fields. When they fill in those fields, we transform the four fields into a single search string with all the necessary parentheses and boolean operators, which will then be parsed by the API.


noahmanger:
Update: the separate parenthetical clauses are now joined by the proper `OR` instead of `+`:
image

noahmanger:
The tests are passing now. @xtine can you check it out?

PR #1209, opened 2017-08-22
Meeting refinements

noahmanger:
This includes a few meeting refinements after pairing with @jenniferthibault . 

image

- Search inputs are now wider
- There's no search input on exec sessions
- The list now uses the `title` field to display and I added a filter for stripping away the words `open meeting` or `executive session` so that you can include the "cancelled" status in the meeting title

Requires an fec-style change

### DIFF
--- a/fec/fec/templates/partials/meeting.html
+++ b/fec/fec/templates/partials/meeting.html
@@ -2,7 +2,7 @@
   <tr class="simple-table__row">
     <td class="simple-table__cell">
       <a href="{{ meeting.url }}" class="t-bold">
-        {% if hearing is meeting %} {{ meeting.title }}{% else %}{{ meeting.date|date:'F j, Y' }}{% endif %}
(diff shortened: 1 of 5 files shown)


jenniferthibault:
Thanks @noahmanger ! 👍 visual styling review, code review needed.

jjones0618:
This new design appears to have sufficient room to add a third column which could easily contain two more links, one for the full meeting audio recording and one for the video recording. This addition makes sense as many people searching for past meetings are looking for either the audio or the video. Why not save them a click if possible? Here's my suggestion:

av mockup

johnnyporkchops:
@jjones0618, Thanks for your comments and suggestions! I think we should share your mockup in one of the regular design parings. They are on Fridays at 2:00. I forwarded the event from outlook to you. I will follow up with you  in slack as well.

johnnyporkchops:
It also is a good idea to create a new github issue for this. https://github.com/18F/fec-cms/issues

PR #1220, opened 2017-08-29
[MERGE WITH GITFLOW] Hotfix/revert model admin

noahmanger:
This reverts the custom admin panels for news and updates introduced in https://github.com/18F/fec-cms/pull/1200 . There is a mysterious bug in production that's preventing content managers from being able to create new updates posts. The bug is not reproducible locally. Hopefully this solves the current problem but we can follow up and restore the functionality fully.

### DIFF
--- a/fec/home/wagtail_hooks.py
+++ b/fec/home/wagtail_hooks.py
@@ -1,8 +1,20 @@
-from wagtail.contrib.modeladmin.options import ModelAdmin, ModelAdminGroup, modeladmin_register
+from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
 from wagtail.wagtailcore import hooks
 
-from .models import Author, PressReleasePage, DigestPage, TipsForTreasurersPage, RecordPage
+from .models import Author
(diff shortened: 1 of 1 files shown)


noahmanger:
Nevermind. This isn't going to work.

fec/home/wagtail_hooks.py
```diff
@@ -1,67 +1,32 @@
-from wagtail.contrib.modeladmin.options import ModelAdmin, ModelAdminGroup, modeladmin_register
+from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
 from wagtail.wagtailcore import hooks
 
-from .models import Author, PressReleasePage, DigestPage, TipsForTreasurersPage, RecordPage
+from .models import Author
 from search.utils.search_indexing import handle_page_edit_or_create, handle_page_delete
 
-
-@hooks.register('after_create_page')
```
LindsayYoung: Wanted to confirm removing this hook won't interfere with logging . We will still get statements when objects are created in prod, right? We need that for ATO requirements

PR #1225, opened 2017-08-29
[OLD WIP] Feature/cms events

noahmanger:
This is an old branch I started a while ago when I was scoping what it would take to use the Wagtail for managing calendar data. Basically it:

- Creates a new `cal` app with a custom `Event` model
- Sets up Wagtail's built-in REST support
- Updates the JS to use data provided from Wagtail

Things that I didn't get to:

- Set up the ability to assign multiple states to events
- Ensure proper formatting of the feeds for Google / iCal
- Figure out how to get current/old calendar event data in here
- Figure out what if/how this would support other calendar data needs

Making this a PR to share my work before I go so someone can pick it up in the future.

cc @LindsayYoung @patphongs


lbeaufort:
Old WIP PR

PR #1227, opened 2017-08-30
Feature/data tests

noahmanger:
This moves over all of the js unit tests from web app and fec-style to the unification branch.


noahmanger:
Ok @xtine got all of the tests working.

PR #1228, opened 2017-08-30
Feature/meeting sunshine doc upload

johnnyporkchops:
Issue: https://github.com/18F/fec-cms/issues/1110
-Adds the ability to upload documents for Sunshine notices with DocumentChooserBlock (instead of typing in URLs to a TextField)
-Template logic says if there are Sunshine notices in the old format (TextField), show them , else show the uploaded documents -(updated, see convo below https://github.com/18F/fec-cms/pull/1228#issuecomment-326115972)
-Hide the old  TextField in the editing interface for meeting page template
-Recreate the logic too specify amended and add numbers  if there are more than one amended
-fix breadcrumb link on meeting_page.html template ( link was still going to /updates?type=meeting)
-Migration included
<br>

 

sunshine_uploads


johnnyporkchops:
Rethinking the template logic:
Now, if there are notices in the old format (TextField), that is all that will show in the rendered template. So if one needs to change or add a notice for a past meeting, it is inaccessible to the editor.
If I cold change to say, in effect: 
``` "If there are only one or the other, show the one that exists, if there are both, then only show the new format. ```
That way, someone can just recreate the list (with changes) in the new format and have the old one remain hidden. The only problem with this is that there will be old notices saved somewhere in the database for this meeting. But maybe that is not an issue to be concerned about

**Comment by noahmanger** [comment]
This looks sound to me. Nice way to deal with the old docs. My $.02 is I wouldn't worry about having duplicate entries. Doesn't seem worth the trouble.

As far as the logic you're after, if you just switch the order of the conditionals, that should do it:

```
{% if meeting.sunshine_act_doc_upld %}
...
{% elif meeting.sunshine_act_links %}
...
{% endif %}
```

That way if there's the new doc uploads, it'll just show those and stop there.

johnnyporkchops:
Ahh, perfect, thanks!

PR #1236, opened 2017-09-06
[MERGE WITH GITFLOW] Hotfix - RAD form link fixes

patphongs:
Resolves https://github.com/18F/fec-cms/issues/1234

### DIFF
--- a/fec/home/templates/home/contact-form.html
+++ b/fec/home/templates/home/contact-form.html
@@ -119,7 +119,7 @@
                   </div>
                   {{ field.errors }}
                 {% elif field.id_for_label == 'id_u_committee_member_certification' %}
-                  <p><a href="#" class="t-sans">How FEC.gov keeps personal information private</a></p>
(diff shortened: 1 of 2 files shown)


jenniferthibault:
@patphongs a quick tip I picked up from @LindsayYoung last week is that the PR title for hotfixes should start with `[MERGE WITH GITFLOW] Hotfix/title.....` so that no one accidentally merges it through Github.

I don't know/have gitflow, so I won't be able to merge, but I thought it was helpful to learn the reason behind the label I had seen before!

jenniferthibault:
I also just realized that you might have already known that, and would have changed it when it wasn't `[WIP]` anymore..... Sorry if this is redundant info!

patphongs:
Thanks for the tip @jenniferthibault ! I was working on this hotfix with Laura B. and we put in WIP there for now until we can get a language clarification for the contact form link (already slacked Lauren Lien about it). Will add that tag to merge with gitflow after it's ready!

PR #1256, opened 2017-09-20
Boolean logic fix and webapp variable updates

patphongs:
This resolves  https://github.com/18F/openFEC-web-app/issues/2291 and https://github.com/18F/fec-cms/issues/1257

- Any quotes that are inserted into the inputs will be removed. Quotes were complicating the query parameters when they did not need to be there.
- removed references to `webAppUrl` variable, no longer needed after merge


PaulClark2:
I have a question about removing the quotes. If you remove the quotes does the query still treat the quotes phrase as a single unit? In other words if I type "independent expenditure," I want results that include the phrase "independent expenditure" not just the two words "independent" and "expenditure."

patphongs:
@PaulClark2 With the advanced search, each box is designed to handle either single words or a single phrase at the moment. So you're really not supposed to put quoted phrases within these inputs: `any of these words`, `all of these words`, or `none of these words`, because it's only designed to take in single words. To get an exact phrase match, you would need to insert that into the `This exact phrase` input box. 

There will probably be more refinements that we will need to make such as putting in multiple phrases into the exact phrase input and also retaining your query within the advanced search box after submission. But this might be a good enough initial implementation I think.

screen shot 2017-09-20 at 7 26 03 pm

PR #1260, opened 2017-09-20
Legal search

xtine:
Migrate search functionality on Legal resources landing page.


fec/legal/templates/partials/legal-search-results-advisory-opinion.jinja
```diff
@@ -8,7 +8,7 @@
 {% for advisory_opinion in advisory_opinions %}
   <h3 class="ao-result__title">
     <a title="{{ advisory_opinion.description }}"
-      href="{{ url_for('advisory_opinion_page', ao_no=advisory_opinion.no) }}">
+      href="/data/legal/advisory-opinions/{{ advisory_opinion.no }}">
```
tbaxter-18f: Why the preference for hard-coding the URL vs using the URL tag and named urls?

xtine: Long story short, the data app within this django project was migrated from a flask app that used Jinja as a templating language. One method to migrate the app from flask to django within an efficient timeframe was to hardcode URLs as they weren't going to change in the future and `url_for` is not compatible in Django or the `django_jinja` package we are using.

PR #1263, opened 2017-09-21
Remove .python-version file

ccostino:
This changeset removes the `.python-version` file from the repository for the time being.  We were never able to move forward on the `pip-tools` install, so this is not needed at the moment and makes dealing with Python updates locally a bit more cumbersome at the moment.  We can bring this back once we have a chance to work on Python dependency management a bit more and provide clear documentation on how to manage these things.

In the interim, if you are using `pyenv` to manage your local Python installations, please be sure you have a global version set, e.g., `pyenv global 3.5.4`.


ccostino:
Closing this PR in favor of making the change in PR #1253.

PR #1265, opened 2017-09-21
CMS default to eastern time zone

LindsayYoung:
put the CMS in the eastern time zone


LindsayYoung:
Hopefully this will help with banners

PR #1270, opened 2017-09-22
Update CI and cloud.gov config to build with CircleCI

vrajmohan:
This changeset removes the old Travis CI config in favor of the new CircleCI config.

It also removes the .python-version file from the repository for the time being. We were never able to move forward on the pip-tools install, so this is not needed at the moment and makes dealing with Python updates locally a bit more cumbersome at the moment. We can bring this back once we have a chance to work on Python dependency management a bit more and provide clear documentation on how to manage these things.

In the interim, if you are using pyenv to manage your local Python installations, please be sure you have a global version set, e.g., pyenv global 3.5.4.


fec/fec/settings/base.py
```diff
@@ -1,3 +1,4 @@
+import locale
```
ccostino: Ah we can take this out now. :-)

fec/search/tests/test_indexing.py
```diff
@@ -90,8 +87,10 @@ def test_create_index_doc_edited(self, scrape):
 
     @requests_mock.Mocker()
     def test_add_document(self, scrape, m):
+        #import pdb
+        #pdb.set_trace
```
ccostino: D'oh!  I forgot to remove these.  At least they're commented out though! 😆

PR #1272, opened 2017-09-25
Fix deploy steps to run commands in proper working directory

ccostino:
This changeset fixes an issue we are having with the deploy step in the CircleCI environment.  It appears that the Django commands are not able to find where certain files are located, particularly those found in the `fec/dist` directory.

There is a larger issue here with the referencing to the path(s) of static assets, but we are fixing the immediate problem first.

Huge h/t to @tbaxter-18f and @cmc333333 for their help in debugging this issue!

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
     "build": "cd fec && webpack",
     "build-icons": "cd fec && gulp minify-icons && gulp consolidate-icons",
     "build-js": "cd fec && webpack",
+    "build-production": "cd fec && webpack --optimize-minimize --define process.env.NODE_ENV='production'",
     "test": "karma start",
     "test-single": "karma start --single-run --browsers PhantomJS",
     "watch": "nswatch"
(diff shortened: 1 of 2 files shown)


ccostino:
There are two other issues going on here that are at play:

* The static asset path(s) are being referenced incorrectly in some cases (`fec/data/templatetags/filters.py` and `fec/home/templatetags/filters.py` - both have lines looking for the `rev-manifest.json` files in `fec/dist/fec/static/js` folder)
* Static assets being managed by both Webpack and Django.

The first issue we can either fix here or immediately after depending on how quickly we need this to work, but that does need to be addressed.  As far as I (and @tbaxter-18f!) can tell, the settings themselves appear okay but referencing those JSON files with a hardcoded relative path is a good source of misery.

The second issue will likely need more time to work through as it is more involved.  But, I have to think there's a better way of managing all of these things that is less confusing and error prone. :-)

tbaxter-18f:
Looks good to me. I see what you mean about running in the wrong directory now.

package.json
```diff
@@ -3,6 +3,7 @@
     "build": "cd fec && webpack",
     "build-icons": "cd fec && gulp minify-icons && gulp consolidate-icons",
     "build-js": "cd fec && webpack",
+    "build-production": "cd fec && webpack --optimize-minimize --define process.env.NODE_ENV='production'",
```
xtine: This is a good idea! Didn't think of this and makes it able to test on local environment as well. 👍

PR #1273, opened 2017-09-25
Feature/banner announcement expiry

johnnyporkchops:
Issue: 
https://github.com/18F/fec-cms/issues/1267#
Adding ability to set  expiry for homepage banner announcements. Recognizing the time proves to be a challenge, while comparing date by full days works as usual. 

<s>Hate to pile on more PRs with migrations that will likely fail Circle , but might as well get them into the mix sooner than later</s>

FYI-Note: The go live and expiry date fields in the "settings" tab (not used in this PR)  seem to work for day -by-day control, but not for time-of-day control.


johnnyporkchops:
If we are indeed switching TIMEZONE in settings/base.py from "UTC" to "US Eastern" as I have heard Lindsay is possibly doing, then this will work, but otherwise, it is tricky to offset UTC with US Eastern in the logic.

fec/home/templatetags/home_page.py
```diff
@@ -1,5 +1,6 @@
 import re
 
+import datetime
```
tbaxter-18f: To fully comply with PEP-8 and Django code conventions, I believe you'd want this to be the first import, before re, right?

fec/home/templatetags/home_page.py
```diff
@@ -13,7 +14,9 @@
 
 @register.inclusion_tag('partials/home-page-banner-announcement.html')
 def home_page_banner_announcement():
-    banners = HomePageBannerAnnouncement.objects.live().filter(active=True).order_by('-date_active')[:2]
+    on = datetime.datetime.today()
+    banners = HomePageBannerAnnouncement.objects.live().filter(active=True,date_active__lte=on, date_inactive__gt=on).order_by('-date_active')[:2]
```
tbaxter-18f: I'm not seeing this live() method. Where is that coming from?

Also, have you considered creating an "active()" manager for the Announcements?

johnnyporkchops: Response to https://github.com/tbaxter-18f comment:
Live is the Wagtail QuerySet methods that returns live pages.http://docs.wagtail.io/en/v1.12.2/reference/pages/queryset_reference.html?highlight=objects.live
I have not considered an"active()" manager for the Announcements. Can you explain?
Also, wagtail core has default  built-in "go-live" and "expiry" DateTime fields in the setting tab in the editing interface.   However, these only seem to respond to days and not days and times n our current configuration. Since these announcements and another "Homepage Pin" feature we have implemented are not pages, we decided to handle them, for the time being, with custom fields in the content tab.

fec/home/templatetags/home_page.py
johnnyporkchops: I do not know, thanks for the prompt.

fec/home/templatetags/home_page.py
```diff
@@ -13,7 +15,8 @@
 
 @register.inclusion_tag('partials/home-page-banner-announcement.html')
 def home_page_banner_announcement():
-    banners = HomePageBannerAnnouncement.objects.live().filter(active=True).order_by('-date_active')[:2]
+    on = datetime.datetime.today()
```
xtine: Can this variable be named something more descriptive like `now` or `datetime_now`?

fec/home/templatetags/home_page.py
xtine: Also I would recommend spaces after each comma for readability.

fec/home/templatetags/home_page.py
johnnyporkchops: Yeah, it should be :-)

tbaxter-18f: "today" would be the most common convention for today, but that makes me wonder, would now = datetime.datetime.now() be better? Are you trying to get to an hour-by-hour deactivation of stale announcements?

fec/home/templatetags/home_page.py
tbaxter-18f: Thanks Johnnyporkchops. My experience with wagtail is limited, and I didn't even think about it exposing some extra managers. Makes total sense now. 

Making an active manager of your own would be basically the same thing. It would let you pre-filter for the active announcements. Depending on exactly how you set it up, you could then call something like Announcement.active_objects.all(). 

See https://docs.djangoproject.com/en/1.11/topics/db/managers/#modifying-a-manager-s-initial-queryset for some docs, or, if you'd like, I have some simple examples laying about I could link up to.

fec/home/templatetags/home_page.py
johnnyporkchops: They both work for hour-by-hour control of the annoucements, but I opted fordatetime.datetime.today() because I was not sure if now(), would cause a unexpected result:
https://stackoverflow.com/questions/14576967/datetime-today-vs-datetime-now

fec/home/templatetags/home_page.py
johnnyporkchops: I think this manager solution might be helpful for reusing announcements and "Pins" elsewhere, if we needed to expand there usage on the site.. The current solution allowed me the fewest edited files and simplest migrations

fec/home/templatetags/home_page.py
tbaxter-18f: Hm, that's interesting, but I think would be negated by the change from UTC to Eastern in settings. I think. 

It is a little worrisome seeing datetime fields being used like datefields. You're _probably_ ok ignoring time, but it feels a little wonky. 

If the goal is that they all become active (and expire) at midnight, give or take an hour or two, you should be fine (other than some misleadingly named fields). If you may want to publish or expire announcements mid-day, then I think this may be problematic. You know your needs better than I do.

fec/home/templatetags/home_page.py
tbaxter-18f: Agreed. It's really just a question of cleanliness, and forcing the desired filters to be consistent. And it would help keep your line length within PEP-8 constraints. But you're right, it may be overkill for a one-shot deal like this.

johnnyporkchops: PEP-8 and better adherence to it was brought up in yesterdays PI planning, including the issue on line-lengths. Thanks for bringing this onto our radar.

fec/home/templatetags/home_page.py
johnnyporkchops: The timezone in settings/base.py ended up being time was set to America/New_York. The fields work for hour-by-hour control --tested on local Django dev server. I'll keep an eye on it and circle back if things don't act as expected, once live. But yes, it some respects, it does not make sense based on docs for the app and Django field references.

johnnyporkchops:
I combined migrations so I am only committing one, as requested and made sure the home folder migration (0094) matches the same on develop.

johnnyporkchops:
@pat, the timezone has been changed to America/New_York already. You should not need to adjust the time anymore for UTC when using this feature.

PR #1282, opened 2017-09-29
[WIP] Feature/1237 homepage cfd section update

patphongs:
Implementing new infoblock design pattern as seen in this issue: https://github.com/18F/fec-cms/issues/1237


johnnyporkchops:
-@patphongs , I changed padding to heights as you, I, and @xttine decided in pairing. 
-Removed max-width on the .grid__item--infoblock  in _grids.scss and added a background-size: 100% to fix the offset problem with the horizontal rule above on the homepage and to allow it to fill the whole block in mobile view
- Added infoblock__icon styles for the icons and some specific styles for the form infoblock_image on H4CC londing
-  I think we can better achieve the desired cropping/resizing by actually using images and using `object-fit: cover;` instead of background images and background-size 100% or cover. But this will have to be for another iteration
-Also the scss could be consolidated better and needs to be linted

johnnyporkchops:
This PR has been combined with https://github.com/18F/fec-cms/pull/1290

patphongs:
Thanks @johnnyporkchops. Closing this in favor of the PR you merged: https://github.com/18F/fec-cms/pull/1290

PR #1300, opened 2017-10-03
MUR search fixes

xtine:
Separates the form actions so both MUR search fields should work properly now.

screen shot 2017-10-03 at 12 27 55 pm

### DIFF
--- a/fec/home/templates/blocks/mur_search.html
+++ b/fec/home/templates/blocks/mur_search.html
@@ -4,36 +4,40 @@
   <div class="slab slab--neutral slab--inline u-padding--left u-padding--right">
     <div class="heading--section heading--with-action">
       <h2 class="heading__left">Search closed MURs</h2>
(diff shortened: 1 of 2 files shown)


patphongs:
@xtine Added to your PR for the MUR number form to search by MUR number. I think it's good to go after that. Going to give it back to you for another review and then feel free to merge. Thanks!

xtine:
@patphongs: thanks for the update. Plus I finally figured out why specifically searching `mur_no` wasn't returning any results. In migrating code, it would only make a legal search query if there was a `search` request, so any `mur_no` requests were being ignored. I changed it so search results should always return results so now `mur_no` searches should work.

@jenniferthibault: this fix also makes the advanced search link link correctly. It just returns whatever a blank search request to the API returns, so I'm not sure what the default sorting is, but at least it returns results!

PR #1320, opened 2017-10-05
Adding sass compression using gulp script

patphongs:
This will remove django-jinja scss compression and add gulp script scss compression.


fec/data/templatetags/filters.py
```diff
@@ -148,11 +148,25 @@ def filesize(value):
 #     return cycles[0]
 
 @library.global_function
-def asset_for(path):
+def asset_for_css(key):
+    """Looks up the hashed asset key in rev-manifest-css.json
+    If the key doesn't exist there, then just return the key to the static file
+    without a hash"""
+
+    assets = json.load(open(os.path.join(settings.STATIC_ROOT, 'css/rev-manifest-css.json')))
```
xtine: So we are going to keep using `STATIC_ROOT` here? It would mean we would have to run `collectstatic` locally each time after a change. I was tinkering referencing directories in other commits so it would be able to point to the correct rev-manifest file within the codebase rather than copying it over to the directory that `STATIC_ROOT` is defined in.

.cfignore
```diff
@@ -6,14 +6,16 @@ bower_components
...
+.env
 
 # C extensions
 *.so
 
 # Packages
 *.egg
 *.egg-info
-dist
+# dist
```
xtine: Should we just remove this line entirely instead?

patphongs: Thanks @xtine I agree, it should be removed entirely

johnnyporkchops:
Great job!  Huge lift by the @patphongs and @xtine

PR #1323, opened 2017-10-10
Streaming info will show up automatically on meeting day(s) between 7:30am and 5:00pm

johnnyporkchops:
The streaming info block will show up on the day of meeting and disappear at end of day (7:30am to 5:00pm America/New_York).
1)  For the occasional meeting that spans two days, it will show up on both. (Some meetings start on one day and then continue on another day. The streaming info will NOT show up during the whole span between start and end date, but rather, **only on the start date and end date**. 
2) In unusual circumstances, the visibility of streaming info can be more finely controlled by editing the meeting date in Wagtail.
3) Streaming info can be removed, if necessary by unchecking `Show streaming info` checkbox in the meeting page template. (For example, on canceled meetings)
4) Should we need to point to a different URL for video and/or captions for some reason, the template editor can simply change the default URL(s) in Wagtail and the info they input will be used just for that meeting.


fec/home/templates/home/meeting_page.html
```diff
@@ -42,15 +42,19 @@ <h1 class="heading__left">{% formatted_title self %}</h1>
...
+    {% now 'F j, Y' as todays %}
+    {% if self.date|date:'F j, Y' == todays or self.end_date|date:'F j, Y' == todays %}
     <div class="row">
       <div class="slab slab--neutral u-padding--left usa-width-five-sixths">
         <div class="usa-width-one-fourth u-padding--left">
-          <a class="u-no-border" href="{{ self.live_video_url }}"><img src={% static 'img/thumbnail--video.png' %} alt="Video"></a>
+          {% comment %}
+           If there is live video or captions URLs provided in the template, use those. otherwise use default settings URLs
+          {% endcomment %}
+          <a class="u-no-border" href="{% if self.live_video_url %}{{ self.live_video_url }}{% else %}{{ settings.LIVE_VIDEO_URL }}{% endif %}"><img src={% static 'img/thumbnail--video.png' %} alt="Video"></a>
```
xtine: This should be a more descriptive alt tag.

fec/home/templates/home/meeting_page.html
```diff
@@ -42,15 +42,19 @@ <h1 class="heading__left">{% formatted_title self %}</h1>
       <p>The Commission meets regularly in executive sessions to discuss pending enforcement actions, litigation and other matters that, by law, must be kept confidential.</p>
     {% endif %}
 
-    {% if self.live_video_url %}
+    {% now 'F j, Y' as todays %}
```
xtine: Why is today plural?

johnnyporkchops: I think I thought it would be easier to distinguish between this as a variable and today as a datetime method if I had to write more complicated code, but I can change it to "today"

fec/home/templates/home/meeting_page.html
johnnyporkchops: OK, just I that is what was put in the original meeting_page.html template, I will change it

johnnyporkchops:
I also added time constraints per request from @dorothyyeager to match banner messages and follow the way we have done it manually in the past (7:30am to 5:00pm)

johnnyporkchops:
@xtine, now that the code added to the template for the streaming info has become more complicated, I have a feeling you might recommend moving it to a template tag and registering an inclusion tag to put it in (?)

xtine:
I don't think this is a use case for a template tag. Template filters and tags are meant to be reusable across templates, and I don't really see a case they would be used on another template. However you're right -- every time any more than basic logic creeps into the template, it means that logic should be moved into the view.

fec/fec/settings/base.py
```diff
@@ -37,6 +37,9 @@
 FEC_TRANSITION_URL = env.get_credential('FEC_TRANSITION_URL', 'https://transition.fec.gov')
 FEC_CLASSIC_URL = env.get_credential('FEC_CLASSIC_URL', 'http://classic.fec.gov')
 
+LIVE_VIDEO_URL = "https://fec.adobeconnect.com/openmeeting/"
```
xtine: These two lines need to be single quotes to match the style of the rest of the file.

fec/home/templates/home/meeting_page.html
```diff
@@ -41,20 +41,29 @@ <h1 class="heading__left">{% formatted_title self %}</h1>
...
       <p>The Commission meets regularly in executive sessions to discuss pending enforcement actions, litigation and other matters that, by law, must be kept confidential.</p>
     {% endif %}
-
-    {% if self.live_video_url %}
+    {% comment %}
+     Show streaming info if it is start_date or end_date of meeting AND between 7:30am and 5:00pm
+    {% endcomment %}
+    {% now 'F j, Y' as today %}
+    {% now 'H:i' as current_time %}
+    {% if self.date|date:'F j, Y' == today or self.end_date|date:'F j, Y' == today %}
```
xtine: So the more I look at this, the more I feel there's too much logic here going on in the template, and should be moved into a view. What do you think about this?

xtine: I also apologize if I was unclear in my answer to your question earlier. The logic here is candidate to put into a view, as I prefer view logic to template tag logic. So I would urge you to investigate a way to have a meeting page view, otherwise then your initial hunch for template tag inclusion was correct.

jenniferthibault:
@johnnyporkchops just checking back in on old PRs—what's the status of this one? Plan to complete it, or should we close it and make an issue for the work instead?

johnnyporkchops:
@ccostino, this has the model/database issue I asked about in slack. Advance apologies for the verbosity.
I want the live_video_url_field to actually render the default URL in the Wagtail editor for this page. 

--Originally, I was getting a deferred object in the field in Wagtail, like this:
`<django.db.models.query_utils.DeferredAttribute object at 0x104547470>` for both existing and newly created meeting pages. However, in the help-text, where I put it as a test, it renders the actual URL I want.

--Once I started from scratch, reloaded latest db dump and removed all my test migrations and just ran one migration, it seems to work. I think my history of migrations as I was trying this had added other defaults to existing meeting page data. So now, the field is blank for existing pages opened in the Wagtail editor, and showing the desired URL for newly created pages. This is fine.

Because of the original existence of the deferred object, I realized that adding a default value to existing page data may be problematic. For this particular usage, however, the live video URL has no significance for past meetings (existing meeting pages) so the value in the database is not important to any template filtering or searching on meeting page data. In fact, we were manually removing it after the meeting, but some may have remained. (this PR removes that manual process, among other things).  I just want to make sure I am not doing something that has other possible undesirable outcomes when it comes to the Wagtail generated data.

I realize that this question is a bit complicated, perhaps it would make more sense in paring. I have some more specific questions as well about using `null=True` and what that actually means.

fec/fec/constants.py
```diff
@@ -218,3 +218,8 @@
     '/home/about/',
     '/home/about/leadership-and-structure/'
 ]
+
+streaming_info = OrderedDict([
+    ("live-video-url", "https://fec.adobeconnect.com/openmeeting/"),
+    ("live-video-captions", "http://www.streamtext.net/player?event=BIS-FEC")
+])
```
ccostino: Out of curiosity, why the need for the `OrderedDict` here as opposed to a regular `dict`?  E.g.,

```python
{
    'live-video-url': 'https://fec.adobeconnect.com/openmeeting/',
    'live-video-captions': 'http://www.streamtext.net/player?event=BIS-FEC',
}

File: fec/fec/settings/base.py (hunk 3 above)
Comment (ccostino) [reply fec/fec/settings/base.py]: Good code style reminder here. :-)

File: fec/home/models.py [hunk 6]
```diff
@@ -982,8 +988,10 @@ class MeetingPage(Page):
     approved_minutes_link = models.URLField(blank=True)
     sunshine_act_links = models.TextField(
         blank=True, help_text='URLs separated by a newline')
-    live_video_url = models.URLField(blank=True)
-    live_video_captions = models.URLField(blank=True)
+    #Testing in below line using both settings and constants to set the url for botht the field value and help-text, in production, I would use one or the other
+    live_video_url = models.URLField(default=live_video_url, help_text=settings.LIVE_VIDEO_URL, null=True, blank=True)
```
ccostino: You should be okay here as far as I can tell.  [According to the Django documentation on the `default` field option](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.Field.default), the value cannot be a mutable object type, but if you desire to do that (and in this case you are), you just return it as a callable instead.

That said though, you are also referencing a particular key in the dictionary, so one other thing you could try is setting the default directly to the value returned by the key itself.  In this case, that would be this:

```python
default=constants.streaming_info['live-video-captions']
```

This will result in another migration to account for the adjustment, but if all goes well in creating it and loading things up, that could be another option.

Also note, according to the docs:  `The default value is used when new model instances are created and a value isn’t provided for the field.`

Meaning that it won't impact any existing records already created.  If you want to adjust those, that would require a data migration, but it sounds like that's not necessary.

fec/home/templates/home/meeting_page.html
```diff
@@ -41,20 +41,29 @@ <h1 class="heading__left">{% formatted_title self %}</h1>
...
+    {% if current_time > '07:30' and current_time < '17:00' and self.show_streaming_info %}
     <div class="row">
       <div class="slab slab--neutral u-padding--left usa-width-five-sixths">
         <div class="usa-width-one-fourth u-padding--left">
-          <a class="u-no-border" href="{{ self.live_video_url }}"><img src={% static 'img/thumbnail--video.png' %} alt="Video"></a>
+          {% comment %}
+           If there is live video or captions URLs provided in the template, use those. otherwise use default settings URLs
+          {% endcomment %}
+          <a class="u-no-border" href="{% if self.live_video_url %}{{ self.live_video_url }}{% else %}{{ settings.LIVE_VIDEO_URL }}{% endif %}"><img src={% static 'img/thumbnail--video.png' %} alt="video play button icon"></a>
```
ccostino: Same note as @xtine's remark above, please think about ways to move this into the view logic instead to simplify the template.  Please reach out if you need help after you've given it a shot! :-)

johnnyporkchops: Thanks @ccostino! I looked at moving  this to a templatetag which did not seem to make sense one I started developing it. I have not tried adding a view for it yet. I will look into that.

fec/fec/constants.py
johnnyporkchops: I copied the style here in constants.py. What is the significance of `OrderDict` here?:
```
update_types = OrderedDict([
    ("press-release", "Press release"),
    ("fec-record", "FEC Record"),
    ("weekly-digest", "Weekly Digest"),
    ("tips-for-treasurers", "Tips for Treasurers"),
    ("commission-meeting", "Commission Meetings"),
])
```

ccostino: Ah, I would assume that might be because those values are intended to show up in a certain order somewhere, but I just took a quick look through the code and that doesn't appear to be the case, at least not anymore.

Either way, that would be the use case for it - wanting to preserve order of key/value pairs for a dictionary-type structure.  Unless you also intend to do that with the streaming video URLs, there's no need for it and you can just use a standard dictionary. :-)

johnnyporkchops:
I do not want to close this PR yet--Even though we are going a different direction with the streaming media (as reflected in the latest meeting-page PR #2166 ), we may still use the logic developed here to control the appearance or lack thereof of the live stream on meeting pages.

johnnyporkchops:
Closing this and saving the code in case we need it in the the future

johnnyporkchops:
Closing this and saving code in case we need it in the future

PR #1337, opened 2017-10-11
[MERGE WITH GITFLOW] Release/public 20171011

johnnyporkchops:
After this release is deployed we need to:

- [ ] Merge and deploy this proxy PR: https://github.com/18F/fec-proxy/pull/60
- [ ] Double check that www.fec.gov/data is functioning after proxy deploy
- [ ] Stop web app instance in prod. Check to make sure www.fec.gov/data is still functioning. Make sure this route 404s: https://fec-prod-web.app.cloud.gov.
- [ ] When we are sure it is not needed anymore, we can delete that app entirely.

@LindsayYoung Can you double check that I didn't miss anything here?


ccostino:
We're nearly finished!  We're going to keep the `web` app in place for now (but it's stopped) until we're sure we don't need.  Likely in another couple of days or so we'll remove it.

PR #1338, opened 2017-10-11
Update Moment library and ignore locales

xtine:
The Moment npm library we are using to parse dates in JS by default will include all the timezone locales which takes a significant amount of space in our `vendor.js` file. We were also relying on an older version which had a security vulnerability, so along with updating moment I have included a line in Webpack to ignore all locales in Moment which results in a file savings of ~1mb (4.1mb -> 3.2mb)

I've also done a profiling of the home page to reflect performance gains to show before and after.
Before:
screen shot 2017-10-11 at 11 29 54 am
Average load times of 1.2s
screen shot 2017-10-11 at 12 57 42 pm
Average load times <1s

~10ms saved on script loading 

Bundle analyzer graphically showing how much moment locales were taking `vendor.js` filesize.
screen shot 2017-10-11 at 11 36 47 am
screen shot 2017-10-11 at 12 14 55 pm

### DIFF
--- a/fec/webpack.config.js
+++ b/fec/webpack.config.js
@@ -65,6 +65,7 @@ module.exports = [
         fileName: 'rev-manifest-js.json',
         basePath: '/static/js/'
       }),
+      new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/),
       // Uncomment to compress and analyze the size of bundles
       // new webpack.optimize.UglifyJsPlugin(),
       // new BundleAnalyzerPlugin()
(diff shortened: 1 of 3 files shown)


patphongs:
Thanks @xtine. Could you please make this PR to the 10/11 release branch since we want this in before release?

xtine:
@patphongs: it already is.

patphongs:
Oops, you're right @xtine! I got confused since I was looking at the mobile view of GitHub. Will review soon

PR #1350, opened 2017-10-17
Changes urls to use HTTPS when possible.

mbrickn:
Howdy!

I just updated some URL's to use https instead of http when possible. This helps users avoid a redirect and improves user experience.

Thank you for your time!


mbrickn:
Hi! Sorry I got sidetracked and forgot about this. Thank you for your help!

PR #1364, opened 2017-10-23
Fix webpack production issue

xtine:
It seemed that setting the production node_env variable inline was not ideal as it broke AO datatables. To fix this, we are setting the value from within the webpack config itself. While it seems counterintuitive, for now it works because setting the node_env variable to production allows for webpack to compress `vendor.js` and because we don't use the variable in any of our JS files currently, it is ok for local development.

I have also used the more preferred compression flag `webpack -p`, which automatically runs UglifyJS for the build-production command.

Fixes: https://github.com/18F/fec-cms/issues/1363

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "build-icons": "cd fec && gulp minify-icons && gulp consolidate-icons",
     "build-sass": "cd fec && gulp build-sass",
     "build-js": "cd fec && webpack",
-    "build-production": "cd fec && webpack --optimize-minimize --define process.env.NODE_ENV='production' && gulp build-sass",
+    "build-production": "cd fec && webpack -p && gulp build-sass",
     "test": "karma start",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
     "build-icons": "cd fec && gulp minify-icons && gulp consolidate-icons",
     "build-sass": "cd fec && gulp build-sass",
     "build-js": "cd fec && webpack",
-    "build-production": "cd fec && webpack --optimize-minimize --define process.env.NODE_ENV='production' && gulp build-sass",
+    "build-production": "cd fec && webpack -p && gulp build-sass",
```
patphongs: Question: Is the `-p` setting utilizing the UglifyJS Webpack Plugin? Is this why we no longer need to run `--optimize-minimize`?

xtine: https://webpack.js.org/guides/production/#cli-alternatives

PR #1380, opened 2017-10-25
Table panel fix

patphongs:
Resolves https://github.com/18F/fec-cms/issues/1377

Removes `toString` as it is unnecessary. Also adds ability to get cycle from the select box rather than the URL string. This was breaking the cycle in the financial totals when table panel is expanded.

Moment was being formatted incorrectly and tests were not passing. By using the correct usage of date formats, this allows tests to pass during builds https://momentjs.com/docs/#/displaying/

### DIFF
--- a/fec/fec/static/js/modules/table-panels.js
+++ b/fec/fec/static/js/modules/table-panels.js
@@ -59,7 +59,7 @@ var renderCandidatePanel = function(showFinancialTotals) {
       var query = URI.parseQuery(window.location.search);
       // Parse all of the time-related variables
       var electionYear = query.election_year;
-      var cycle = query.cycle || query.election_year;
+      var cycle = query.cycle || query.election_year || $('#cycle').val();
       var electionFull = query.election_full === 'true' ? true : false;
 
       // Build a string showing the range covered by the financial totals
--- a/fec/fec/tests/js/top-entities-breakdown.js
+++ b/fec/fec/tests/js/top-entities-breakdown.js
@@ -284,7 +284,8 @@ describe('Top entities breakdown', function() {
       it('updates the coverage dates for the current year', function() {
         var today = new Date();
         var lastYear = today.getFullYear() - 1;
-        var formattedToday = moment(today, 'DD/MM/YYYY');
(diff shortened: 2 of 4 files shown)


fec/fec/static/js/modules/table-panels.js
```diff
@@ -59,7 +59,7 @@ var renderCandidatePanel = function(showFinancialTotals) {
       var query = URI.parseQuery(window.location.search);
       // Parse all of the time-related variables
       var electionYear = query.election_year;
-      var cycle = query.cycle || query.election_year;
+      var cycle = query.cycle || query.election_year || $('#cycle').val();
```
xtine: Why is there a triple comparison going on?

patphongs: Currently I am not sure everywhere that table-panels.js is being used and the #cycle select box may not be available on every instance. Therefore I opted to keep the existing logic and tack on the #cycle check.

xtine: So in this case I think it would seem be a bit confusing where cycle is actually coming from. Do you think you could figure out a way to explicitly find the value of cycle?

fec/fec/tests/js/top-entities-breakdown.js
```diff
@@ -284,7 +284,7 @@ describe('Top entities breakdown', function() {
       it('updates the coverage dates for the current year', function() {
         var today = new Date();
         var lastYear = today.getFullYear() - 1;
-        var formattedToday = moment(today, 'DD/MM/YYYY');
+        var formattedToday = moment(today).format('MM/DD/YYYY');
```
xtine: Nice catch on these.

fec/fec/static/js/modules/table-panels.js
patphongs: So looking around a bit, noticed that the House candidates datatable page is not pulling the query parameters into the URL string like the senate and president is. I imagine that the cycle and election_year are important for those other offices because they are running on 4 and 6 year cycles. Need to figure out how to add cycle to the house query string.

PR #1432, opened 2017-11-01
[MERGE WITH GITFLOW] localize timezone-aware date and handle DST(daylight savings time)

johnnyporkchops:
This is another attempt at solving timezone error related to timezone unaware date objects in the banner announcements code. Tested locally and it does not give error, but still need to see if other tests/monitors report errors/warnings. i.e.- Kibana

### DIFF
--- a/fec/home/templatetags/home_page.py
+++ b/fec/home/templatetags/home_page.py
@@ -1,6 +1,6 @@
 import datetime
 import pytz
-from pytz import timezone
+
 import re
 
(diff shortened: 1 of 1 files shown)


patphongs:
Tested this in dev and did not see any banner timezone errors in the logs after implementation. Banner also went up and expired as expected in dev. Pushing this to stage now so that we can get this out in the release and also do a last check in stage.

PR #1433, opened 2017-11-01
update breadcrumb-links for mur

pkfec:
removed data from breadcrumb_links
added missing breadcrumb_links for current murs


pkfec:
thank you!

PR #1440, opened 2017-11-02
[MERGE WITH GIT FLOW] Hotfix/fix redirects

lbeaufort:
Addresses https://github.com/18F/openFEC/issues/2730

Fix broken redirects by importing redirect from django.shortcuts package


pkfec:
tested in my local. looks good.

PR #1449, opened 2017-11-04
Display political party on committee profile page

pkfec:


pkfec:
- fetch party_type variable from committee api.
- add party_type variable  to template_variables for committee.

@xtine and @fec-jli please review this change.
cc @PaulClark2 @jwchumley @hcaofec

PR #1450, opened 2017-11-04
[MERGE WITH GITFLOW] Hotfix-wagtail html editing 

johnnyporkchops:
Issue : https://github.com/18F/fec-cms/issues/1415
Restores  ace-builds noconflict js files to allow html editing. Tested locally on Master
- This is a node module that is not being included:  fec-cms/node-modules/ace-builds
- In this hotfix, the  files are just included in /static/js/vendor which might not be correct, but works to restore editing until we can fix the build to include it


xtine:
This has been fixed by https://github.com/18F/fec-cms/pull/1478.

PR #1452, opened 2017-11-06
Fix/tooltip under alignment

xtine:
This fix establishes override properties for the calendar detail tooltips so they are properly aligned on month-at-a-glance view. I also restored the top tooltip arrows so they show up now!

screen shot 2017-11-06 at 10 55 47 pm
screen shot 2017-11-06 at 10 56 10 pm
screen shot 2017-11-06 at 10 55 40 pm

https://github.com/18F/fec-cms/issues/1274

### DIFF
--- a/fec/fec/static/img/tooltip-point.svg
+++ b/fec/fec/static/img/tooltip-point.svg
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg width="20px" height="10px" viewBox="0 0 20 10" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 3.8.3 (29802) - http://www.bohemiancoding.com/sketch -->
+    <title>tooltip-point</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="tooltip-point">
+            <polygon id="Path-2" fill="#112E51" points="0 10 2.00851609 10 10.00819 10 17.9925609 10 20 10 10.00819 0"></polygon>
+            <polygon id="Triangle-2" fill="#FFFFFF" points="10 2.5 17.5 10 2.5 10"></polygon>
+        </g>
+    </g>
+</svg>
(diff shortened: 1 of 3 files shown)


jenniferthibault:
@xtine in the second screenshot, we seem to have lost the pointy tip of the tooltip altogether, and in the first, it's in a reaaaaaly weird place inside the container. 

Here's an example of the shape from design files, though I if my memory is right, I think the point might have been adjusting it's location to what it was pointing at...can't remember.

screen shot 2017-11-06 at 7 49 16 pm

Is this something we could address in this pass? I'm concerned that having the point inside the calendar tooltip looks unexpected and broken.

xtine:
@jenniferthibault: wow, okay I didn't realize that arrow was wrong (just an interesting design 😆) because I've never worked with the calendar tooltip before. Can you post what it should look like in here?

And yeah some time even before merge the tooltip point was lost. I can definitely work on getting that back. I assume the point should just be below the (i) icon in the filter label?

jenniferthibault:
My memory isn't total garbage! It did used to move! 

https://github.com/18F/fec-style/pull/474

I'm trying to hunt down what the design is for the tooltip in the calendar grid view, because the overlapping close X and the event title are off too. Hold plz while I dig some old issues up.

jenniferthibault:
@xtine via https://github.com/18F/FEC/issues/561 I found an old image of what the grid view tooltip is supposed to look like:

screen shot 2017-11-06 at 8 22 43 pm


You're right about the point—it should be directly centered under the thing that initiates it opening, which is either the circular information icon, or the calendar event

jenniferthibault:
And here are additional Calendar tooltip PRs that I found while looking, in case they're helpful:
- https://github.com/18F/fec-style/pull/445
- https://github.com/18F/fec-style/pull/546
- https://github.com/18F/fec-style/pull/402
- https://github.com/18F/fec-style/pull/474

xtine:
@jenniferthibault: thanks for digging up the old PRs and calendar detail design, I found out that the override properties for calendar detail tooltip weren't showing up so I added specificity to match the designs.
Also looking at the stylesheets I noticed that the old tooltip svgs were missing from the latest fec-style so I restored those as well.

xtine:
@jenniferthibault: I think we should make a non-dependent task. We need to migrate pattern library to circle and I still need to make a style asset migration script so those need to be done first of course. If those weren't issues I would be up for a documentation step but we are a bit behind already.

jenniferthibault:
Sounds good, I just made that a separate, non-dependent issue in the pattern library for fixing the busted tooltips based on the work that happened in this PR 👍 

https://github.com/18F/fec-pattern-library/issues/20

PR #1471, opened 2017-11-08
Make senate special cycles in election dropdowns data-driven

lbeaufort:
## Summary
Senate special elections can occur in years when there isn't usually an election in that state. Currently we're hard-coding states with specials in 2018, which isn't very sustainable for future cycles and states. This PR adds to `api_caller.py` so that senate specials on the data/elections pages are data-driven. 

- Addresses # [2621](https://github.com/18F/openFEC/issues/2621)
Change the logic so only the senate special cycles are added to the dropdown and not count backwards by 6 for that year. This will automatically add senate election years as new special elections are added to the `trc_election` tables. Made tests for new functionality. 

This also addressed a bug I found where extra elections were being added to AL due to senate special logic:

## Before
image

## After
image

www.fec.gov/data/elections/office/state/cycle
www.fec.gov/data/elections/office/cycle
Example: http://fec.gov/data/elections/senate/AL/2018/
`views.py` -> `elections()`
`api_caller.py`

### DIFF
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -311,3 +311,67 @@ def _get_sorted_documents(ao):
     sorted_documents = sorted(ao['documents'], key=itemgetter('description', 'document_id'), reverse=False)
     sorted_documents = sorted(sorted_documents, key=itemgetter('date'), reverse=True)
     return sorted_documents
+
+
+def call_senate_specials(state):
+    """ Call the API to get Senate special election information for
+        given state. Returns a list of dictionaries
+        Example: [{details for election1}][{details for election2}]
+    """
+    api_response = _call_api('election-dates',
+                             election_type_id='SG',
+                             office_sought='S',
+                             election_state=state)
+
+    special_results = api_response['results']
+
+    return special_results if 'results' in api_response else None
+
+
+def format_special_results(special_results):
+    """ Takes special_results, which is a list of dictionaries,
+        returns a list of election years. Round odd years up to even.
+        Example: [2008, 2000]
+    """
+    senate_specials = []
+
+    for result in special_results:
+
+        # Round odd years up to even years
+        result['election_year'] = result['election_year'] + (result['election_year'] % 2)
+
+        senate_specials.append(result['election_year'])
+
+    return senate_specials
+
+
+def get_regular_senate_cycles(state):
+    """ Get the list of election cycles based off Senate class
+    """
+    senate_cycles = []
+
+    for senate_class in ['1', '2', '3']:
+        if state.upper() in constants.SENATE_CLASSES[str(senate_class)]:
+            senate_cycles += utils.get_senate_cycles(senate_class)
+
+    return senate_cycles
+
+
+def get_all_senate_cycles(state):
+    """  Add together regularly scheduled and special senate elections
+        Return a list of election years sorted in descending order
+    """
+    senate_specials = format_special_results(call_senate_specials(state))
+    senate_regular_cycles = get_regular_senate_cycles(state)
+
+    all_senate_cycles = senate_regular_cycles
+
+    for special_year in senate_specials:
+        # Prevent duplicates
+        if special_year not in all_senate_cycles:
+            all_senate_cycles.append(special_year)
+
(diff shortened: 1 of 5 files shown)


fec/data/api_caller.py
```diff
@@ -311,3 +311,67 @@ def _get_sorted_documents(ao):
...
+        Example: [{details for election1}][{details for election2}]
+    """
+    api_response = _call_api('election-dates',
+                             election_type_id='SG',
+                             office_sought='S',
+                             election_state=state)
+
+    special_results = api_response['results']
+
+    return special_results if 'results' in api_response else None
```
ccostino: This is nice and straight forward, but there is also room for some improvement in lines 326 - 328:

* When accessing a key in a dictionary, you want to be mindful of what happens if the key doesn't exist.  In this situation, if there is no `results` key in the `api_response` dictionary that could result in a `KeyError` being thrown, therefore resulting in a potential 500 server error response.  A safeguard against this is to use the [`get()`](https://docs.python.org/3.5/library/stdtypes.html#dict.get) method on [dictionary objects](https://docs.python.org/3.5/library/stdtypes.html#dict) and supply a default value in the case of a key not being found.
* Compound statements like this `return` statement with the `if` clause in it are okay, but can be hard to parse and maintain in the future when going back to read code.  It's more of a style call than anything, but just something to be mindful of while writing code to be as helpful to others as possible, including your future self! :-)

Given the above points, I might have written this code like this instead:

```python
return api_response.get('results', None)
```

fec/data/api_caller.py
```diff
@@ -311,3 +311,67 @@ def _get_sorted_documents(ao):
...
+        Example: [2008, 2000]
+    """
+    senate_specials = []
+
+    for result in special_results:
+
+        # Round odd years up to even years
+        result['election_year'] = result['election_year'] + (result['election_year'] % 2)
+
+        senate_specials.append(result['election_year'])
```
ccostino: As noted above, just be careful when accessing dictionary keys and error handling!

fec/data/api_caller.py
```diff
@@ -311,3 +311,67 @@ def _get_sorted_documents(ao):
...
+
+    return special_results if 'results' in api_response else None
+
+
+def format_special_results(special_results):
+    """ Takes special_results, which is a list of dictionaries,
+        returns a list of election years. Round odd years up to even.
+        Example: [2008, 2000]
+    """
+    senate_specials = []
```
ccostino: This is totally fine; you could also consider [setting a default value](https://docs.python.org/3.5/tutorial/controlflow.html#default-argument-values) to `[]` in the method definition for the `special_results` argument, e.g.,:

```python
def format_special_results(special_results=[])
```

fec/data/api_caller.py
```diff
@@ -311,3 +311,67 @@ def _get_sorted_documents(ao):
...
+    return senate_specials
+
+
+def get_regular_senate_cycles(state):
+    """ Get the list of election cycles based off Senate class
+    """
+    senate_cycles = []
+
+    for senate_class in ['1', '2', '3']:
+        if state.upper() in constants.SENATE_CLASSES[str(senate_class)]:
```
ccostino: There shouldn't be the need for the explicit call to `str()` here around the `senate_class` value, since it would already be a `string` type.  Was there an issue without it?

Edit:  Ah, I see it was pulled from elsewhere, and sometimes that's a good opportunity to do a bit of refactoring/cleanup as well. :-)

fec/data/tests/test_utils.py
```diff
@@ -71,9 +72,32 @@ def test_get_senate_cycles(self):
...
+            'election_notes': "Sen. Robert Byrd's Seat.",
+            'primary_general_date': '2016-09-16T16:09:22.555513'}]
+
+        # According to our mock call, WV has a special in 2010 and not 2018
+        westvirginia = api_caller.format_special_results(
+            call_senate_specials_mock_wv)
+
+        assert 2010 in westvirginia
+        assert 2018 not in westvirginia
+
```
ccostino: This is great, excellent test!

PR #1479, opened 2017-11-09
refactor python tests in fec-cms repo

pkfec:
Added and refactored test_issues.py django test case
Added and refactored test_legal_search.py django test case
Refactored dajngo test case

## Summary (required)

- Addresses # [_Issue number_]

-  

## Screenshots

- _Include a screenshot of the new/updated features in context (“in the wild”). If it is an interface or front end change, include both a Before and After screenshot._

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

____


pkfec:
@LindsayYoung and @vrajmohan  Please review this PR.

In test_legal_search.py, out of 8 tests,  below two are failing. Please take a look and let me know if i missed anything.

1. test_search_type_regulations_redirects
2. test_search_regulations

LindsayYoung:
test_search_type_regulations_redirects is failing because you found a real regression. Yay, for testing! That issue has been logged here https://github.com/18F/fec-cms/issues/1477 

We can remove the redirect test for now and we can add it back when the redirect is working again.

@vrajmohan I am not sure what is supposed to be going on in the offset for the search test, maybe you could provide more context.

Thanks @pkfec!

vrajmohan:
Looks like `test_search_regulations` was not correctly migrated. In https://github.com/18F/openFEC-web-app/blob/develop/tests/unit/test_legal_search.py#L51, the URL includes `/regulations`.

fec/data/tests/test_legal_search.py
```diff
@@ -0,0 +1,132 @@
...
+                              data={
+                                  'search': 'in kind donation',
+                                  'search_type': 'regulations'})
+        print("::: Test4 ::: search_regulations ::: RESPONSE :::", response.status_code)
+        assert response.status_code == 200
+        load_legal_search_results.assert_called_once_with('in kind donation',
+                                                          'regulations', offset='0')
+
+    # Test5 : OK. This test works only if offset value is set
+    # as a string and not integer.
```
vrajmohan: The test should work with integers. The root cause is that we are missing the integer conversion from https://github.com/18F/openFEC-web-app/blob/develop/openfecwebapp/routes.py#L550 in the migration.

vrajmohan:
We should remove all the `print` calls in the tests. They may help in understanding but we should remove them before committing the code.

pkfec:
print statements are just for testing purpose. i will remove once i have all the tests working.

fec/data/tests/test_legal_search.py
pkfec: all other tests are working fine without the integer conversion. i do not see why we have to explicitly do the conversion for test_search_pagination test.

vrajmohan: The fix is not in the test code but in the actual production code. As you can see in Test 6, the call should happen with an `int` argument.

pkfec:
@LindsayYoung Could you please review this PR?

PR #1480, opened 2017-11-09
Grid-item-w-image (for H4CC and Homepage getting started sections-PR 1441)

johnnyporkchops:
**This PR replaces https://github.com/18F/fec-cms/pull/1441. Past history /screenshots can be seen there.** (This also incorporates requested review-edits from closed PR https://github.com/18F/fec-cms/pull/1466)

This PR specifically solves icon stretching problems that happened in mobile sizes when placing icon images in the generic grid items
This also solves the problem of unacceptably large image/icon-heights, taking up too much screen area in mobile sizes. 

This PR uses the object-fit property, among others, to apply styles to images that are inside of  responsive grid items.
This includes three new grid styles that expand upon existing styles--following the intended convention of modularity and the use of existing building blocks, rather that creating new.

**Browser compatibility:**
Browser support for object-fit is very good except for Internet Explorer Edge/11. However it has been prioritized for inclusion by MDN as of 10/17/17, and there are some workarounds that have been implemented here to limit any degradation of experience for IE users.

Screenshots show testing in Internet Explorer. By including 100%-width SVGs , instead of JPGs, icon stretching is eliminated. The mobile experience degradation is at an acceptable level for IE users.

ie-11-screenshot

ie-11-screenshot_mobile

### DIFF
--- a/fec/fec/settings/dev.py
+++ b/fec/fec/settings/dev.py
@@ -10,7 +10,6 @@ for t in TEMPLATES:
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 
-ALLOWED_HOSTS = ['*']
 
 try:
     from .local import *  # noqa
--- a/fec/fec/settings/dev.py
+++ b/fec/fec/settings/dev.py
@@ -10,6 +10,8 @@ for t in TEMPLATES:
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 
+ALLOWED_HOSTS = ['*']
+
 try:
     from .local import *  # noqa
 except ImportError:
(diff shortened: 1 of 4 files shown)


johnnyporkchops:
This includes requested previously reviewed and closed PR https://github.com/18F/fec-cms/pull/1466

johnnyporkchops:
@jeanninehunter , @JonellaCulmer.  Is this image using our primary color? It seems like maybe it was done with a variation of that color for effect (?) I am asking because we are now using the svg and coloring the background with the primary color using CSS. If it is a different color, I can change the CSS to match that hex value.
img1-1x

JonellaCulmer:
According to the pattern library, that deep red color should be #961A16 and when I pulled the image it was #961B16

jenniferthibault:
Correct, it's our `$deep-red` #961A16 which is a little brighter than `$crimson`. https://fec-pattern-library.app.cloud.gov/docs/colors.html

@JonellaCulmer I'm guessing the slight variance you're picking up is from using a color picker over pixelation (unless you opened it as an SVG and ready the color value in settings?)

Issue #1434, opened 2017-11-01
Implement New 24- and 48-Hour Report Toggle Default


JonellaCulmer:
**What we're after:** For interested groups, the latest independent expenditures (IEs) are often more valuable than IEs listed in regularly scheduled reports. In order to get users to the most recent IEs made, we need to make the default search view of IEs submitted through `24- and 48-Hour Reports`.

**Design:**
![modifiedietoggle](https://user-images.githubusercontent.com/31663028/32291571-d44f5278-bf13-11e7-8b36-3fae211c18ad.jpg)

**Completion Criteria:**
- [x] Swap the `24- and 48-Hour Reports` toggle with the `Regularly scheduled reports` toggle in the filters panel so that `24- and 48-Hour Reports` is the top option
- [x] Set the `24- and 48-Hour Reports` as the default option, so data in that view is populated automatically on the page

History in #1375

jenniferthibault:
I got this started in the branch `feature/fix-ie-report-toggle`

A midstream question, and a good opportunity for someone from the FEC to train a novice...
on https://github.com/18F/fec-cms/blob/feature/fix-ie-report-toggle/fec/data/templates/macros/filters/ie-reports.jinja#L14 the value set for the 24- and 48- Hour Reports toggle is  already set to `value="true"` , which makes me think it's the toggle option that's applied by default, but that doesn't seem to be the case looking at the existing interface, which applies the "Regularly scheduled reports" option by default, even though it's set to `value="false"`

Is there another property or file that determines which option is applied by default that I'm overlooking?

jenniferthibault:
I got some help from a colleague at 18F, and finished up the work in https://github.com/18F/fec-cms/pull/1496 which is now ready for review

AmyKort:
Is this ready to close?

jenniferthibault:
Yes, closed via implementation #1496

PR #1496, opened 2017-11-15
Make 24/48 hr reports default filter selection for viewing independent expenditures

jenniferthibault:
## Summary

- Addresses #1434 
- Changes the 24/48 hour report toggle to be the default option selected on page load. To support this, it also moves the selected option into the first position.
- I also caught a small typo in the tooltip content while I was looking.

[Independent expenditures data table](https://www.fec.gov/data/independent-expenditures)

## Screenshots

If I did everything correctly, the before/after should look like this:

**Before:**
screen shot 2017-11-15 at 1 43 29 pm

**After:** 
screen shot 2017-11-15 at 1 42 28 pm

______________

❗️❗️  Please test this by running it locally (as you usually would)—I've had trouble with my local env so have not been able to test the functionality.

### DIFF
--- a/fec/data/templates/macros/filters/ie-reports.jinja
+++ b/fec/data/templates/macros/filters/ie-reports.jinja
@@ -10,13 +10,13 @@
       </div>
     </div>
   </div>
-  <label for="summary{{id_suffix}}" class="toggle">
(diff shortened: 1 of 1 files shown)


lbeaufort:
@xtine and @jenniferthibault  - I'm reviewing this now.

PR #1506, opened 2017-11-17
Get started homepage edits

patphongs:
## Summary

@jenniferthibault realized that we need to have h4 headings in the Get started section of the homepage. This adds this refinement.

-  Homepage

## Screenshots

screen shot 2017-11-17 at 10 11 44 am

### DIFF
--- a/fec/fec/static/scss/components/_type-styles.scss
+++ b/fec/fec/static/scss/components/_type-styles.scss
@@ -57,7 +57,7 @@
   font-style: italic;
   font-size: u(1.4rem);
 
-  &.normal {
+  &.t-normal {
     font-style: normal;
   }
 }
--- a/fec/fec/static/scss/components/_type-styles.scss
+++ b/fec/fec/static/scss/components/_type-styles.scss
@@ -56,6 +56,10 @@
   font-weight: normal;
   font-style: italic;
   font-size: u(1.4rem);
+
+  &.normal {
+    font-style: normal;
+  }
 }
 
 .t-block {
--- a/fec/fec/static/scss/components/_type-styles.scss
+++ b/fec/fec/static/scss/components/_type-styles.scss
@@ -56,10 +56,6 @@
   font-weight: normal;
   font-style: italic;
   font-size: u(1.4rem);
-
-  &.t-normal {
-    font-style: normal;
-  }
 }
 
 .t-block {
@@ -93,6 +89,7 @@
 
 .t-normal {
   font-weight: normal;
+  font-style: normal;
 }
 
 .t-highlight {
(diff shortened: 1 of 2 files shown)


fec/home/templates/home/home_page.html
```diff
@@ -60,17 +60,17 @@ <h2 class="heading--section u-margin--bottom">Get started</h2>
       <ul class="grid grid--3-wide grid--flex t-sans">
         <li class="grid__item grid__item--with-img grid__item--with-button">
           <img src="{% static 'img/feature--candidate-support.jpg' %}" alt="">
-          <h3 class="t-sans u-margin--top">Find out how individuals can support federal candidates</h3>
+          <h4>Find out how individuals can support federal candidates</h4>
           <div>Including making contributions, volunteering, internet activities and filing complaints</div>
```
jenniferthibault: Question for my own understanding: is it common to put a new div in just to add a paragraph? Or would a different approach be to use a `<p>` tag with classes? cc @xtine

xtine: There should be no blank divs.

patphongs: @xtine The div is populated with text, so it's not blank. But I can understand @jenniferthibault about wanting to make the HTML more semantic with a `<p>` tag rather than a div. I actually started out with making it a p tag, however, I ran into situations where the p tag was giving me a 1.6rem font size rather than the 1.4rem in the design comp. 
Using <p> tag in code:
screen shot 2017-11-17 at 1 34 19 pm

xtine: When I say blank div, I mean a div without a class.

patphongs:
@xtine I just tried removing the div around the subtext, but it caused misalignment in the buttons (see below). So I think it needs some kind of container around it. In a recent commit, I put in a `<p>` tag but then also added an additional t-note class and removed the italics within the t-note with a normal class. Do you think this would be a good approach?

screen shot 2017-11-20 at 10 13 17 am

fec/fec/static/scss/components/_type-styles.scss
```diff
@@ -56,6 +56,10 @@
   font-weight: normal;
   font-style: italic;
   font-size: u(1.4rem);
+
+  &.normal {
```
xtine: Can you use `.t-normal` here instead?

patphongs:
@xtine I just changed the class name to `t-normal` and also added the `t-sans` class to the paragraph to apply the sans typography as specified in the design comp: https://user-images.githubusercontent.com/11636908/30135715-01bf5e30-932a-11e7-9a2b-fd0e1804e1fe.png

fec/fec/static/scss/components/_type-styles.scss
```diff
@@ -56,6 +56,10 @@
   font-weight: normal;
   font-style: italic;
   font-size: u(1.4rem);
+
+  &.t-normal {
```
xtine: What I mean, is that there is already a `t-normal` at line 94. Can we not use that one instead of making this specific class?

patphongs:
@xtine Thanks, I didn't realize there was an existing t-normal class. I have removed the duplicate I created and used that class instead.

Issue #1461, opened 2017-11-07
Restore redirect for most recent cycle


patphongs:
This PR https://github.com/18F/fec-cms/pull/1448 removed the redirect functionality to redirect you to the most recent cycle with filing information. 

`url_for` function is using a flask syntax. We need to convert this to using a django redirect function.

patphongs:
PR https://github.com/18F/fec-cms/pull/1528 resolves this issue. Closing...

PR #1528, opened 2017-11-29
Fixed redirect to cycle with data

patphongs:
## Summary

- Addresses #1461 
- Previous implementation was using a `url_for` function to construct the redirect, which is using a flask function. To fix this, we needed to convert the redirect to use a django url constructor. Django's `reverse()` was used in the PR.

-  Data section of the site. Examples of url that was previously effected: http://localhost:8000/data/committee/C90011677/. The redirect now redirects to the most recent cycle with data.

### DIFF
--- a/fec/data/urls.py
+++ b/fec/data/urls.py
@@ -8,7 +8,7 @@ urlpatterns = [
     url(r'^data/search/$', views.search),
     url(r'^data/advanced/$', views.advanced),
     url(r'^data/candidate/(?P<candidate_id>\w+)/$', views.candidate),
-    url(r'^data/committee/(?P<committee_id>\w+)/$', views.committee),
+    url(r'^data/committee/(?P<committee_id>\w+)/$', views.committee, name='committee-by-id'),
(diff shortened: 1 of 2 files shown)


lbeaufort:
@patphongs I've tested locally but not exhaustively.

fec/data/urls.py
```diff
@@ -8,7 +8,7 @@
     url(r'^data/search/$', views.search),
     url(r'^data/advanced/$', views.advanced),
     url(r'^data/candidate/(?P<candidate_id>\w+)/$', views.candidate),
-    url(r'^data/committee/(?P<committee_id>\w+)/$', views.committee),
+    url(r'^data/committee/(?P<committee_id>\w+)/$', views.committee, name='committee-by-id'),
```
ccostino: Note that with this extra `name` argument any template making use of this URL can reference it via the name now, too!  Here's [an example in the Django documentation](https://docs.djangoproject.com/en/1.10/topics/http/urls/#examples) in case that helps. :-)

patphongs: Thanks @ccostino ! It would be nice to add a name for all of them, I think it just helps to understand semantically what each pattern does.

PR #1529, opened 2017-11-29
Feature/senate next election year

fec-jli:
## Summary (required)

- Addresses # [_Issue number_]

1) modify function **def get_senate_cycles** in utils.py to get next election year from a formula instead of from constants.py

2)remove the NEXT_SENATE_ELECTIONS from constants.py

3)add test case into test_utils.py 
def test_get_senate_cycles

## Related issue on openFEC repo #2692:
[https://github.com/18F/openFEC/issues/2692
](url)

### DIFF
--- a/fec/data/tests/test_utils.py
+++ b/fec/data/tests/test_utils.py
@@ -69,7 +69,11 @@ class TestCycles(unittest.TestCase):
         assert utils.get_cycles(2020) == range(2020, 1979, -2)
 
     def test_get_senate_cycles(self):
-        assert utils.get_senate_cycles(1) == range(2018, 1979, -6)
+        current_cycle = 2018
+        assert utils.get_senate_cycles('1') == range(current_cycle, 1979, -6)
+        assert utils.get_senate_cycles('2') == range(current_cycle + 2, 1979, -6)
+        assert utils.get_senate_cycles('3') == range(current_cycle + 4, 1979, -6)
+
 
     def test_state_senate_cycles(self):
 
--- a/fec/data/tests/test_utils.py
+++ b/fec/data/tests/test_utils.py
@@ -70,9 +70,28 @@ class TestCycles(unittest.TestCase):
 
     def test_get_senate_cycles(self):
         current_cycle = 2018
-        assert utils.get_senate_cycles('1') == range(current_cycle, 1979, -6)
-        assert utils.get_senate_cycles('2') == range(current_cycle + 2, 1979, -6)
-        assert utils.get_senate_cycles('3') == range(current_cycle + 4, 1979, -6)
+        assert utils.get_senate_cycles('1', current_cycle) == range(
+            2018, 1979, -6)
+        assert utils.get_senate_cycles('2', current_cycle) == range(
+            2020, 1979, -6)
+        assert utils.get_senate_cycles('3', current_cycle) == range(
+            2022, 1979, -6)
+
+        current_cycle = 2020
+        assert utils.get_senate_cycles('1', current_cycle) == range(
+            2024, 1979, -6)
+        assert utils.get_senate_cycles('2', current_cycle) == range(
+            2020, 1979, -6)
+        assert utils.get_senate_cycles('3', current_cycle) == range(
+            2022, 1979, -6)
+
+        current_cycle = 2022
+        assert utils.get_senate_cycles('1', current_cycle) == range(
+            2024, 1979, -6)
+        assert utils.get_senate_cycles('2', current_cycle) == range(
+            2026, 1979, -6)
+        assert utils.get_senate_cycles('3', current_cycle) == range(
+            2022, 1979, -6)
 
 
     def test_state_senate_cycles(self):
--- a/fec/data/utils.py
+++ b/fec/data/utils.py
@@ -215,22 +215,24 @@ def get_next_senate_elections(current_cycle):
     For example, class 1 will always be up for election in years
     with year % 6 = 2, for example 2018.
     """
-    next_election_cycles = {
+    next_election_decoder = {
         2: {'1': current_cycle, '2': (current_cycle + 2), '3': (current_cycle + 4)},
         4: {'2': current_cycle, '3': (current_cycle + 2), '1': (current_cycle + 4)},
         0: {'3': current_cycle, '1': (current_cycle + 2), '2': (current_cycle + 4)}
     }
-    return next_election_cycles.get(current_cycle % 6)
+    return next_election_decoder.get(current_cycle % 6)
 
 
-def get_senate_cycles(senate_class, current_cycle=current_cycle()):
+def get_senate_cycles(senate_class, cycle=None):
     """
     Returns an list of elections based on senate class
     Uses get_next_senate_elections to find out
     which classes are up for election
     Adds years to current cycle depending on what position the class is in
     """
-    senate_classes = get_next_senate_elections(current_cycle)
+    if cycle is None:
+        cycle = current_cycle()
+    senate_classes = get_next_senate_elections(cycle)
     return range(senate_classes.get(senate_class), constants.START_YEAR, -6)
 
 
--- a/fec/data/utils.py
+++ b/fec/data/utils.py
@@ -13,7 +13,6 @@ def current_cycle():
     year = datetime.datetime.now().year
     return year + year % 2
 
-
 class LRUCache(cachecontrol.cache.BaseCache):
     """A thread-safe least recently updated cache adapted to work with
     Cache-Control.
@@ -206,13 +205,43 @@ def process_ie_data(totals):
     return financial_summary_processor(totals, constants.IE_FORMATTER)
 
 
+def get_current_year_class():
+    """Get Class number for current year
+    """
+    if (current_cycle() - constants.END_YEAR) % 6 == 0:
+        current_year_class = '1'
+    elif (current_cycle() - constants.END_YEAR + 4) % 6 == 0:
+        current_year_class = '2'
+    elif (current_cycle() - constants.END_YEAR + 2) % 6 == 0:
+        current_year_class = '3'
+    return current_year_class
+
+
 def get_senate_cycles(senate_class):
-    if senate_class in '1':
-        next_election = current_cycle()
-    elif senate_class in '2':
-        next_election = current_cycle() + 2
-    elif senate_class in '3':
-        next_election = current_cycle() + 4
+
+    if get_current_year_class() in '1':
+        if senate_class in '1':
+            next_election = current_cycle()
+        elif senate_class in '2':
+            next_election = current_cycle() + 2
+        elif senate_class in '3':
+            next_election = current_cycle() + 4
+
+    elif get_current_year_class() in '2':
+        if senate_class in '1':
+            next_election = current_cycle() + 4
+        elif senate_class in '2':
+            next_election = current_cycle()
+        elif senate_class in '3':
+            next_election = current_cycle() + 2
+
+    elif get_current_year_class() in '3':
+        if senate_class in '1':
+            next_election = current_cycle() + 2
+        elif senate_class in '2':
+            next_election = current_cycle() + 4
+        elif senate_class in '3':
+            next_election = current_cycle()
 
     return range(next_election, constants.START_YEAR, -6)
 
--- a/fec/data/utils.py
+++ b/fec/data/utils.py
@@ -205,45 +205,33 @@ def process_ie_data(totals):
     return financial_summary_processor(totals, constants.IE_FORMATTER)
 
 
-def get_current_year_class():
-    """Get Class number for current year
+def get_next_senate_elections(current_cycle):
     """
-    if (current_cycle() - constants.END_YEAR) % 6 == 0:
-        current_year_class = '1'
-    elif (current_cycle() - constants.END_YEAR + 4) % 6 == 0:
-        current_year_class = '2'
-    elif (current_cycle() - constants.END_YEAR + 2) % 6 == 0:
-        current_year_class = '3'
-    return current_year_class
+    Returns an dictionary of senate classes in chronological order
+    based on current cycle year
+    Uses mod 6 to determine what senate classes are up for election
+    in the current cycle
+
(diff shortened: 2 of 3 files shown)


fec/data/utils.py
```diff
@@ -207,7 +207,13 @@ def process_ie_data(totals):
...
 
 def get_senate_cycles(senate_class):
-    next_election = constants.NEXT_SENATE_ELECTIONS[str(senate_class)]
+    if senate_class in '1':
+        next_election = current_cycle()
+    elif senate_class in '2':
+        next_election = current_cycle() + 2
+    elif senate_class in '3':
+        next_election = current_cycle() + 4
+
```
ccostino: Would we ever have a situation where there are more than 3 variations of senate classes to calculate off of?  With this setup, we'd have to continue modifying this `if` block, and now we're dealing with both the code itself and the data being processed, instead of just the data as we had before.  Furthermore, we probably want this to be a calculation we can leverage instead, so we never have to modify anything in the future around this. :-)

I did some searching and this `get_senate_cycles` method is only called in the [`api_caller.py`](https://github.com/18F/fec-cms/blob/develop/fec/data/api_caller.py#L348-L357) module.  What if we did a bit more cleanup overall here:

* Keep the `NEXT_SENATE_ELECTIONS` constant instead of removing it, but change it to contain the number of years to add to the current cycle depending on the `senate_class` for the dictionary values instead of set years like it currently has.  Keep the key names as they are (strings), because there is a loose dependency here in mapping those keys to the keys used in [`SENATE_CLASSES`](https://github.com/18F/fec-cms/blob/develop/fec/data/constants.py#L632-L636) constant.
* Modify the [`get_senate_cycles`](https://github.com/18F/fec-cms/blob/develop/fec/data/api_caller.py#L348-L357) method in the `api_caller.py` module to iterate through the `NEXT_SENATE_ELECTIONS` items instead, and pass the value of the current dictionary item to this method instead.
* Keep this method, but instead of doing a check on the `senate_class`, just use the value passed in as is as a part of the calculation, e.g., `next_election = current_cycle() + senate_class`; perhaps rename the argument to something more descriptive, such as `senate_class_range`.

So in the end, here is what we end up with based on these suggestions.  We keep the `NEXT_SENATE_ELECTIONS` constant but modify it to this:

```python
NEXT_SENATE_ELECTIONS = {
    '1': 0,
    '2': 2,
    '3': 4
}
```

For the  [`get_senate_cycles`](https://github.com/18F/fec-cms/blob/develop/fec/data/api_caller.py#L348-L357) method, it turns into this:

```python
def get_regular_senate_cycles(state):
    """ Get the list of election cycles based off Senate class
    """
    senate_cycles = []

    for senate_class, senate_class_range in constants.NEXT_SENATE_ELECTIONS.items():
        if state.upper() in constants.SENATE_CLASSES[senate_class]:
            senate_cycles += utils.get_senate_cycles(senate_class_range)

    return senate_cycles
```

And the method above changes to this:

```python
def get_senate_cycles(senate_class_range):
    next_election = current_cycle() + senate_class_range
    return range(next_election, constants.START_YEAR, -6)
```

----

**HOWEVER**, I think this could all be cleaned up even more if we figured out the calculation we could leverage so we don't even need to keep track of different ranges for each senate class.  @lbeaufort and @LindsayYoung, are you able to weigh in on this at all?

lbeaufort: @ccostino I like your approach here, but I'm not sure I 100% follow. I'll have to take a closer look. Here's what I came up with, which doesn't look great but also wouldn't need a constants file. 
```
def get_next_senate_elections(current_cycle):
    """
    Returns an dictionary of senate classes in chronological order
    based on current cycle year
    Uses mod 6 to determine what senate classes are up for election
    in the current cycle

    For example, class 1 will always be up for election in years 
    with year % 6 = 2, for example 2018. 
    """

    if current_cycle % 6 == 2:
        return {
                '1': current_cycle,
                '2': current_cycle + 2,
                '3': current_cycle + 4
                }
    if current_cycle % 6 == 4:
        return {
                '2': current_cycle,
                '3': current_cycle + 2,
                '1': current_cycle + 4
                }
    if current_cycle % 6 == 0:
        return {
                '3': current_cycle,
                '1': current_cycle + 2,
                '2': current_cycle + 4
                }


def get_senate_cycles(senate_class):
    """
    Returns an list of elections based on senate class
    Uses get_ordered_senate_classes to find out which classes are up for election
    Adds years to current cycle depending on what position the class is in
    """

    senate_classes = get_next_senate_elections(current_cycle())

    next_election = senate_classes[senate_class]

    return range(next_election, constants.START_YEAR, -6)
```

ccostino: Thank you, @lbeaufort!

First, what parts of what I had outlined above were confusing?  I'd be happy to answer questions or clarify further to make sure everyone understands!

Second, your suggestion is exactly what I was looking for:  the formula/pattern for how this could be calculated/derived!  Bonus points for the clear explanation and example in the method doc string! 🎉 :-)

This is definitely a step in the right direction, and I feel like it could be tightened up further, but I'm also not sure how at the moment.   I'll keep mulling it over a bit and ask for some additional feedback, but if nothing comes to fruition soon we could just move ahead as is.  Thanks again!

fec/data/tests/test_utils.py
```diff
@@ -69,7 +69,11 @@ def test_get_cycles(self, current_cycle):
         assert utils.get_cycles(2020) == range(2020, 1979, -2)
 
     def test_get_senate_cycles(self):
-        assert utils.get_senate_cycles(1) == range(2018, 1979, -6)
+        current_cycle = 2018
+        assert utils.get_senate_cycles('1') == range(current_cycle, 1979, -6)
+        assert utils.get_senate_cycles('2') == range(current_cycle + 2, 1979, -6)
+        assert utils.get_senate_cycles('3') == range(current_cycle + 4, 1979, -6)
+
```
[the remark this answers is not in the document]
fec-jli: Thank You, this is first time I use github PR to track the codes. 
I am happy to start using github.

fec/data/utils.py
fec-jli: @lbeaufort great, that is very clear. will test in my local. Thank You.

fec-jli: @ccostino 
for 2018: 
NEXT_SENATE_ELECTIONS = {
    '1': 0,
    '2': 2,
    '3': 4
}

for 2020:
NEXT_SENATE_ELECTIONS = {
    '1': 4,
    '2': 0,
    '3': 2
}

for 2022;
NEXT_SENATE_ELECTIONS = {
    '1': 2,
    '2': 4,
    '3': 0
}
.... 2024,2026...

this is the rule, look like it is not easy to put into constant.py
Thanks

ccostino: @fec-jli you're right, it's not easy to put in `constants.py`, nor would I want to - we should be able to derive it based on the calculations from the formula.  @lbeaufort's work does this, and I wonder if there's a way we can make it even more concise.

fec-jli: @ccostino I am a java developer before, not familiar with Python, I find out there is no "switch / case" method in Python after google online.
I agree with you there should be a more concise way to do this.
that is a math problem:))
for now, We will use Laura's way to see if we can find better way in the future.

ccostino: @fec-jli I understand, which is why I wanted to focus on it a bit; to provide more Python exposure, including how to think with the language! :-)  Yes, the missing `switch` statement is a departure from other languages such as Java, C++, C, etc.

I agree that if there doesn't seem to be any improvement at the moment to just move ahead, but I wanted to make sure we did our due diligence in thinking about it first!

LindsayYoung:
How about something along the lines of 
```
next_election_decoder = {
    2:{
        '1': current_cycle,
        '2': current_cycle + 2,
        '3': current_cycle + 4
    },
    4:{
        '2': current_cycle,
        '3': current_cycle + 2,
        '1': current_cycle + 4
    },
    0:{
        '3': current_cycle,
        '1': current_cycle + 2,
        '2': current_cycle + 4
    }
}

senate_elections = next_election_decoder[current_cycle %6] 
```
Please double check I assigned the cycles correctly

fec/data/utils.py
```diff
@@ -206,8 +205,44 @@ def process_ie_data(totals):
...
+    elif (current_cycle() - constants.END_YEAR + 2) % 6 == 0:
+        current_year_class = '3'
+    return current_year_class
+
+
 def get_senate_cycles(senate_class):
-    next_election = constants.NEXT_SENATE_ELECTIONS[str(senate_class)]
+
+    if get_current_year_class() in '1':
+        if senate_class in '1':
```
LindsayYoung: One python style tip here is to use `==` for checking equality

fec-jli: @LindsayYoung Thank you, got it. 
where to open Python tip? in sublime?

fec-jli:
@LindsayYoung  you are so great. I spent a couple of hours to research and write two level dictionary, you already put here 3 hours ago:)  almost same as yours. 
I don't know why I cannot reply under your codes. Thank You.

-------
next_election_decoder = {
    2:{
        '1': current_cycle,
        '2': current_cycle + 2,
        '3': current_cycle + 4
    },
    4:{
        '2': current_cycle,
        '3': current_cycle + 2,
        '1': current_cycle + 4
    },
    0:{
        '3': current_cycle,
        '1': current_cycle + 2,
        '2': current_cycle + 4
    }
}

fec/data/utils.py
```diff
@@ -206,9 +205,33 @@ def process_ie_data(totals):
...
+    """
+    next_election_cycles = {
+        2: {'1': current_cycle, '2': (current_cycle + 2), '3': (current_cycle + 4)},
+        4: {'2': current_cycle, '3': (current_cycle + 2), '1': (current_cycle + 4)},
+        0: {'3': current_cycle, '1': (current_cycle + 2), '2': (current_cycle + 4)}
+    }
+    return next_election_cycles.get(current_cycle % 6)
+
+
+def get_senate_cycles(senate_class, current_cycle=current_cycle()):
```
lbeaufort: I'm not sure I've seen a function run in the parameters before!

`def get_senate_cycles(senate_class, current_cycle=current_cycle()):`

lbeaufort: Very cool.

fec/data/tests/test_utils.py
```diff
@@ -69,7 +69,30 @@ def test_get_cycles(self, current_cycle):
         assert utils.get_cycles(2020) == range(2020, 1979, -2)
 
     def test_get_senate_cycles(self):
-        assert utils.get_senate_cycles(1) == range(2018, 1979, -6)
+        current_cycle = 2018
```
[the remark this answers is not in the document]
ccostino: Big +1 to this, definitely great to see! :-)

fec/data/utils.py
ccostino: I would caution against this if only from a maintenance point of view.  I don't think there's anything wrong per se, but a statement like this would be easy to misread (case in point, I missed it last night!) and therefore not pick up on the fact that a function call is happening as a part of the argument assignments in this method definition.  I'd recommend one of two things here, depending on what our needs are.

If it helps to be able to pass in a `cycle` to this method (e.g., we have a situation where we'd pass in an election cycle other than the current one), then I'd keep the parameter but make it optional and set its default to `None`, then check for that and set it inside the method itself if need be:

```python
def get_senate_cycles(senate_class, current_cycle=None):
    """
    Returns an list of elections based on senate class
    Uses get_next_senate_elections to find out
    which classes are up for election
    Adds years to current cycle depending on what position the class is in
    """

    if cycle is None:
        cycle = current_cycle()

    senate_classes = get_next_senate_elections(cycle)
    return range(senate_classes.get(senate_class), constants.START_YEAR, -6
```

If this is not the case, and we only ever want to deal with the current cycle here, then I'd just remove the parameter all together and use `current_cycle()` directly in the method itself:

```python
def get_senate_cycles(senate_class):
    """
    Returns an list of elections based on senate class
    Uses get_next_senate_elections to find out
    which classes are up for election
    Adds years to current cycle depending on what position the class is in
    """

    senate_classes = get_next_senate_elections(current_cycle())
    return range(senate_classes.get(senate_class), constants.START_YEAR, -6)
```

This way it's much clearer to see what's happening and reduce any potential unintended consequences from things being called in spots that aren't generally expected. :-)

fec-jli: I add current_cycle as parameter , the reason is to test whole 6 year cycle in data/tests/test_utils.py,
I want to make sure the logic work correctly for all cycles in the future.

if cause the confuse here, I prefer to do this:
   if cycle is None:
        cycle = current_cycle()

Thanks for suggestion.

ccostino: Yup, that's totally fine and I think you did the right thing there, especially with multiple tests! :-)

It sounds like the first use-case is appropriate, so modifying the parameter to default to `None` to make it optional and then falling back to the current cycle if no value is passed in will serve the need here, and put us in a good spot going forward. :-)

ccostino:
Awesome, awesome work everyone! 🎉 :-)

fec/data/utils.py
ccostino: I know this is a little after the fact, but I did want to highlight!

No, this is just a tip @LindsayYoung is sharing about performing equality checks in Python.  The [`in`](https://docs.python.org/3.5/library/stdtypes.html#common-sequence-operations) keyword in Python is used to check and see if a value is in a list (array), but the [`string` (`str`)](https://docs.python.org/3.5/library/stdtypes.html#str) type has a [`string sequence`](https://docs.python.org/3.5/library/stdtypes.html#textseq) quality to it, which allows strings to behave like lists of characters.

In this case, the value of `senate_class` is `'1'`, `'2'`, or `'3'`, which are single character strings and therefore could potentially match the string `'1'`, but this isn't a good place for checking with `in`.  What we'd be trying to check here is if the `senate_class` value is _equal_ to the other string value, not if it's a value in a string sequence (list).

Does this help explain what we mean?  If this doesn't make sense, please let us know and we can explain further!

fec-jli: got it. Thanks again.

PR #1530, opened 2017-11-29
set npm package.json

xtine:
## Set NPM details
- Set `package.json` details (name, version) so that `fec-cms` can be initialized as a local NPM package for `fec-pattern-library` to access assets.
- Remove `all.scss` because pattern library will be able access scss files to compile.

Does not impact any part of the working application. Setting up `fec-cms` as an npm package only sets details for local usage.


package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "fec-cms",
+  "version": "1.0.0",
```
ccostino: Should we set the `license` and `repository` too while we're at it just to get rid of the warnings?

xtine: What `license` should we use? I just added the repo info.

ccostino: 🤔 really good question, I'm not sure!  We'll have to ask around and see what would be appropriate.

xtine: Just checked, `fec-style` doesn't have a license and it's a published package... 😆

ccostino: Whoops, that probably shouldn't have been the case, d'oh!

xtine: Okay, we have quickly discovered https://github.com/18F/open-source-policy/blob/master/policy.md#open-source-licenses, which directs us to use `CC0`. It's already defined on the bottom of our `LICENSE.md`, but I'm explicitly including in the license field here as well.

PR #1636, opened 2017-12-08
[MERGE WITH GITFLOW] Hotfix: Correct Candidate two-year summaries

xtine:
## Summary
- Addresses https://github.com/18F/fec-cms/issues/1634
- The `election_full` url query string used to be passed into the application in openFEC-web-app as a boolean. However after merge, we failed to account for this and passed it into `data/views.py` as a string. This caused a failure in logic because if we check `if election_full`, it would always return `True` as a string. We fix this by converting `election_full` from a string to a boolean right after the get request.
- I also fixed a UI bug caused by newer versions (62+) of Chrome that would show overflows on side bars that take up the entire window which affected Candidate pages.

- Candidate summary pages.

## Screenshots
Correct behavior:
screen shot 2017-12-08 at 2 40 26 pm

Fixes this Chrome overflow issue with side nav: 
screen shot 2017-12-08 at 1 21 21 pm


xtine:
I'm just going to create a new PR. :P

PR #1637, opened 2017-12-08
[MERGE WITH GITFLOW] Hotfix/fix candidate 2 yr summary cycle

xtine:
## Summary
- Addresses https://github.com/18F/fec-cms/issues/1634
- The `election_full` url query string used to be passed into the application in openFEC-web-app as a boolean. However after merge, we failed to account for this and passed it into `data/views.py` as a string. This caused a failure in logic because if we check `if election_full`, it would always return `True` as a string. We fix this by converting `election_full` from a string to a boolean right after the get request.
- I also fixed a UI bug caused by newer versions (62+) of Chrome that would show overflows on side bars that take up the entire window which affected Candidate pages.

- Candidate summary pages.

## Screenshots
Correct behavior:
screen shot 2017-12-08 at 2 40 26 pm

Fixes this Chrome overflow issue with side nav: 
screen shot 2017-12-08 at 1 21 21 pm

### DIFF
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -1,7 +1,10 @@
 from django.shortcuts import render, redirect
+from django.urls import reverse
 from django.http import Http404
 from django.http import JsonResponse
 from django.conf import settings
+
+from distutils.util import strtobool
 
 import datetime
 import github3
@@ -89,6 +92,9 @@ def candidate(request, candidate_id):
     if cycle is not None:
         cycle = int(cycle)
 
+    # set election_full to boolean from passed string variable
+    election_full = bool(strtobool(str(election_full)))
+
     candidate, committees, cycle = api_caller.load_with_nested(
         'candidate', candidate_id, 'committees',
         cycle=cycle, cycle_key='two_year_period',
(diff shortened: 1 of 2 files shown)


ccostino:
@cptechiegal you're right, having a test case would be useful to guard against any future regressions and ought to be a part of the work.  There's a larger piece to that, though, in that we don't have a good test harness in place at the moment to catch these types of bugs and easily write tests for them when it comes to the CMS, especially after the work done to merge in the web app.

In this particular case, there is the backend side of it (which we could write tests for to ensure that all query string parameters are properly processed, not even just this particular endpoint) and there is the visual component side of it, which tipped us off of the bug in the first place (visual components tend to be tested with something like [Selenium](http://www.seleniumhq.org/), but that requires a large amount of setup and maintenance in and of itself - there are other ways to test on the client-side, though).

This is work that needs to be done, but it hasn't been prioritized over other things.  That said, everyone on the team is certainly empowered and encouraged to create new issues for things that they notice!  So I'd encourage you to just create the new issue with what you feel needs to be done and value it will add when you notice things like this. :-)  When you do, please just add the `Needs grooming` and `Needs prioritization` labels to it and make sure it's brought up in a future sprint grooming session so it can be slotted for a sprint planning session in the future.  This way the team won't lose track of the work and it can be properly accounted for and prioritized against other user stories.

PR #1638, opened 2017-12-11
Edit icon assets for sizing, relevance, and consistency

jenniferthibault:
## Summary

This PR makes some small changes that clean up icon files. It:
- Removes an unused icon that we don't plan to reuse in the future (because it was a specifically styled representation of a person, and our style has become more generic)
- Replaces the hand icons that hold money and an envelope to make them the same size (before they were different base sizes)
- Adjusts the size of the `hand-money` icon in context to account for its new size
- Reduced the size of the external sources card icon, so that it's visually consistent with the other card icons. That's unrelated to the above changes, but has been bugging me for a while

- Data landing page: https://www.fec.gov/data/ 

## Screenshots

**Before**
screen shot 2017-12-11 at 1 00 01 pm

**After**
screen shot 2017-12-11 at 12 59 52 pm


xtine:
`info-person` should also be removed here: https://github.com/18F/fec-cms/blob/93934c424f8ac12addfe688e01ca8174ae4e0cec/fec/fec/static/scss/_icon-variables.scss#L62 and https://github.com/18F/fec-cms/blob/develop/fec/fec/static/icons/output/info-person.svg

xtine:
@jenniferthibault: you're right! `_icon-variables.scss` is automatically generated from the `output` folder, so just removing old icons from `output` should solve that. I'll tweak the build icons task to clear output folder so out of date icons are removed.

Issue #1640, opened 2017-12-11
Create a feature workspace so new concepts can be tested away from the release cadence


jenniferthibault:
Designers and researchers need to be able to prototype new/improved concepts in code in an environment that's separate from the release pipeline environments, `dev` and `staging`. Currently, our only option is to run a local instance of the site or to deploy testing features to `dev`, which has resulted in features prematurely being merged into a release, or merged without thorough review. 

Access to a `feature` space on the `fec-cms` repo will: 
- Allow us to better simulate complex interactions with real data
- Create a publicly shareable URL that doesn't require researchers (or users) to run the site locally in order to conduct usability tests. 
- Prevent sudden blockers to release schedules because a feature being tested in `dev`tests poorly. requires further design changes based on testing, or isn't otherwise fully release-ready. 

**Completion criteria:**

- [x] Credentials are added to cloud.gov for `fec-cms`  
- [x] Make a CMS database
- [x] Import latest CMS data into database
- [x] Push the CMS app (via Lindsay "We wouldn't add API, celery worker celery beat, eregs or proxy to it")
- [x] Hook up the `FEC_API_URL` to dev
- [x] Create S3 bucket to store wagtail uploaded files and images
- [x] Add instructions to README on creating, maintaining, and using a feature space
- [x] GitHub branches can be directed toward the new feature space :ta-da:

jenniferthibault:
It's perhaps helpful to know that this is a capability we used to have, but I'm not sure why it went away. Could be something to do with merging all of the github repositories?

ccostino:
It largely went away due to resource constraints in our cloud.gov org account, and having to manually deploy to keep it up to date.  As long as we have the resources (and I believe we do), this shouldn't be difficult to setup now.  We might still need to do the manual deployments for the time being but we could try and tie-in some workflows with CircleCI for automated ones as well in the future I imagine. :-)

ccostino:
This is mostly ready to go!  The last thing I'm unsure about is wiring up the S3 bucket for this space.  I did create a new one and bound it to the app, but I just need to double check what else is needed to use it.

The only other thing left is getting some instructions written up.

ccostino:
All set and ready for review!

ccostino:
**NOTE:**  There is no automated deployer setup with the `feature` space, and for now that's intentional; you can only manually deploy to it.  If we want to support automated deploys to it, I think that should be a separate issue and work.

ccostino:
[...]

PR #1700, opened 2018-01-12
Add support back in for the cloud.gov feature space

### LINKED ISSUE #904: Use S3 credentials directly from S3 service
I was looking into an unrelated issue with @jontours earlier today and noticed that a user-provided service includes credentials that appear to be used to access an S3 service instance. There's nothing wrong with that, but it's generally simpler to access service credentials by binding services to applications, rather than copying and pasting credentials into a user-provided service. Do you all have a requirement to pass credentials in a user-provided service rather than binding to the S3 instance?

cc @LindsayYoung @ccostino

ccostino:
Fixes #904 
Addresses #1640 

This changeset makes the adjustments necessary to support the CMS in the feature space again within our cloud.gov account.  This also includes an updated CircleCI configuration to help improve that setup a bit.

**NOTE:** When this is merged into `develop` and deployed to the `dev` space and then subsequently deployed to the `stage` and `prod` spaces in the next release, this should automatically bind the S3 content bucket to the app in the respective environment as well.

### DIFF
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -48,6 +48,7 @@ ENVIRONMENTS = {
     'dev': 'DEVELOPMENT',
     'stage': 'STAGING',
     'prod': 'PRODUCTION',
+    'feature': 'FEATURE',
 }
 FEC_CMS_ENVIRONMENT = ENVIRONMENTS.get(env.get_credential('FEC_CMS_ENVIRONMENT'), 'LOCAL')
 CONTACT_EMAIL = 'webmanager@fec.gov'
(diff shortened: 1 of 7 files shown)


fec/fec/settings/base.py
```diff
@@ -248,14 +249,13 @@
...
+    AWS_SECRET_ACCESS_KEY = env.get_credential('secret_access_key')
+    AWS_STORAGE_BUCKET_NAME = env.get_credential('bucket')
+    AWS_S3_REGION_NAME = env.get_credential('region')
     AWS_S3_CUSTOM_DOMAIN = env.get_credential('CMS_AWS_CUSTOM_DOMAIN')
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_LOCATION = 'cms-content'
-    AWS_S3_REGION_NAME = env.get_credential('CMS_AWS_DEFAULT_REGION')
-
 
 UAA_CLIENT_ID = env.get_credential('CMS_LOGIN_CLIENT_ID', 'my-client-id')
```
patphongs: This is the login settings for the feature space to login via cloud.gov. Do we need separate `CMS_LOGIN_CLIENT_ID` and `CMS_LOGIN_CLIENT_SECRET` in order for wagtail login on feature to work? Otherwise, we get a mismatch for the registered value when attempting to login?

ccostino: Oh good question, @patphongs!  Did you try logging in yet?  If so, did you notice this happening, or are you asking if it will happen?

If the latter, I had tried logging in to the admin and was able to get in just fine, so I didn't notice anything wonky.  Still, it'd be good to check!

patphongs: @ccostino I noticed this happening when attempting to login to the wagtail admin from feature. I get the following error message `Invalid redirect https://fec-feature-cms.app.cloud.gov/auth/callback did not match one of the registered values`. I recall this happening on dev at one point and I think you had fixed it, but I don't remember what the solution was.

ccostino: Great catch, we'll have to take a look at some of the older issues/PRs around that or see if there's documentation, I don't remember off the top of my head what we had to do.

ccostino: Found the documentation!

* [cloud.gov Identity Provider Documentation](https://cloud.gov/docs/services/cloud-gov-identity-provider/)
* [cg-django-uaa Quick Start Guide](http://cg-django-uaa.readthedocs.io/en/latest/quickstart.html)

I updated the credentials in `feature` based on this info and restaged the CMS application.  @patphongs, would you mind trying again?  Thanks!

patphongs: works, thank you!

ccostino: [Documentation is now in the wiki!](https://github.com/18F/fec-cms/wiki/Setup-and-manage-UAA-credentials)

PR #1701, opened 2018-01-12
Add Coverage end date to Source reports

johnnyporkchops:
This PR addresses a few issues that are in grooming. It may be salvaged to implement an interim solution that is being discussed, if/when one is agreed upon. See issues referenced below for discussions:
- Addresses [1735](https://github.com/18F/fec-cms/issues/1735) (in grooming)
- Addresses  [1470](https://github.com/18F/fec-cms/issues/1470) **(Currently implemented solution)**

Also these two OpenFEC  PRs have been merged by @lbeaufort to accommodate possible solutions:
https://github.com/18F/openFEC/pull/2901- Add committee_pcc to /elections/ endpoint
https://github.com/18F/openFEC/pull/2898 -Add coverage end date to /elections/endpoint

**Completion criteria, Issue 1470:**
- [x] change source reports link text from "view", to "View all"
- [x] /elections/  EndPoint API needs to be modified to return `coverage end date`
- [x] add handlebars template to handle this particular mullt-line/conditional content within datatable column

@hcaofec has done some research into  backend changes that need to take place and may be of help.

## Screenshots
jan-12-2018 15-42-46


johnnyporkchops:
feb-02-2018 06-46-29

johnnyporkchops:
The team decided leaving the View Link url as is, rather than a temporary change is the way to go. However , adding the coverage end date is going to go forward after some grooming from content team.

johnnyporkchops:
-[ ] Added the source reports coverage end date and language . 
-[ ] Created handlebars templates to render in column
final

JonellaCulmer:
@johnnyporkchops What's the line height in the source reports column?

johnnyporkchops:
@JonellaCulmer , line height on the `td` is 1.8, inherited. A similar example of multiline text in columns is the Version column on the  reports datatables.

johnnyporkchops:
Thanks @patphongs . I was thinking I do not need the ` if url ` statement because there will always be a URL, but we can change that when/if  we refine this for [1735](https://github.com/18F/fec-cms/issues/1735)

johnnyporkchops:
@lbeaufort  -- Is this an issue where this one candidate who does not return a committee ID in the URL that is built for this page?
The example page: https://www.fec.gov/data/elections/senate/FL/2018/
SHOEMAKER, FRANKLIN ED
The URL created: https://www.fec.gov/data/reports/house-senate/?committee_id&cycle=2018&is_amended=false
I think you said you had some insight into this when we discovered it while pairing on this task.

PR #1706, opened 2018-01-17
[WIP - Don't merge yet] Fixed RFAIs that reference Form 1

patphongs:
## Summary

- Addresses https://github.com/18F/fec-cms/issues/1663
_RFAI's that reference a statement of organization, also known as `RQ2`, was not showing in the filings table. The problem was that nulls were being filtered out from the regularly filed reports. Now that null values are shown, we can now see all the RFAI's within the Regularly filed reports section of the committee's filing tab_

**Note:** It was requested that we group the RFAIs that reference form 1 with the statement of organization. However, after reviewing the code, it is a lot more difficult to implement this given the current abilities of our API. A possible easier solution would be to refactor the RFAIs to their own section that is separate from the Regularly filed reports section.  

-  Committee's filing tab

## Screenshots
Below is an example with all the RFAIs shown for committee ID C00637983
screen shot 2018-01-17 at 3 08 57 pm

### DIFF
--- a/fec/fec/static/js/pages/committee-single.js
+++ b/fec/fec/static/js/pages/committee-single.js
@@ -570,7 +570,8 @@ $(document).ready(function() {
           path: ['committee', committeeId, 'filings'],
           query: _.extend({
             form_type: ['F3', 'F3X', 'F3P', 'F3L', 'F4', 'F7', 'F13', 'RFAI'],
-            sort: ['-coverage_end_date', 'report_type_full', '-beginning_image_number']
+            sort: ['-coverage_end_date', 'report_type_full', '-beginning_image_number'],
+            sort_hide_null: ['false']
           }, query),
           callbacks: {
             afterRender: filings.renderModal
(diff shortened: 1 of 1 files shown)


bmathesonFEC:
@patphongs Just to confirm, there's no way to select a NULL or NA value? Similarly, is there no way to exclude (<>) criteria in an API call, right?  I'm asking because I think this could get us the results we need:
(form_type = F1 OR form_type = RFAI) AND report_type = NA

(Sorry to ask questions you've probably already asked and answered)

patphongs:
@bmathesonFEC Good question! We can show nulls, however, there's no way to show nulls of only a specific form_type or report_type right now. The easiest way would be for the data to be cleaned up. Right now any `request_type = RQ1` has a null `form_type`. If we can populate the `form_type` for RQ1's to be form_type=F1, then the Form 1 RFAIs will show up. Still trying to see how the form_type gets populated, as RQ2s seem to have a form type, just not RQ1s.

patphongs:
Closing this issue, exclude logic has been added to the API capabilities and this PR is now old and was replaced by this PR: https://github.com/fecgov/fec-cms/pull/1835

PR #1708, opened 2018-01-20
Include odd year special elections in Election dropdown as even years

cptechiegal:
## Summary 
- Addresses Issue# 1644 https://github.com/18F/fec-cms/issues/1644
Handles odd year special elections by changing to even year

- fec-cms/fec/data/views.py 
The candidate() function was updated to include new code for creating a new election years list of only even years from the API's returned election-years list.
- fec-cms/fec/fec/data/templates/macros/cycle-select.jinja
This macro uses the election years list for displaying the election years in the Elections pulldown menu. 

## Screenshots
Test Case: House Candidate Karen Handel had special election in 2017
https://www.fec.gov//data/candidate/H8GA06286/

-Before screenshot
issue1644-odd year special elections-before

- After screenshot
issue1644-odd year special elections-after

## Related PRs
List related PRs against other branches: none

## Related Issues
Issue:2845 https://github.com/18F/openFEC/issues/2845 
- This PR1708 also resolves issue2845 with Time Periods 
____

### DIFF
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -12,7 +12,12 @@
       >
     {% for each in cycles | sort(reverse=True) %}
       <option
+          {# If odd year election, increment to even election year #}
+          {% if each % 2 > 0  %}
+          value="{{ each + 1}}"
+          {% else %}
           value="{{ each }}"
+          {% endif %}
           {% if cycle and cycle <= each and cycle > (each - duration) %}selected{% endif %}
         >
         {% if range %}
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -11,14 +11,10 @@
         data-duration="{{ duration }}"
       >
     {% for each in cycles | sort(reverse=True) %}
-      {# if each is odd year special election, change to even year #}
-      {% if (each % 2 > 0) %}   
-        {% set each = each+1 %}
-      {% endif %}
-      <option
+        <option
           value="{{ each }}"
           {% if cycle and cycle <= each and cycle > (each - duration) %}selected{% endif %}
-      >
+        >
         {% if range %}
           {{ each|fmt_year_range }}
         {% else %}
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -9,7 +9,7 @@
         name="cycle"
         data-cycle-location="{{ location }}"
         data-duration="{{ duration }}"
-      >
+    >
     {% for each in cycles | sort(reverse=True) %}
         <option
           value="{{ each }}"
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -11,15 +11,14 @@
         data-duration="{{ duration }}"
       >
     {% for each in cycles | sort(reverse=True) %}
+      {# if each is odd year special election, change to even year #}
+      {% if (each % 2 > 0) %}   
+        {% set each = each+1 %}
+      {% endif %}
       <option
-          {# If odd year election, increment to even election year #}
-          {% if each % 2 > 0  %}
-          value="{{ each + 1}}"
-          {% else %}
           value="{{ each }}"
-          {% endif %}
           {% if cycle and cycle <= each and cycle > (each - duration) %}selected{% endif %}
-        >
+      >
         {% if range %}
           {{ each|fmt_year_range }}
         {% else %}
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -107,8 +107,8 @@ def candidate(request, candidate_id):
         cycle=cycle, cycle_key='two_year_period',
         election_full=election_full,
     )
-# cycle corresponds to the two-year period for which the committee has financial activity.
-# when selected election cycle is not in list of election years, get the next election cycle
+    # cycle corresponds to the two-year period for which the committee has financial activity.
+    # when selected election cycle is not in list of election years, get the next election cycle
     if election_full and cycle and cycle not in candidate['election_years']:
 
         next_cycle = next(
@@ -155,7 +155,7 @@ def candidate(request, candidate_id):
     #  for displaying elections for pulldown menu in Candidate pages
     even_election_years = list()  # empty list
     for year in candidate['election_years']:
-        print("year is {}".format(year))
+ #       print("year is {}".format(year))
         if year % 2 > 0 :
             even_election_years.append(year+1)  # make even year
         else:
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -10,8 +10,6 @@ import datetime
 import github3
 import json
 import re
-# python debugger
-import pdb      
 
 from data import api_caller
 from data import constants
@@ -87,15 +85,11 @@ def advanced(request):
 
 
 def candidate(request, candidate_id):
-    pdb.set_trace()
     # grab url query string parameters
     cycle = request.GET.get('cycle', None)
-    print("cycle is:{}".format(cycle))
-    #print("election_years is:{}".format(election_years))
    
     election_full = request.GET.get('election_full', True)
-    print("election_full is:{}".format(election_full))
-
+ 
     if cycle is not None:
         cycle = int(cycle)
 
@@ -155,8 +149,7 @@ def candidate(request, candidate_id):
     #  for displaying elections for pulldown menu in Candidate pages
     even_election_years = list()  # empty list
     for year in candidate['election_years']:
- #       print("year is {}".format(year))
-        if year % 2 > 0 :
+         if year % 2 > 0 :
             even_election_years.append(year+1)  # make even year
         else:
             even_election_years.append(year)  # already even year
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -147,8 +147,8 @@ def candidate(request, candidate_id):
 
      # Addresses issue#1644 - make any odd year special election an even year 
     #  for displaying elections for pulldown menu in Candidate pages
-    #  Using Set to ensure no duplicate years in list
-    even_election_years = { year + (year % 2) for year in candidate.get('election_years', []) }
+    #  Using Set to ensure no duplicate years in final list
+    even_election_years = list({ year + (year % 2) for year in candidate.get('election_years', []) })
 
     # In the case of when a presidential or senate candidate has filed
     # for a future year that's beyond the current cycle,
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -149,7 +149,7 @@ def candidate(request, candidate_id):
     #  for displaying elections for pulldown menu in Candidate pages
     even_election_years = list()  # empty list
     for year in candidate['election_years']:
-         if year % 2 > 0 :
+        if year % 2 > 0 :
             even_election_years.append(year+1)  # make even year
         else:
             even_election_years.append(year)  # already even year
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -145,15 +145,10 @@ def candidate(request, candidate_id):
         'candidateID': candidate['candidate_id']
     }
 
-    # Addresses issue#1644 - make any odd year special election even years 
+     # Addresses issue#1644 - make any odd year special election an even year 
     #  for displaying elections for pulldown menu in Candidate pages
-    even_election_years = list()  # empty list
-    for year in candidate['election_years']:
-        if year % 2 > 0 :
-            even_election_years.append(year+1)  # make even year
-        else:
-            even_election_years.append(year)  # already even year
-
+    #  Using Set to ensure no duplicate years in list
+    even_election_years = { year + (year % 2) for year in candidate.get('election_years', []) }
 
     # In the case of when a presidential or senate candidate has filed
     # for a future year that's beyond the current cycle,
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -10,6 +10,8 @@ import datetime
 import github3
 import json
 import re
+# python debugger
+import pdb      
 
 from data import api_caller
 from data import constants
@@ -85,9 +87,14 @@ def advanced(request):
 
 
 def candidate(request, candidate_id):
+    pdb.set_trace()
     # grab url query string parameters
     cycle = request.GET.get('cycle', None)
+    print("cycle is:{}".format(cycle))
+    #print("election_years is:{}".format(election_years))
+   
     election_full = request.GET.get('election_full', True)
+    print("election_full is:{}".format(election_full))
 
     if cycle is not None:
         cycle = int(cycle)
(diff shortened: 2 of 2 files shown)


fec/data/templates/macros/cycle-select.jinja
```diff
@@ -11,10 +11,14 @@
         data-duration="{{ duration }}"
       >
     {% for each in cycles | sort(reverse=True) %}
+      {# if each is odd year special election, change to even year #}
+      {% if (each % 2 > 0) %}   
+        {% set each = each+1 %}
```
xtine: This seems to be pushing more logic into the template, I suggest to move this into the view instead.

cptechiegal:
Reviewed the web app view.py code for handling odd year special elections. 
Found existing code related to Issues https://github.com/18F/fec-cms/issues/1945 and 1612 
(1945 originally reported 404 errors similar to issue 1642)

fec/data/templates/macros/cycle-select.jinja
```diff
@@ -11,7 +11,7 @@
         data-duration="{{ duration }}"
       >
     {% for each in cycles | sort(reverse=True) %}
-      <option
+        <option
```
patphongs: fix option tag indent to line up

fec/data/views.py
```diff
@@ -143,6 +145,16 @@ def candidate(request, candidate_id):
         'candidateID': candidate['candidate_id']
     }
 
+    # Addresses issue#1644 - make any odd year special election even years 
+    #  for displaying elections for pulldown menu in Candidate pages
+    even_election_years = list()  # empty list
+    for year in candidate['election_years']:
+        if year % 2 > 0 :
```
patphongs: remove the space between the `0` and `:`

fec/data/views.py
```diff
@@ -143,6 +145,16 @@ def candidate(request, candidate_id):
...
+            even_election_years.append(year+1)  # make even year
+        else:
```
patphongs: I would suggest removing the `else` as it's not really needed. By default it will return the `election_year` that is already even anyways.

cptechiegal: Hi @pat 
The else is needed, otherwise the  even_election_years list won't contain the election years that are already even.

patphongs: @cptechiegal You're right! Completely missed that, thanks for looking into it

patphongs:
@pat sorry for accidentally pinging you in this PR, I'm the Pat that this is meant for :)

P.S. I like your username!

fec/data/views.py
lbeaufort: @patphongs @cptechiegal I believe a more "Pythonic" way of writing this would be: 
```
even_election_years = []
for year in candidate['election_years']:
    even_election_years.append(year + year % 2)
```
@ccostino  thoughts?

fec/data/views.py
```diff
@@ -143,6 +145,16 @@ def candidate(request, candidate_id):
         'candidateID': candidate['candidate_id']
     }
 
+    # Addresses issue#1644 - make any odd year special election even years 
+    #  for displaying elections for pulldown menu in Candidate pages
+    even_election_years = list()  # empty list
```
ccostino: When instantiating a variable to an empty data structure, I recommend using the literal syntax (and this goes for JavaScript as well).  In this case, you'd use `[]` for the empty list.  This is both for clarity and for reducing function calls to improve the performance of your code.

However, please see notes below as this is likely not even needed and could be improved further!

fec/data/views.py
ccostino: @lbeaufort that's a great start, and I think it can be taken even further with a [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) and perhaps something else.  Let's explore this!

NOTE:  I recommend taking a look through that whole tutorial, especially for folks still getting their feet under them with the Python side of things. :-)

One thing though, before we dive into that:  In addition to the note I left above at the start of this block regarding instantiating data structures, I want to call out again about being careful when accessing dictionary keys directly (e.g., `candidate['election_years']`) as this can result in a `KeyError` being thrown (meaning the dictionary key is not found).  When dealing with information coming from requests in a web app, this can happen a fair amount so I'd recommend going the route of using the [`get()`](https://docs.python.org/3/library/stdtypes.html#dict.get) method and provided a default value so your code always runs as you expect.  In this case, since it's a list being expected in the processing, you could write this:

```python
candidate.get('election_years', [])
```

This way you are guaranteed to be processing a list, even if it's empty (which means the loop won't do anything).

All of that said, back to list comprehensions:  these are a cornerstone of Python and one of its greatest strengths and tools.  This entire block could be reduced to a single line of code with a list comprehension and leveraging the fact that the modulo operation in the equation `x + (x % 2)` (x being a numeric value, the election year in this case) will return either the year as is (if it is even) or the desired cycle year (the end year in the cycle) if it's odd.  Here's what I mean:

```python
# "cycles" is a common term and variable that we use for this kind of data throughout both applications
# because it denotes a collection (list) of years that meet a certain criteria:  that they are all even numbered and # two years apart.
cycles = [ cycle + (cycle % 2) for cycle in candidate.get('election_years', []) ]
```

Let's run through this with an example with real values.  Let's say that `candidate['election_years']` is a list that contains these years: `[2013, 2014, 2015, 2016, 2017, 2018]`.  If the code above were to process this list, it would return the following: `[2014, 2014, 2016, 2016, 2018, 2018]`.  Try it out in the Python interpreter (`python` in your command line) to see for yourself!  You can use this code:

```python
# Sample dictionary to mimic application code and provide example data
candidate = {'election_years': [2013, 2014, 2015, 2016, 2017, 2018]}

# List comprehension to mimic application code - running this will print out the result, which will be a list
[ cycle + (cycle % 2) for cycle in candidate.get('election_years', []) ]
```

In running this you'll notice one other thing, though, which is that the output has multiple values that are identical.  We probably don't want that, so we can leverage another Python data structure to give us just the unique values:  the [`set`](https://docs.python.org/3/tutorial/datastructures.html#sets).  If you take a look at that documentation, you'll also see that Python also has a set comprehension, which works just like a list comprehension.  Let's adjust the code a bit more to leverage this:

```python
# Same code as above except for the syntax to leverage a set comprehension instead of a list comprehension
cycles = { cycle + (cycle % 2) for cycle in candidate.get('election_years', []) }
```

With the example values from above, this could return the values in any order - sets are *unordered*.  For example, it could return a set that looks like this:  `{2016, 2018, 2014}` (the curly braces around a list of values is how Python represents sets - not to be confused with dictionaries, which are key/value pairs!).  This isn't quite what we want either, I'm assuming we need these values in order.  Furthermore, a `set` isn't the data structure we ultimately want, we still need a `list`.

This is where `list()` _does_ come in handy.  We can convert the `set` into a `list`, which can be sorted.  If we take our set comprehension just a couple of steps further by converting it to a list, we should end up with what we want.  Here's how this all looks together:

```python
# Calculate and process all of our cycle values - the even number years
cycles = list({ cycle + (cycle % 2) for cycle in candidate.get('election_years', []) })

# Sort the cycles to be in order - sort is performed in place, meaning that the list is modified directly!
cycles.sort()
```

To review what just happened, we converted the `set` back to a `list` so we have just the unique values in the format we need for the application, and then we sorted them with the [`sort()`](https://docs.python.org/3/library/stdtypes.html#list.sort) method that is found on `list` objects.  The `sort()` method accepts an optional `key` parameter to control how the sorting is performed and the order desired, but by default it'll sort a list of numeric values in ascending order, which is typically how we process this information throughout the API and CMS.

Again, the final result we're probably looking for here is this:

```python
cycles = list({ cycle + (cycle % 2) for cycle in candidate.get('election_years', []) })
cycles.sort()
```

There's an argument to be made for abstracting out the cycle calculation logic to its own method as well, e.g., this:

```python
def get_cycle(year):
    """
    Returns the cycle end year for a given election year.
    """
    
    return year + (year % 2)
```

Or, as a fun exercise, as a [`lambda`](https://docs.python.org/3/tutorial/controlflow.html#lambda-expressions) expression:

```python
get_cycle = lambda year: year + (year % 2)

# Example usage - just like a function!
cycle = get_cycle(2019)

# Returns 2020
```

And using that in the code instead, as we could leverage the method elsewhere, but perhaps an exercise for another time. :-)

I know this is a long walk through, but I wanted to make sure this was all explained clearly and introduce folks to a few concepts that help make the code more resilient, concise, and Pythonic!

fec/data/views.py
```diff
@@ -231,7 +243,7 @@ def candidate(request, candidate_id):
         'party_full': candidate['party_full'],
         'incumbent_challenge_full': candidate['incumbent_challenge_full'],
         'election_year': election_year,
-        'election_years': candidate['election_years'],
+        'election_years': even_election_years,
```
ccostino: If you use the example given above, you'll need to change this variable name as well to `cycles`.

fec/data/views.py
ccostino: I realized I may have taken this a bit too far in looking back on the original code and wanted to be sure I called that out.  If we do want to preserve all of the years we have to be processed but convert all of the odd number years to even numbers, then the first part of the solution is all that is needed, the original list comprehension:

```python
cycles = [ cycle + (cycle % 2) for cycle in candidate.get('election_years', []) ]
```

And if that's the case, we likely don't have to sort it either.  In any case, I hope this was all useful to folks!

patphongs: ❤️ Thank you both @ccostino @lbeaufort for helping to review the python code here. The detailed explanation is super helpful and avoiding potential errors is a big plus! 

The only thing I would change about this, however, is instead of using `cycles` as the variable name, we keep it as `even_election_years`. 

```even_election_years = {cycle + (cycle % 2) for cycle in candidate.get('election_years', [])}```

`cycles` is a variable currently used within the views.py and an attribute in the API and not really applicable for candidates. The important distinction I have noticed between `cycles` and `election_years` is that cycles only returns 2 year chunks (also equivalent to an `fec_election_year` attribute, confusing I know!). Cycles is ideal for committees since we always display their data in 2 year chunks. But `election_years` is more appropriate for candidates as it returns a list of the exact years that a candidate is running from the candidate's form 2. Given that Presidential and Senate candidates have 4 and 6 year intervals as their election years, we need it to key off election_years rather than cycles. 

On the UX side of things, I think there was a lot of confusion and discussion as to what to call these time intervals. What was eventually decided is that for a candidate, cycles would be equivalent to "time period" and and election_years would be equivalent to "election". This is a recent change on the front end that hasn't really been adapted in our code yet. Just given the nature of how we use these different semantics can be confusing, and honestly, I think it's still in flux.

screen shot 2018-01-31 at 9 45 23 pm

Comment (ccostino) [reply fec/data/views.py]: Ah, thank you @patphongs for explaining the distinction!  I had completely forgotten about that and you're right, that's a major difference in time frames. :-)

File: fec/data/views.py (hunk 6 above)
Comment (ccostino) [reply fec/data/views.py]: Disregard per @patphongs note above!

**Comment by cptechiegal** [comment]
Thanks @ccostino , @lbeaufort  and @patphongs  for the good feedback from your code reviews. 
I will apply the changes suggested later today.

**Comment by cptechiegal** [comment]
@pat and @carlo - I have completed the changes suggested as follows:

- fec-cms/fec/data/views.py
1) Refactored  / Changed the creation of the Even Elections list code to include a Set construct to ensure the resulting List of even election years has no duplicates
2) Also used a Comprehension construct to create the list instead of explicit For statement which is more compact and has faster execution- ref: https://stackoverflow.com/questions/16341775/what-is-the-advantage-of-a-list-comprehension-over-a-for-loop#16341841
3) I did not need to include a Sort statement on the list since there is a sort performed in the UI cycle-select component ( Jinja macro ) prior to displaying the list.

-fec-cms/fec/fec/data/templates/macros/cycle-select.jinja
1) The indentation was corrected for the <select> section
2) No indentation change was made to the <option> section since it was already correct

These changes are ready for review and merge.

**Comment by cptechiegal** [comment]
@patphongs there is a refactoring task here that maybe we should create as a separate ticket: Refactor candidate-cycle-select macro to improve readability and maintainability. 

Refactoring the code to call candidate-cycle-select macro for all 5 financial webpages would be an improvement to code readability and future code maintenance.  
Currently, the election-lookup search webpage is sharing the cycle-select macro with 3 of the 5 candidate financial webpages. The election-lookup search webpage displays 2 year cycle periods whereas the 3 financial webpages are displaying single election years using the same cycle-select macro.  
Meanwhile the other 2 of the 5 candidate financial webpages (i.e., raising and spending) are using the candidate-cycle select macro to display single Election years.
 
Proposal to refactor as follows:
1) Update the candidate-cycle-select macro to handle durations as is done in the cycle-select macro.  Then this candidate-cycle-select macro will be more able to handle all 5 of the candidate financial webpages.
2) Update the 3 templates for the 3 financial webpages to that currently use cycle-select macro to use the updated candidate-cycle-select.

This work is technical debt that existed prior to this PR and can be addressed separately to allow this PR to move forward into production and resolve the immediate issue - issue 1644.

File: fec/data/views.py [hunk 7]
```diff
@@ -143,6 +145,11 @@ def candidate(request, candidate_id):
         'candidateID': candidate['candidate_id']
     }
 
+     # Addresses issue#1644 - make any odd year special election an even year 
+    #  for displaying elections for pulldown menu in Candidate pages
+    #  Using Set to ensure no duplicate years in list
+    even_election_years = { year + (year % 2) for year in candidate.get('election_years', []) }
```
ccostino: This is good that we're able to use the `set` here, but does it still need to be a `list` when all is said and done?  It may not if this is working as expected (a `set` is an iterable object in Python, just like a `list`), but for the sake of clarity and consistency, I'm wondering if we should explicitly make this a `list` anyway, which would require just wrapping the whole set comprehension with a call to `list()`, like so:

```python
list({ year + (year % 2) for year in candidate.get('election_years', []) })
```

cptechiegal: Hello @ccostino I can make a further change to cast the set to list if you prefer
but I did verify by testing that this is working now as a set.

patphongs: @ccostino just doing some reading about python list vs set. So in this scenario, we are casting even_election_years as a set, and what you're suggesting is to turn the set into a list. If the purpose of a set is to ensure no duplicate years, then transforming that set into a list is to just being explicit that we want a list? I guess I'm still unsure as to why we need to transform it to a list?  Does that make the code more efficient or reduce errors in the return? Thanks for helping to review this, learning a lot about python!

ccostino: @cptechiegal and @patphongs yes, with this change, `even_election_years` is now a `set` instead of what it originally was with the previous code, which was a `list`.

As @cptechiegal pointed out with testing, I'm not surprised it worked, as both data types are known as [iterables](https://docs.python.org/3.5/glossary.html#term-iterable) in Python (here is [documentation on the `set` class](https://docs.python.org/3.5/library/stdtypes.html#set) specifically), meaning that they are objects that can return their members one at a time (fun fact:  strings (`str`) in Python are also iterable objects!).  My main concern is just that the previous code was working with a `list`, so I'd caution against changing the data type if only to avoid unintended consequences.

In this case it's likely not an issue, but seeing as how it's data ultimately being returned to the client side, which is expecting a simple and likely mutable data structure serving as a collection of values, being explicit about that will help in avoiding any future gotchas.  Hence my suggestion to convert the `set` back to a `list` as a part of the processing.  So to answer your questions specifically, @patphongs,  no, it won't make the code any more efficient, but yes, it could reduce errors/unintended side effects, especially if down the road we need to do some additional processing with the data before returning it to the client. :-)

Furthermore, if you do find that you need to sort things on the server side in the application code, you'll need it to be a `list`.

As I noted above, all this involves is wrapping the set comprehension with a call to `list()` to do the conversation!

cptechiegal: @ccostino this has been updated to  List() as requested.

PR #1719, opened 2018-01-25
Feature/new banner alert

johnnyporkchops:
**What we're after:** Following the government shutdown, we received feedback that the banner announcing the agency's closure could not be seen. We have designed a new alert banner for situations in which greater prominence/attention is required. This issue will implement the design created in issue:  https://github.com/18F/fec-cms/issues/1717 

This PR adds a new template solely for the purpose of adding the special alert banner, keeping it away from the day-to-day publishing of usual banner announcements so no one is tempted to use it for inappropriate purposes. A preview template is added as well, similar to the banner announcements preview.

- Addresses # https://github.com/18F/fec-cms/issues/1717
**Completion criteria:**
- [x] Refine front-end (as seen in [prototype](https://user-images.githubusercontent.com/31663028/35339834-9d7cb4d4-00ef-11e8-82e8-4e4d943a1f7b.gif)) based on design here: https://github.com/18F/fec-cms/issues/1717
- [x] Incorporate logic to ensure this prominent banner remains at the top, in the event multiple banners are in use
- [x] Make the help-text warning more prominent and clear 
- [ ] Finalize prior to February 8, 2018

**Screenshot**
jan-24-2018 21-59-38


johnnyporkchops:
Here is a screenshot of the latest editing interface for this template. This seems to be the most control we have over help text for now,
editor

johnnyporkchops:
Here are updated screenshots of the  banner live and the editing interface with the new name for the template:
alone

**Always on the top:**
always_on_top

**Editor:**
editor

jenniferthibault:
Closing the loop on comments from standup, it looks like @JonellaCulmer 's feedback for changing the template name to `Alert for emergency use only` was integrated, so now this just needs technical review from @patphongs

fec/home/models.py
```diff
@@ -505,25 +505,59 @@ def get_author_office(self):
...
+            FieldPanel('link_url_2'),
+            FieldPanel('date_active'),
+            FieldPanel('date_inactive'),
+            FieldPanel('active'),
+            ],
+            heading="Home page banner announcement"
+        )
+    ]
+
+class AlertForEmergencyUseOnly(Page):
```
patphongs: Needs an extra blank line above this

fec/home/models.py
```diff
@@ -505,25 +505,59 @@ def get_author_office(self):
...
+            ],
+            heading="This 'alert for emergency use only' feature \
+            is used exclusively for an agency shutdown, or emergency \
+            event in which the agency as a whole cannot assist the \
+            regulated community or the public. The use of this feature \
+            requires approval by Amy Kort or Wei Luo prior to deployment."
+        )
     ]
 
 class CustomPage(Page):
```
patphongs: needs extra line above this class

fec/home/templatetags/home_page.py
```diff
@@ -20,19 +20,26 @@ def home_page_banner_announcement():
     eastern = timezone('America/New_York')
     datetime_now = eastern.localize(datetime.datetime.today())
     banners = HomePageBannerAnnouncement.objects.live().filter(active=True, date_active__lte=datetime_now, date_inactive__gt=datetime_now).order_by('-date_active')[:2]
+    alert_banners = AlertForEmergencyUseOnly.objects.live().filter(alert_active=True, alert_date_active__lte=datetime_now, alert_date_inactive__gt=datetime_now).order_by('-alert_date_active')[:2]
+
+
 
     return {
```
patphongs: too many blank lines here, need to delete 2 lines

fec/home/templatetags/home_page.py
```diff
@@ -20,19 +20,26 @@ def home_page_banner_announcement():
...
 
 
 # This is for the Wagtail preview for Home Page Banner Announcement
 @register.inclusion_tag('partials/draft-home-page-banner-announcement.html')
 def draft_home_page_banner_announcement():
     draft_banners = HomePageBannerAnnouncement.objects.all().order_by('-date_active')
+    draft_alert_banners = AlertForEmergencyUseOnly.objects.all().order_by('-alert_date_active')
+
 
     return {
```
patphongs: need to delete an extra blank line above

fec/home/templatetags/home_page.py
```diff
@@ -20,19 +20,26 @@ def home_page_banner_announcement():
...
-        'draft_banners': draft_banners
+        'draft_banners': draft_banners,
+        'draft_alert_banners': draft_alert_banners
     }
 
 @register.inclusion_tag('partials/home-page-news.html')
```
patphongs: need one extra blank line above

fec/home/templates/partials/home-page-banner-announcement.html
```diff
@@ -1,14 +1,31 @@
 {% load wagtailcore_tags %}
+{% if alert_banners %}
+
+<section class="slab slab--secondary message--red">
```
jenniferthibault: @johnnyporkchops @patphongs : Historically in this system we don't name classes with colors, since colors are variables. What are other alternatives here?

fec/home/templates/partials/home-page-banner-announcement.html
```diff
@@ -1,14 +1,31 @@
...
+  <div class="container">
+  {% for alert_banner in alert_banners %}
+    <div>
+      <div class="message message--red message--large message--info message--inline u-no-margin-top u-no-margin-bottom">
```
jenniferthibault: same question, re: naming `red message`

fec/home/templates/partials/home-page-banner-announcement.html
johnnyporkchops: @jenniferthibault, I noticed that too--I did not name this class, it already existed and we will have to see where else it is being used before changing it. I'll take a look.

jenniferthibault: Thanks so much!

johnnyporkchops:
@jenniferthibault , It looks like the  "red--message" class was added by Christine (or Noah) when the  original home page banner was made and only used there so renaming it would not be a problem. What would you suggest renaming it to?

fec/home/templates/partials/home-page-banner-announcement.html
johnnyporkchops: this is only used in the banner so renaming it is not a problem

jenniferthibault:
@johnnyporkchops I would love to hear your ideas on naming!

johnnyporkchops:
The color is variable is `$deep-red: #961a16` and also  `$secondary-focus: $deep-red`. It makes sense how these were originally intended to be used. 
So based on this similar existing use-case:
`.card__image__container--secondary {
    background-color: $secondary;
  }
`
..we would call it `.message-secondary-focus`. But the I shied away from this because it is not in focus state in a Section 508 sense of the word 'focus'. I had the same dilemma when when I recently created the lighter red boxes for the icon container:
` .grid__icon--deep-red {
  img {
    background-color: $deep-red;
  } `
Which should probably be changed too(along with its sister style: `.grid__icon--navy` ) because they also use a color in the classname).
@jen, @jonella, @patphongs . Maybe we can use ".message--emergency" for the deep-red style for non-focused items. What do you think.

jenniferthibault:
`.message--emergency` makes great sense to me, it describes the purpose of the banner it's used for instead of its color = 👍 👍

johnnyporkchops:
Thanks @jenniferthibault  `.message--emergency` it  is.

johnnyporkchops:
@jenniferthibault, I realized that we should no use "emergency" becuase the existing "message--red" is also used for the usual home page banner alerts which are not an "emergency." After thesaurasing the words 'alert' and 'prominent', the best I could come up with is "vivid". So I changed the classname to message--vivid.
Also @JonellaCulmer, do you think we should use the bang icon(exclamation point) instead of the info icon (i) for this? See screenshot:
alert_icon_instead_of_info_icon

JonellaCulmer:
@johnnyporkchops While my inclination is to change as few details for similarly functioning patterns/features, I agree with you. This banner stands apart from the other banner type and certainly exudes more ALERT than it does INFO. It's also used in only one particular scenario and so in that, it's special. Change away!

patphongs:
This PR has been closed in favor of https://github.com/18F/fec-cms/pull/1743. This change needs to be a hotfix due possible upcoming shutdown this week.

Issue #1782, opened 2018-02-13
Districts listed out of order on Election search results page - Dev space


patphongs:
When displaying the search results for a state, the districts are out of order. We need to take a look at what kind of sort order is in the API

**Example:** https://fec-dev-proxy.app.cloud.gov/data/elections/?cycle=2018&state=AZ

**API Call:** https://fec-dev-api.app.cloud.gov/v1/elections/search/?api_key=DEMO_KEY&per_page=100&cycle=2018&state=AZ

- [X] Front-end fix
- [x] API fix

lbeaufort:
It looks like it's related to the changes I made to use a new table with [this PR](https://github.com/18F/openFEC/commit/7a836fabcbe66a11c8f1ae5e43a28757d0fd7baa) 

Thank you for catching that, Pat. I could probably get a fix in before the release branch is cut, especially because Jun is cutting it in the afternoon. 

I should probably add an automated test for sort order - I think I was under the impression that the CMS sorted the house results by district.

patphongs:
Thanks @lbeaufort, I think it is a good idea to have the CMS handle the sort in the API call. Since the API already handles sorting districts, I made a PR (#1783) to just handle the sort in the form submit itself rather than having to change any of the backend code. I added you as a reviewer, let me know if you have any questions.

lbeaufort:
I'm removing the "high priority" label because we have a short-term workaround on the front end side that's already been implemented.

lbeaufort:
API fix merged: https://github.com/18F/openFEC/pull/2936

PR #1783, opened 2018-02-14
Fix district sort on election results page

patphongs:
## Summary

- Addresses #1782 
This adds the district sort to the election page search form

Example:
-  http://localhost:8000/data/elections/?cycle=2018&state=CA

### DIFF
--- a/fec/data/templates/election-lookup.jinja
+++ b/fec/data/templates/election-lookup.jinja
@@ -56,6 +56,7 @@
                 </select>
               </div>
               <div class="search-controls__submit">
+                <input type="hidden" value="district" name="sort" />
                 <button type="submit" class="button--search--text button--standard">Search</button>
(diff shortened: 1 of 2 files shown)


lbeaufort:
@patphongs this looks good to me - quick thinking! I also checked presidential years and they appear in order "P,S,H by district."

I'm thinking I should maybe also look into fixing the sort on the API side, and add an automated test - @LindsayYoung has worked with me on [this PR](https://github.com/18F/openFEC/commit/7a836fabcbe66a11c8f1ae5e43a28757d0fd7baa) that I believe scrambled the districts. Lindsay - thoughts?

lbeaufort:
@patphongs I checked with @LindsayYoung and we can't anticipate any problems with addressing this sort issue from both sides. Thank you to you both!

PR #1789, opened 2018-02-14
fixed scss for numbered lists and alternating-bg cards

johnnyporkchops:
Addresses:
add numbers to `list--numbered class` : https://github.com/18F/fec-cms/issues/1781
add transparent border to `.card--alternating-bg` : https://github.com/18F/fec-cms/issues/1734 
### **Numbered lists:**
numbered_lists

### **Alternating-bg cards before/after**
**Before:**
jump

**After:**
no_jump

### DIFF
--- a/fec/fec/static/scss/components/_cards.scss
+++ b/fec/fec/static/scss/components/_cards.scss
@@ -212,6 +212,7 @@
 .card--alternating-bg {
   min-height: u(8.3rem);
   padding: 0;
+  border: 2px solid transparent;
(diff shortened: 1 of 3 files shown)


johnnyporkchops:
Added the same styles for numbered lists to apply to html-blocks as well as rich-text blocks. That was the missing piece making testing this so confusing!

johnnyporkchops:
See the even more offensive example of jumping in upadted screenshot above

PR #1823, opened 2018-03-01
update moment package

pkfec:
Updated the fullcalendar dependency package **` moment`** from 2.17.1  to 2.19.3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4774,7 +4774,7 @@
       "integrity": "sha1-6NRY1kt9y1unDOiQ3aJIi4kRU5o=",
       "requires": {
         "jquery": "3.2.1",
-        "moment": "2.17.1"
+        "moment": "2.19.3"
       }
     },
     "function-bind": {
@@ -5604,9 +5604,9 @@
       }
     },
     "handlebars": {
-      "version": "4.0.10",
-      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.0.10.tgz",
-      "integrity": "sha1-PTDHGLCaPZbyPqTMH0A8TTup/08=",
+      "version": "4.0.11",
+      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.0.11.tgz",
+      "integrity": "sha1-Ywo13+ApS8KB7a5v/F0yn8eYLcw=",
       "dev": true,
       "requires": {
         "async": "1.5.2",
@@ -6412,7 +6412,7 @@
         "escodegen": "1.7.1",
         "esprima": "2.5.0",
         "fileset": "0.2.1",
-        "handlebars": "4.0.10",
+        "handlebars": "4.0.11",
         "js-yaml": "3.9.1",
         "mkdirp": "0.5.1",
         "nopt": "3.0.6",
@@ -6878,7 +6878,7 @@
             "escodegen": "1.8.1",
             "esprima": "2.7.3",
             "glob": "5.0.15",
-            "handlebars": "4.0.10",
+            "handlebars": "4.0.11",
             "js-yaml": "3.9.1",
             "mkdirp": "0.5.1",
             "nopt": "3.0.6",
@@ -7923,9 +7923,9 @@
       }
     },
     "moment": {
-      "version": "2.17.1",
-      "resolved": "https://registry.npmjs.org/moment/-/moment-2.17.1.tgz",
-      "integrity": "sha1-/tlQYGPzaxDwZsi1mhRNf66+HYI="
+      "version": "2.19.3",
+      "resolved": "https://registry.npmjs.org/moment/-/moment-2.19.3.tgz",
+      "integrity": "sha1-vbmdJw1tf9p4zA+6zoVeJ/59pp8="
     },
     "ms": {
       "version": "2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-z6I7xYQPkQTOMqZedNt+epdLvuE=",
       "dev": true,
       "requires": {
-        "acorn": "5.1.1",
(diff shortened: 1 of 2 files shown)


pkfec:
@lbeaufort Please review.


cc : @ccostino

pkfec:
got it. so we should always update package.json to install or update latest packages and or dependencies. Thanks for clarifying.

ccostino:
Yup, you got it! :-)

fec-jli:
**1) For fullcanlendar package**
fullcanlendar repo: [https://github.com/fullcalendar/fullcalendar]
the latest version is V3.8.2 (release date 01/31/2018)
our current version is V3.3.1 (release date 04/01/2017)

fullcanlendar release notes: [https://github.com/fullcalendar/fullcalendar/releases]
since fullcanlendar V3.6.0 release, you will see:
**Incompatibilities:**

> •The viewRender callback might now be fired AFTER events have been rendered to the DOM. However, the eventRender/eventAfterRender/eventAfterAllRender callbacks will always be fired after viewRender, just as before.
> •The internal Grid class (accessed via $.fullCalendar.Grid) has been removed. For monkeypatching, use DayGrid/TimeGrid directly.

in our repo, only calendar.js using this package, we use Grid class, calendar not display and get this error:
`uncaught typeError cannot read property 'prototype' of undefined`


after try to change Grid to DayGrid, calendar displays, but get some other errors.
`uncaught typeError cannot read property 'url' of undefined`


a)This ticket is not easy to modify canlendar.js to compatible the newer version in short time.
b)need javascript expert work on this.
c)so we keep the current version 3.3.1.  the last version support our code is v3.4.0


**2)For moment.js**
the latest version is 2.20.1
we upgrade to 2.20.1 and test on most pages using moment,js as possible as I can.


**3)in fullcalendar package, its dependencies like this:**
"dependencies": {
    "jquery": "2 - 3",
    "moment": "^2.9.0"
  },
so it will support the moment version 2.x.x. the cap is our moment version in our dependencies.


**4)For Gemnasium**
 [https://gemnasium.com/github.com/18F/fec-cms]
you will see the latest moment is 2.20.1, so we should be good to pass Gemnasium.

pkfec:
@ccostino and @lbeaufort  : Please review.

patphongs:
We should open a separate ticket to debug JS in calendar so that we can upgrade full calendar. This PR while it does upgrade our moment everywhere else, it doesn’t for the calendar

pkfec:
I did try to upgrade the FC v ^3.4.1 and moment V2.19.1 and V2.20.1  and tested Calendar and other data pages in my local. They all seemed to work OK. The FC 3.4.1 has some minor updates, otherwise same as the current version 3.3.1 that we are using. 

so far it looks like the FC ^3.4.1 is the stable version we can use in our apps. However, if we want to update FC to the latest version,  front end expertise need to dig deep into the libraries and fix all the JS errors.

fec-jli:
@patphongs yes, this ticket is only for fix Gemnasium alert about **moment** version .
until now, Gemnasium no alert about **fullcalendar** version.

I agree with create another ticket for upgrade **fullcalendar** package.   Thanks

lbeaufort:
@fec-jli @patphongs @pkfec There isn't a Gemnasium alert for our `moment` version - we're at 2.19.3 now.

Gemnasium alert: 
REGULAR EXPRESSION DENIAL OF SERVICE
Moment is vulnerable to a low severity regular expression denial of service vulnerability.

Affected versions: Pior to 2.19.3
Fixed versions: **2.19.3**

fec-jli:
@lbeaufort 
see in package-lock.json,  the moment version is still 2.17.1, that is Gemnasium alert about. Gemnasium only check the lock version.
[https://github.com/18F/fec-cms/blob/master/package-lock.json](https://github.com/18F/fec-cms/blob/master/package-lock.json)

I think the alert is not because fullcalendar package. in fullcalendary's dependance: you can see, moment: ^2.9.0 , the first 'caret` means it can support 2.x.x.  it will same as our repo moment version.

for some reasons, our package.json, the moment is v2.19.1, but package-lock.json still v2.17.1.
looks like node_module didn't rebuild from clean space.

one more thinking, even we upgrade to 2.19.1 , it is still red color in Gemnasium, that means we will get alert soon. so I prefer upgrade moment to v2.21.1 (today's version) or ^2.21.1(I am not 100% sure for using `caret` here is safe or not). I have tested v2.20.1.this on local most of cms pages.

fec-jli:
when we upgrade our moment version to 2.20.1, after deploy correctly, the moment version inside fullcalendar will become 2.20.1 too.

fec-jli:
It looks like don’t create the new package-lock.json during deployment every time. you can check `develop` and `master`,  the moment.js version is 2.19.3 in package.json, but in package-lock.json , still be v2.17.1.    Can we remove old package-lock.json file before deploy? or any other better way to modify circle config or ignore file to handle this?  I did this way on my local : 1)remove node_modules folder  2)remove old package-lock.json.  3)modify package.json. 4)`npm install`,  the new package-lock.json will be created on my local. everything works fine. I have tried moment.js v2.20.1 on local..

ccostino:
Ah, is the updated version of `package-lock.json` committed in this PR along with the `package.json` file changes?  With that file being present, it wouldn't be recreated on deploy as it is used as the source of truth for the `npm install` process.

pkfec:
updated the moment package to V2.20.1  in package.json and also pushed the package-lock.json file upstream.

@ccostino  Please review.

ccostino:
I'm sorry, my initial comment didn't fully explain that you still had to commit both files.  In re-reading the message I left it's not clear.  Thanks for raising the flag and asking about that, @fec-jli!

fec-jli:
Thanks Carlo, Laura and Pat.
 We do learn a lot from this ticket.

Issue #1391, opened 2017-10-26
Limit individual contributor zip search to 5 digits


lbeaufort:
So that users can retrieve accurate contributor data based off zipcode, only the first 5 digits should be matched on the [Individual Contributor Search](https://www.fec.gov/data/receipts/individual-contributions/?two_year_transaction_period=2018&min_date=01%2F01%2F2017&max_date=12%2F31%2F2018).  Currently, searching 5-digit zip excludes contributors with 9-digit zips, and searching 9-digit zips only returns data where the filer has provided the zip+4. 

**Email from data user:**
_"I’m trying to do a search by zip code and would like it to match both 5 and 9 digit entries.
 
For example:
•         If I pass in the contributor name of “frank” and “10050” for– I get 9 records back [link]( https://www.fec.gov/data/receipts/individual-contributions/?two_year_transaction_period=2018&contributor_name=frank&min_date=01%2F01%2F2017&max_date=12%2F31%2F2018&contributor_zip=11050) 
•         However, if I add the zip code of 110502814 – I get an extra record for a total of 10  [link](https://www.fec.gov/data/receipts/individual-contributions/?two_year_transaction_period=2018&contributor_name=frank&min_date=01%2F01%2F2017&max_date=12%2F31%2F2018&contributor_zip=110502814&contributor_zip=11050)
 
In this example, all that I have on my end is the 11050 zip code and I would like to use that information to get back all 10 records."_ 

Two proposed changes:
**CMS**
The zip code on this page should be limited to 5 characters

**API**
The zip search should match on the left(5) of the zip. 

Thanks to Jason in Public Records for bringing this to my attention.

- [ ] limit front-end input to 5 digits

PaulClark2:
What's the status of this issue?

patphongs:
@PaulClark2 Put this back on my plate to do for 5.2 since this got missed the previous time. @lbeaufort Is working on an API fix and I will add a front-end fix to not allow zip codes greater than 5 digits.

patphongs:
This issue is done for the front-end, however, still waiting on https://github.com/fecgov/openFEC/pull/3003 back-end PR to be implemented in order to merge this PR on the front-end

AmyKort:
Is this ready to merge?

lbeaufort:
@AmyKort thanks to @hcaofec and @patphongs this issue is complete!

PR #1840, opened 2018-03-07
[DON'T MERGE YET] Limit individual contributor zip input to 5 digits

patphongs:
## Summary

- Addresses #1391 
Limits zip code input to 5 digits for the individual contributor search http://localhost:8000/data/receipts/individual-contributions/

-  Individual contributor search

### DIFF
--- a/fec/data/templates/partials/individual-contributions-filter.jinja
+++ b/fec/data/templates/partials/individual-contributions-filter.jinja
@@ -22,7 +22,7 @@ Filter individual contributions
   <div class="accordion__content">
     {{ text.field('contributor_city', 'City') }}
     {{ states.field('contributor_state') }}
-    {{ text.field('contributor_zip', 'ZIP code')}}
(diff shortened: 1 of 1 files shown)


lbeaufort:
@patphongs I needed to restart my browser to see the changes - looks good to me! Would it be hard to add a validation message when people try to input more than 5 digits? To me, it seems like we could implement "as-is" and see if we get feedback about it. 

Also - let's make sure we don't merge this until we know the [corresponding API work(https://github.com/18F/openFEC/issues/2980) can make it into this release (5.2).

patphongs:
> Would it be hard to add a validation message when people try to input more than 5 digits? To me, it seems like we could implement "as-is" and see if we get feedback about it.

I think we can test to see if anyone is confused or complains about this. In my opinion, zip codes being 5 digits is fairly common. If this becomes a problem, we can certainly throw some kind of message and additional validation on the field.

lbeaufort:
@patphongs - I tested locally and this looks good with the API change.

Issue #1803, opened 2018-02-20
Calendar filters not working on homepage


AmyKort:
The "all filing deadlines," "all outreach" and "all legal events" filters on the homepage take users to all calendar events, rather than filtering for the specific type of event.

![image](https://user-images.githubusercontent.com/13237823/36430556-69a5eb3c-1623-11e8-8c04-82596d1f5541.png)

patphongs:
This fix has been merged in the PR #1843. Closing...

PR #1843, opened 2018-03-08
[WIP] Fixed events and deadlines filter buttons on homepage

patphongs:
## Summary

- Addresses #1803 
Calendar filter buttons on the homepage were calling old API codes. These changes update to the correct API codes.

-  Homepage: Events and deadlines section

### DIFF
--- a/fec/home/templates/home/home_page.html
+++ b/fec/home/templates/home/home_page.html
@@ -38,17 +38,17 @@
         <li class="grid__item grid__item--with-button">
           <div class="t-sans t-bold">Next filing deadline:</div>
           <div class="js-next-filing-deadline"></div>
-          <a class="button button--alt button--calendar" href="/calendar/?category=report-E&category=report-M&category=report-MY&category=report-Q&category=report-YE">All filing deadlines</a>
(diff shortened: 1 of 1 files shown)


AmyKort:
@llienfec is probably a better person to answer your question, @patphongs . But it looks to me like we'd need to capture year end and mid year reports in the "all filings" filter.

rhoughfec:
@AmyKort @patphongs yes, it needs to include Mid-Year and Year-End reports as well when we do "all filings." Thanks!

patphongs:
Just investigated and it appears that the year-end report does show up in the API call, however, the mid-year report does not. We need to see how that is coded into the database. The `calendar_category_id=21` should be pulling up the mid-year and the year-ends. But it's not. 

@lbeaufort Do you have time to discuss this with me? I'm not sure how the data gets transformed in our database tables based on what is inserted into our database.

Here's the call which should return 2017 and 2018 reporting deadlines. https://fec-dev-api.app.cloud.gov/v1/calendar-dates/?api_key=DEMO_KEY&per_page=500&calendar_category_id=21&calendar_category_id=25&calendar_category_id=26&calendar_category_id=27&min_start_date=2017-01-01&max_start_date=2018-12-31&_=1520900912793

lbeaufort:
@patphongs  - per our conversation, can you add category 21 to the "all reporting deadlines" link?

Issue #1839, opened 2018-03-07
USAJobs API not returning 'WhoMayApply'


johnnyporkchops:
**Response from vendor_help@usajobs.gov regarding the first posting that did not return WhoMayApply:**
_This job is being posted with GOVT_WhoMayApply set to 15511.
We have looked into this and have confirmed it is a bug that we will fix in a future release of USAJOBS._
_Also a note on WhoMayApply. In the coming months we will be updating the documentation on developer.usajobs.gov to include information on the new HiringPath field that will be replacing the WhoMayApply field in future._

I will continue to followup on this and update issue with any developments. We might want to  consider removing this field from the /careers page:
https://www.fec.gov/about/careers/

(Also inquiring into wether the six digit code mentioned in the response may be an interim way the determine the WhoMayApply value)

rhoughfec:
Thanks for flagging this @johnnyporkchops I had noticed this the other day. Looping in @AmyKort so we can determine if we need to take action or just wait for the bug fix from USAJOBS.

AmyKort:
I would suggest removing the field from the page.

johnnyporkchops:
I will map hiring path abbreviation(s) listed for each job to the corresponding description available at https://data.usajobs.gov/api/codelist/hiringpaths and render to a field titled `Open to` on the /Carreers page

AmyKort:
Merged,so closing.

PR #1885, opened 2018-03-25
Careers page - USAJobs API- Replace WhoMayApply with HiringPath description(s), refactor python

johnnyporkchops:
**Addresses Issue:** [1839 - USAJobs API not returning 'WhoMayApply'](https://github.com/fecgov/fec-cms/issues/1839 )
**Affected areas of application:** [/about/careers](https://www.fec.gov/about/careers/)

USAJObs is no longer showing WhoMayApply as an endpoint parameter. Instead, the HiringPath parameter can be used. 

(Also paired with @ccostino and @lbeaufort to refactor python and Update User Provided Service (UUPS) with new (non-public) USAJobs API Key.)

The hiring path(s) listed for each job are abbreviations, and not very descriptive or helpful. So in addition to querying the api to search open FEC Jobs, I am also querying the hiring paths endpoint to get a list of codes and their associated descriptive text which I can then use to map the hiring path(s) for each job to the associated descriptions.

job item hiring path:
``` 
HiringPath": [
                      "fed-competitive",
                      "fed-excepted",
                       "vet"
                            ] 
```

Subset of data returned from https://data.usajobs.gov/api/codelist/hiringpaths:

```CodeList": [
        {
            "ValidValue": [
                {
                    "Code": "DISABILITY",
                    "Value": "Individuals with disabilities",
                    "LastModified": "2017-06-30T08:16:59.55",
                    "IsDisabled": "No"
                },
                {
                    "Code": "FED",
                    "Value": "Federal employees",
                    "LastModified": "2018-03-08T17:02:05.403",
                    "IsDisabled": "Yes"
                },
                {page- Addresses # [_Issue number_]
(. . . etc.)
```

new_careers


fec/home/templatetags/open_jobs.py
```diff
@@ -1,43 +1,71 @@
...
+        for result in search_results.get('SearchResultItems', None):
+            matched_object_descriptor = result.get('MatchedObjectDescriptor', '')
+
+            jobs_dict = {
+                'position_title': matched_object_descriptor.get('PositionTitle', ''),
+                'position_id': matched_object_descriptor.get('PositionID', ''),
+                'position_uri': matched_object_descriptor.get('PositionURI', ''),
+                'position_start_date': dateutil.parser.parse(matched_object_descriptor.get('PositionStartDate', '')),
+                'position_end_date': dateutil.parser.parse(matched_object_descriptor.get('PositionEndDate', '')),
+                'job_grade': matched_object_descriptor.get('JobGrade', {})[0].get('Code', ''),
```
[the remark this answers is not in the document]
johnnyporkchops: @ccostino , I took a stab at this, but its definitely not correct. I am not sure what should the resulting action be if the length of the list is more than 1?

fec/home/templatetags/open_jobs.py
```diff
@@ -1,43 +1,76 @@
...
+    search_results = responses.get('SearchResult', {})
+
+    #iterate over returned job data
+    if 'SearchResultItems' in search_results:
+        for result in search_results.get('SearchResultItems', None):
+            matched_object_descriptor = result.get('MatchedObjectDescriptor', {})
+            if matched_object_descriptor.get('JobGrade')[0]:
+                job_grade = matched_object_descriptor.get('JobGrade', [])[0].get('Code', '')
+            else:
+                job_grade = ''
```
ccostino: Excellent, @johnnyporkchops!   This is the way to account for this and use the value below when building the dictionary for output. :-)

The `if` statement still needs a bit of work here as I could still see a situation where it'd just throw an error, but you're definitely on the right track!  Here's what I'd do here to account for this:

```python
if len(matched_object_descriptor.get('JobGrade', [])) > 0:
```

Unpacking this from the inside out, this adds a default value to return if the key doesn't exist so we always deal with a list, which is an iterable object.  The [`len` function](https://docs.python.org/3/library/functions.html#len) in Python operates on any object that is an iterable type or supports counting in some fashion.  Since we're dealing with a list, we can just check to see how many elements are in it and if it's greater than 0, then we know we can access the first element as you do in the line below.

Does this make sense?  Let me know if you have any questions!

johnnyporkchops: @ccostino Am I correct in returning returning `'' ` with the else statement?

johnnyporkchops:
@ccostino Oh, I guess `get('Code', '') ` returns `''` so I do not need else statement(?)

fec/home/templatetags/open_jobs.py
ccostino: Yes, that's correct, and to answer your other question, you do still need this because of what happens in the case that the initial `if` statement evaluates to `False`, meaning it can't find any list elements.

Alternatively you could just set `job_grade = ''` above the `if` statement and then it'd be overwritten with the value (if one is found), in which case you wouldn't need the `else`, but that becomes personal preference.

johnnyporkchops: @ccostino , thanks for your comments. I tested this by removing the `else` condition and feeding  dummy json in with `"JobGrade":[]` and it errors, but with the else statement, it does not. Now it makes more sense. 
Another thing I noticed: It looks like when there is no value for a key in an object, this API shows an empty string, for example: 
```  
"WhoMayApply": {
      "Name": "",
      "Code": ""
        },
 ```
which I  assume could be considered another safety net against errors, and assume is why my old code without the default values did not break when they removed these values. Just curious, Is that correct, and is that something the API designer must specify?

Comment (ccostino) [reply fec/home/templatetags/open_jobs.py]: Ah, that is good, so they at least provide the field even if there's no value available.  This helps promote and maintain consistency and is something that an API architect would have to specify though it is considered a best practice.  This way a consumer of the API (another developer using the API, such as yourself) can rely on the documentation to always know what's going to returned by an endpoint and what the default values are, or what is returned in the case of no value being available.

Comment (ccostino) [reply fec/home/templatetags/open_jobs.py]: Furthermore, this being the case with the consistency in the API means we may not need to be as defensive in our programming as I've guided here, but that is also still a good practice on the consumer side of an API; you never know when a new release might change or break something, so you want your code to be ready for as much as possible! :-)

**Comment by johnnyporkchops** [comment]
@lbeaufort , writing automated tests is a great idea, especially since we want to know as soon as possible if a change on USAJobs end causes an issue on our end. Also the point of this PR is to make this code live up to the standards of the rest of the application, so tests would be the next logical step, had we been writing code for the API.

File: fec/home/templatetags/open_jobs.py [hunk 3]
```diff
@@ -1,43 +1,73 @@
...
-        'authorization-key': settings.USAJOBS_API_KEY,  
-        'user-agent': "jcarroll@fec.gov",
-        'host': "data.usajobs.gov",
-        'cache-control': "no-cache",
-        }
-    
-    response = requests.request("GET", url, headers=headers, params=querystring)
+        'authorization-key': settings.USAJOBS_API_KEY,
+        'user-agent': 'jcarroll@fec.gov',
+        'host': 'data.usajobs.gov',
```
ccostino: This isn't a security incident, but I think these two header values should also come from the user provided credentials that can be loaded into the settings as well, like the API key.  The reason I suggest this is it reduces the attack surface of someone trying to either get the key or spoof using the account.

fec/home/templatetags/open_jobs.py
```diff
@@ -1,43 +1,73 @@
...
+                'position_id': matched_object_descriptor.get('PositionID', ''),
+                'position_uri': matched_object_descriptor.get('PositionURI', ''),
+                'position_start_date': dateutil.parser.parse(matched_object_descriptor.get('PositionStartDate', '')),
+                'position_end_date': dateutil.parser.parse(matched_object_descriptor.get('PositionEndDate', '')),
+                'job_grade' : job_grade,
+                'low_grade': matched_object_descriptor.get('UserArea', {}).get('Details', {}).get('LowGrade', ''),
+                'high_grade': matched_object_descriptor.get('UserArea', {}).get('Details', {}).get('HighGrade', '')
+            }
+            #map hiring-path code(s) for each job to description(s)
+            hiring_path_codes = codes_responses.get('CodeList', [])[0].get('ValidValue')
```
ccostino: One last change here:  Just like above with the check for the `job_grade` above, I'd do an identical check on the list length here before just grabbing the index directly to make sure that there's an element to access.

ccostino: I'd also make sure that the last piece returns an empty list by default: `get('ValidValue', [])`

This way the list comprehension below making use of `hiring_path_codes` won't ever fail either.

ccostino:
Some tests around this would be good, @lbeaufort, good idea on creating an issue to capture that work so we can get to it in the future. :-)

fec/home/templatetags/open_jobs.py
```diff
@@ -1,43 +1,71 @@
...
+                'low_grade': matched_object_descriptor.get('UserArea', {}).get('Details', {}).get('LowGrade', ''),
+                'high_grade': matched_object_descriptor.get('UserArea', {}).get('Details', {}).get('HighGrade', '')
+            }
+            #map hiring-path code(s) for each job to description(s)
+            hiring_path_codes = codes_responses['CodeList'][0]['ValidValue']
 
+            hiring_path = [item for item in result['MatchedObjectDescriptor']['UserArea']['Details']['HiringPath']]
+            hp = []
+            for path in hiring_path:
+                hpa = [item for item in hiring_path_codes if item['Code'] == path.upper()]
```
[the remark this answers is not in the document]
johnnyporkchops: @ccostino , I made some changes here but I am not sure it completely ensures against error

johnnyporkchops:
@ccostino , I have not put in variables for the other sensitive info yet, but I first wanted to see if I am on the right track for your [last review changes](https://github.com/fecgov/fec-cms/pull/1885/files/a8aa9d1392257463050d8bf1e3c052c7f462740f#r179261852). I think there is still a possibility of error with the `hiring_path_codes` , but as not sure how to approach it.

fec/home/templatetags/open_jobs.py
```diff
@@ -1,43 +1,76 @@
...
+                'position_end_date': dateutil.parser.parse(matched_object_descriptor.get('PositionEndDate', '')),
+                'job_grade' : job_grade,
+                'low_grade': matched_object_descriptor.get('UserArea', {}).get('Details', {}).get('LowGrade', ''),
+                'high_grade': matched_object_descriptor.get('UserArea', {}).get('Details', {}).get('HighGrade', '')
+            }
+            #map hiring-path code(s) for each job to description(s)
+            if len(codes_responses.get('CodeList', [])) > 0:
+                hiring_path_codes = codes_responses.get('CodeList', [])[0].get('ValidValue', [])
+            else:
+                hiring_path_codes = ''
```
ccostino: @johnnyporkchops this is very close!  The only thing that has to change is this last piece here in the `else` clause.  You'll want `hiring_path_codes = []` instead, because you're basing a list comprehension off of `hiring_path_codes` just below, which will expect a list object, not a string object.

(It may still work, because [a string is a sequence type in Python](https://docs.python.org/3/library/stdtypes.html#textseq), but it's not the behavior we're going for here.)

johnnyporkchops:
Thanks @ccostino , good point about creating tests while its still fresh in our minds.

PR #1889, opened 2018-03-28
Add primary alternating bg card styles

jenniferthibault:
## Summary

- Addresses pattern library issue https://github.com/fecgov/fec-pattern-library/issues/71
This adds the classes necessary to flesh out all of the card patterns in both primary and secondary color combinations, and makes them ready for future use. It does this by adding a `primary` colored variation to the `.card--alternating-bg` class

It also adds a few ready-made icons to demonstrate the correct color pairing of icons/backgrounds for the cards. 

-  This doesn't impact any styles in use on production, it only adds classes for the future.
- It will also enable patterns in the pattern library to display correctly 

## Screenshots

- Creates the following style:
screen shot 2018-03-28 at 12 08 36 am

## Related PRs

branch | PR
------ | ------
Pattern library: feature/card-horizontal | [Create horizontal card subcategory 96](https://github.com/fecgov/fec-pattern-library/pull/96)

____

Since this is a style adjustment that doesn't impact anything on production, it would be AWESOME to get this merged into develop before tomorrow's release branch is cut so that the pattern library work stays unblocked


fec/fec/static/scss/components/_cards.scss
```diff
@@ -226,6 +226,11 @@
     vertical-align: middle;
   }
 
+  .card__image__container--primary {
+    background-color: $primary;
+    border-radius: 4px 0 0 4px;
+  }
+
```
patphongs: Not sure if these additional style classes are needed: `.card__image__container--primary` and `.card__image__container--secondary`. Do you know of any other instance in which it's used? `.card--secondary` and `.card--primary` classes seem to set the background color already and the border-radius is also taken care of by the `.card` class. I think all you really need is just the new SVG images right?

jenniferthibault: @patphongs it's my understanding that they're for the cards that look like this specifically, where the full background is only half of the card:
screen shot 2018-03-28 at 12 37 38 pm

You can see an example of one live on this page: https://www.fec.gov/help-candidates-and-committees/candidate-taking-receipts/contribution-types/

patphongs: Ah ok! Yes that makes sense then, I was concentrating on the wrong card type. The background setting makes more sense now. I would still remove the `border-radius: 4px 0 0 4px;` on both the `.card__image__container--secondary` and `.card__image__container--primary` classes, as those are taken care of by the `.card` class already.

patphongs: AND SCRATCH THAT! I just saw that the border radius is setting it on the background to the left. This looks good to go!

Issue #1521, opened 2017-11-28
Committee filing filters not working on Internet Explorer


AmyKort:
When trying to narrow down this committee filings by Form Type (i.e. PAC and Party financial statements Form 3X) or Years (i.e. 2015-2016), users are  unable to select either, or any option for that matter, after clicking on the arrow for the menu options. This is only happening in IE.

IE version 11 (test windows 10 as well)

* URL: https://www.fec.gov/data/filings/?data_type=processed&amp;committee_id=C00001313

johnnyporkchops:
I think this but has something to do with IE11/Edge support for `querySelectorAll` and `matches()`. We seem to have polyfills or fixes for this built into our codebase that is transpiled to the cached js files, but for some reason they are still throwing errors. I will continue to dig into to JS directories to figure this out. 

This error goes away when emulating IE 10 via the developer console (function > F12 > emulaton
In the meantime, this could be fixed by emulating IE10 instead of Edge in the IE compatibility meta-tag on /meta-tags.jinja:
` <meta http-equiv="X-UA-Compatible" content="IE=EmulateIE10"> `
Hopefully future releases of IE Edge and IE, will fix this. (though IE is notorious for fixing one bug and creating another when it comes to browser versions.)

Does anyone think emulating IE10, just on /data templates is problematic? It is not ideal, but may be a viable interim solution until wither IE or fixes thsi problem or we can find and fix it in our code.

Anyone who has any input or ideas on how to fix the code so it works without setting the X-UA meta tag, please chime in.

johnnyporkchops:
Thanks @xtine .I think it does already use jQuery. Question: Is jQuery code transpiled into vanilla JS via Babel or something along those lines?

xtine:
What I mean, is that the jquery selector method `$('.selector')` is equivalent to `document.querySelectorAll('.selector')`. Except, jQuery has backwards compatible support for legacy and outlier browsers.

Babel is only needed to compile ES6 code to ES5 (standard JavaScript). The only place that uses ES6 is the legal section that utilizes React. If you look at the Webpack configuration, we are just combining and minifying JavaScript files.

johnnyporkchops:
Clicking the dropdown in IE debug mode triggers the `querySelectorAll` error. That error persists even when the dropdown works properly in IE10 emulation mode. I was unable to determine how, if at all,  that error might actually cause the dropdown problem in IE 11/Edge.

AmyKort:
[...]

PR #1901, opened 2018-04-03
Fix datatable checkbox dropdown bug in IE

johnnyporkchops:
**Addresses:**
Committee filing filters not working on Internet Explorer #1521
**Areas of application affected:** 
all datatables
This is actually a bug across AA datatables where ,in Internet Explorer (IE), clicking on a dropdown item causes the dropdown to close and no item to be chosen. This is a bug that is caused by the javascript event created to close the menu when the user tabs (focuses) away from the dropdown and into the document body.
In order for this event to be fired recognized when tabbing (or focusing) away from the dropdown item, we must use [focusin](https://developer.mozilla.org/en-US/docs/Web/Events/focusin) instead of [focus](https://developer.mozilla.org/en-US/docs/Web/Events/focus) because the former [bubbles](https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles) and the latter does not.
IE11 handles this bubbling improperly and causes a `focusin` event on the document (outside the dropdown) upon clicking anywhere inside the dropdown. The menu closes and no item is ever chosen. 
**Two ways to fix:**
1) To fix this I have removed bubbling on event handler for this using `stopImmediatePropigation()`.Keyboard actions still work on this item including tabbing, enter and escape. The only caveat is that the dropdown will not close automatically when one tabs past the last item in the list. (enter, excape, or a click outside the dropdown does close it) But is does tab to the next filter item after the dropdown, so users with screenreaders will not be affected.
2) The other way to fix this without this small functionality hit would be to downgrade the meta tage to emulate IE 10. This works. I would like to hear others' thoughts on which of the two seems best, This PR is setup for the first option.
**Also, once this is decided, we may want to  change this to a hotfix because this is a fairly obtrusive bug**

### DIFF
--- a/fec/fec/static/js/modules/dropdowns.js
+++ b/fec/fec/static/js/modules/dropdowns.js
@@ -90,6 +90,7 @@ Dropdown.prototype.handleFocusAway = function(e) {
       !this.$panel.is($target) && !$target.is(this.$button)) {
     this.hide();
   }
+  e.stopImmediatePropagation()
 };
 
 Dropdown.prototype.handleKeyup = function(e) {


johnnyporkchops:
removing the `stopImmediatePropigation()` and changing `focusin` to `focus` makes this pass the JS test that it was failing before, but I would be remiss in not pointing out that the same functionality still is missing---closing the dropdown when tabbing past the final choice.The dropdown cans still be closed by escape, enter, or clicking away, and the next tab-indexed item does gain focus as it should.

johnnyporkchops:
So it was decided that emulating IE10 with the meta tag was the least obtrusive solution. So that is what this PR changes.

PR #1883, opened 2018-03-23
Feature/new audit search page

johnnyporkchops:
This migrates the [classic audit search application](https://classic.fec.gov/auditsearch/auditsearch.do) from fec classic to the new fec website
This is the culmination of several PRs and issues listed below:
**PR:** 
[WIP]-Datatable page for audit cases #1381
**Issues:**
[fec-cms] set default sub_category_id=all for audit-case endpoint #1882
Audit search styling and tags #1777
Build audit search page #1304
Review and Prioritize Feedback Received from Audit Team #1724

Issue #1882, opened 2018-03-23
[fec-cms]  set default sub_category_id=all for audit-case endpoint


johnnyporkchops:
@fec-jli commented on [Tue Mar 20 2018](https://github.com/fecgov/openFEC/issues/3016)

when missing sub_category_id, reset sub_category_id=all.

johnnyporkchops:
Since we were changing the sub category select to blank html() on-change, setting to value to all did not work because you cannot set the value on nothing. Now we explicitly set the html to a value (and text) of 'all' which allows us to remove a few other lines of code as well.
before:
` $select.html('')`
after:
` $select.html('<option selected value="all">All</option>')`
If UX reviewers decide that they still want a blank sub category dropdown at first and then "Choose a subcategory" once a primary is selected, this can be prepended to the top of the select options. Line 25 of `audit-category-sub-category.js` is commented out with this functionality  right now.

Issue #1777, opened 2018-02-09
Audit search styling and tags 


johnnyporkchops:
- [x] ImplementSCSS styles and any changes to html templates to make sub categories list indented under category as seen in below design mockup. 
- [x] Apply highlighting for chosen sub-category AND identify current current category/sub-category chosen in head of table, above filters(this is a proposed solution for MVP) --This is so that users know it is not a "filter" option but rather which category/sub-category combination is being currently searched.

Category/Sub-Category mockup (not live code screnshot):
![image](https://user-images.githubusercontent.com/5572856/36054250-5451f6c6-0dc3-11e8-8cf5-6157fac24441.png)

Related issues/PRs:
https://github.com/18F/fec-cms/issues/1304
https://github.com/18F/fec-cms/issues/1724

JonellaCulmer:
This is done. So closing.

Issue #1724, opened 2018-01-29
Review and Prioritize Feedback Received from Audit Team


JonellaCulmer:
We received great feedback from the audit team on the audit search currently in development. You can track that work here: https://github.com/18F/fec-cms/issues/1304

So that we can keep track of the feedback, we have a document of all we have received here: 
https://docs.google.com/document/d/1oOIPdsIMhxfymYo6Scmny5kMseb499jefw8ZinLwQGE/edit#

Some of the work is currently in progress, others need to be discussed and prioritized. 

cc: @jwchumley @patphongs @fec-jli

JonellaCulmer:
I've updated the feedback document to prioritize the recommendations/suggestions received into three categories: `What we can do`, `What we should do later`, and `What we can't do`. 

These categories are mainly to keep us organized and `What we can do` include task to ensure the implementation of a functional and usable audit search. `What we can do later` items are mostly "nice to have's" but are not critical to the successful launch of our audit search. Feedback listed under `What we can't do`, are things that are not supported by the user testing we've completed, run contrary to the rest of the site's design, or will not improve the usability of the search.

Here's the link again: https://docs.google.com/document/d/1oOIPdsIMhxfymYo6Scmny5kMseb499jefw8ZinLwQGE/edit#

cc: @jwchumley @patphongs @fec-jli @johnnyporkchops

llienfec:
@AmyKort asked me to :eyes: the audit priority document ^.  @JonellaCulmer did a great job analyzing the priorities and the items in each category make a lot of sense to me based creating an MVP, ranking longer term work, and only making changes that are supported by the usability testing that we conducted in 2017.

AmyKort:
Thanks, @llienfec and @JonellaCulmer !

JonellaCulmer:
Okay, great! Thanks @llienfec!

fec-jli:
good job Jonella.
1)for "What we can do now" part:
I agree with you. to add into current PR.
I have fixed "audit report result count is always 1000" problem. I will continue to work on others.
John will help to fix CSS issue.

2)for "What we should later" part:
I have done "Party committees into separate section" on my local. compare with classic auditsearch page. I remove useless committees type filters. will deploy to feature space later. 
for other requirements, we can do later.
Thanks

JonellaCulmer:
Thanks, @fec-jli. Since the Party committee move is already in progress I've moved it up to the `What we can do now`. Thanks again for taking a look. 

[...]

PR #1920, opened 2018-04-12
Feature/final audit search page

johnnyporkchops:
Replaces:
https://github.com/fecgov/fec-cms/pull/1883
History:
This migrates the classic audit search application from fec classic to the new fec website
This is the culmination of several PRs and issues listed below:
PR:
[WIP]-Datatable page for audit cases #1381
Issues:
[fec-cms] set default sub_category_id=all for audit-case endpoint #1882
Audit search styling and tags #1777
Build audit search page #1304
Review and Prioritize Feedback Received from Audit Team #1724


johnnyporkchops:
Latest changes for MVP:
- [ ] Link from http://127.0.0.1:8000/legal-resources/enforcement/audit-reports/
- [ ] Vertical alignment of findings column - @johnnyporchops
- [ ] Make PACS checkbox- @fec-jli 
- [ ] Test in IE , can you select more than one filter? @fec-jli 
- [ ] Breadcrumbs question @JonellaCulmer / @AmyKort

JonellaCulmer:
I swear these are the last few changes needed for the audit search. 

1. The default filter tag for `Primary Category: All` needs to be modified to match the updated filter category. It should read `Finding and issue category: all`. And update according when a category is selected, as it currently does, but remain as `Finding and issue category:`. Ex: `Finding and issue category: Allocation Issues`.
screen shot 2018-04-16 at 4 57 43 pm

2. After some followup discussion, I don't believe the double sort feature currently implemented on the audit search is ready for primetime. In order to use the double sort, users must be aware that they need to use the `shift` key in order to select multiple columns for sort. In the event that another column is chosen to sort, there may be difficulty in getting back to that original sort without having to refresh the entire page. 
gif1_audit

Finally, upon clicking on a particular column to turn off, the opposite column sorting is removed. These issues lead me to believe this feature should be turned off prior to implementation. Gif of the last issue below. 
gif2_audit

fec-jli:
Thanks @JonellaCulmer 
Let me answer your second question.
for double columns sort, the Erica just wants to happen when initial loading page, like classic website. after user click any other column to sort, no way to sort two columns at same time except user knows that secret `shift` key:)
we won't tell user how to know secret `shift` key. we just show you this feature for bonus and for a fun:)
until now, no website can do sort two columns at any time. Hopefully this can answer your second question.

johnnyporkchops:
Question 1 above:  Changed `Primary Category` to `Findings and issue` . Thanks for catching this @JonellaCulmer:
apr-16-2018 21-13-55

JonellaCulmer:
Thanks, all. I think we should move forward with removing the feature and explore additional sorting options at a later date. I don't think the feature is at a point now where it would function without further explanation from us on the front-end, which is not a best practice. 

cc: @johnnyporkchops @fec-jli @jenniferthibault @AmyKort

fec-jli:
Thanks @jenniferthibault.
remove default sort two columns may take a lot of work on api and some change on cms side.
especially for api, I need re-go over the endpoint function to remove all multi-sorting functions.
I really don't want to keep that function inside api. it involves many codes.  maybe can disable front -end is easier. I know on classic auditsearch, the users really wants to default two columns sorting feature. Thanks

fec-jli:
I just disable *initial two columns sorting* from front-end (fec-cms). 
modify this file to disable:   fec/static/js/pages/datatable-audit.js
and leave the python codes in endpoint. Thanks
@johnnyporkchops I have pushed the change to branch, Can you release to feature space?

JonellaCulmer:
With the addition of new committee types, we need to make some design changes to the front-end to accommodate them. Additionally, we need to remove one filter and edit the copy of three others. Below is an updated screenshot of what the committee types filters needs to look like. 

Under Party committees:
- [ ] Remove "National party nonfederal account" filter. There are no current audits of that type and that filter is misleading. We do not need it at this time and can always add it back later. 

Under IE committee types:
- [ ] Edit filter: Independent expeditor (person or group) to say Independent expenditure filer (not a committee) This update is being tracked in this issue: https://github.com/fecgov/fec-cms/issues/1907

Under PACs: (These changes 
- [ ] Edit filter: "Non - qualified non - party with non - contribution account" to say "PAC with non-contribution account - nonqualified"
- [ ] Edit filter: "Qualified non - party with non - contribution account" to say "PAC with non-contribution account - qualified"

Updated mock-up:
Committee types will be nested within an accordian and all be authorized committee filters will be including inside individual dropdowns. The authorized committee filters will use checkboxes as they currently do. 
screen shot 2018-04-20 at 1 29 43 pm

cc: @johnnyporkchops @fec-jli

johnnyporkchops:
`Remove "National party nonfederal account" filter. There are no current audits of that type and that filter is misleading. We do not need it at this time and can always add it back later.`
Is seems like removing a filter because there are no audits of that type is misleading. If  user wants to know if there are any audits of this type, "0 results" will tell them that there are none. Also, if we have audits of this type in the future, then it has to be added back in, which could be overlooked.

johnnyporkchops:
@JonellaCulmer slacked:
Our users aren't used to seeing that as an option since before we just listed lump categories such as PAC, Party, Congressional candidate, Independent expenditures, etc. Without there being any audits currently that fall under that category and that term not used so much now, I don't think it will hurt to remove it for now.
Additionally, that category can fit into the Party - qualified and unqualifed categories as well. So there still exists a way to categorize it.

Issue #1838, opened 2018-03-06
Icons missing from site search results list


jenniferthibault:
**Steps to recreate**

- Use the universal site search to search for any keyword or phrase (not a candidate or committee).
  - <img width="300" alt="screen shot 2018-03-06 at 6 52 30 pm" src="https://user-images.githubusercontent.com/11636908/37065624-8648e2f6-216f-11e8-9b76-a5ca359472d0.png">

- View the [search results page](https://www.fec.gov/search/?type=candidates&type=committees&type=site&query=contribution+limits)

**Expected behavior**
There should be an icon of a star showing up next to each match from the suggested search results list, and a document icon showing up next to each regular result. These icons help visually separate the results list items.

**Observed current behavior**
Doing a quick inspect shows that the code is there, but for some reason they're not showing up. This will need a bit of investigating/grooming.

<img width="748" alt="screen shot 2018-03-06 at 6 48 58 pm" src="https://user-images.githubusercontent.com/11636908/37065503-1bc829b4-216f-11e8-9565-95ce852e8b15.png">

<img width="752" alt="screen shot 2018-03-06 at 6 48 44 pm" src="https://user-images.githubusercontent.com/11636908/37065521-1e5413e6-216f-11e8-9350-06914fc19861.png">

johnnyporkchops:
@jenniferthibault , It looks like adding this to @line 44 of _icons.scss will make that document icon show up:
` &.i-document {
    @include u-icon-bg($document, $base);
  }`
I could not test search for contribution limits locally, but if  I search a candidate on local CMS, it returns results with the i-candidate icons. Changing one to `i-document` in the inspector before adding the line above did not show the document icon, but after adding it and building SASS, it did. And what comment from @johnnyporkchops would be complete without an accompanying GIF!
![i-document](https://user-images.githubusercontent.com/5572856/37067724-22461062-2179-11e8-8151-9f9bede43f26.gif)

jenniferthibault:
This might be another clue to a systemic bug. I just found that icons are also missing from the Document feed pages. I f you take a look and find that it's a different cause, I'm glad to make a separate issue!

https://fec.gov/about/reports-about-fec/strategy-budget-and-performance/

![screen shot 2018-03-19 at 9 27 57 pm](https://user-images.githubusercontent.com/11636908/37630724-86e3f3e8-2bbc-11e8-932a-9b3cadb79ab6.png)

johnnyporkchops:
[...]

PR #1961, opened 2018-05-01
Restore i-document,i-star and i-speaker icon classes

johnnyporkchops:
Restores missing star, document and speaker(audio) icons to search result, document and meeting pages
- Addresses : Icons missing from site search results list #1838
- Also restores audio icon to meeting pages
- Document searches do not work locally or on feature, so the only way (i could think of) to test this locally is to change the CSS class in the inspector to the newly added classes to make sure the icons show up. see screenshot below.

https://github.com/fecgov/fec-cms/blob/develop/fec/fec/static/scss/components/_icons.scss

### Search Results:
may-01-2018 14-44-16

### Document Pages:
may-01-2018 14-52-35

### Meeting page
may-01-2018 15-03-44

### DIFF
--- a/fec/fec/static/scss/components/_icons.scss
+++ b/fec/fec/static/scss/components/_icons.scss
@@ -51,6 +51,12 @@
   &.i-star {
     @include u-icon-bg($star, $base);
   }
+
+  &.i-speaker {
+    @include u-icon-bg($speaker, $base);
+  }
+
+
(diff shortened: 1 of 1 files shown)


JonellaCulmer:
@johnnyporkchops Unable to test the star icon locally. But the speaker and document icons are there.

Issue #1943, opened 2018-04-24
Implement new contact page design


AmyKort:
**What we're after:** We've reorganized and streamlined our contact page language and design to help users better find the FEC services they are looking for. Our next step is to implement the new design. Great background at: https://github.com/fecgov/fec-cms/issues/1895

**Designs:**
![contact_updated_ig](https://user-images.githubusercontent.com/31663028/39281163-4ac5fc30-48d1-11e8-913d-9bd578999304.png)

**Completion criteria:** 
- [x] Create wagtail template, or discover applicable, mobile-friendly template to function similarly to the hard-coded version (column 1 (left), then column 2(right)) (also talk to Jonella about line spacing)
- [x] Front-end and content pair to build contact page in Wagtail
- [x] Publish new wagtail contact page live
- [x] Delete hard-coded contact page
- [x] Modify link to contact page to say `www.fec.gov/contact` and remove `-us`
- [x] Add a wagtail redirect for for `/contact-us/` to go to `/contact/`

JonellaCulmer:
Once upon a time, we planned on removing the `-us` from the `www.fec.gov/contact-us` link in an effort to move away from first-person POV. Emileigh started a PR some time ago that is now out of date, but I think we can incorporate the intent of the PR in this implementation issue as well. 

Old PR: https://github.com/fecgov/fec-cms/pull/784

dorothyyeager:
Took a look at the templates in the pattern library to see if there was one that functioned similarly to the current page. The closest seemed to be the "custom page" which does permit two columns, but would require us to swap what's in the design. It looks like the best option will be to build a new Wagtail template.

johnnyporkchops:
@JonellaCulmer , do you have the text for the above mockup by any chance?

JonellaCulmer:
@johnnyporkchops I can compile the text in a google doc for you. 
The image in the mock-up is what is currently implemented.

JonellaCulmer:
@johnnyporkchops Here's the page content: https://docs.google.com/document/d/1dv-n3Wbqr3pnvu23vEdFN_xIfertUG6iWZS5aFwm0fQ/edit#

cc: @dorothyyeager

dorothyyeager:
@johnnyporkchops Happy to review the PR for content if needed.

JonellaCulmer:
This work has been merged into the next release, so closing this issue.

dorothyyeager:
[...]

PR #1986, opened 2018-05-09
Contact page new design and wagtail template

johnnyporkchops:
Implement new design for contact page and make it editable in Wagtail 
New url:  `/contact`

- [x] Requested text file for content of right column (currently has the old content as placeholder text)

- Addresses # https://github.com/fecgov/fec-cms/issues/1943

```
modified: fec/static/scss/components/_contact-items.scss
modified: home/blocks.py
new file: home/migrations/0101_auto_20180531_1615.py
modified: home/templates/blocks/contact-info.html
new file: home/templates/home/contact_page.html 
```
## Screenshots
may-09-2018 00-08-52


johnnyporkchops:
@JonellaCulmer and @dorothyyeager , when you start building the page i wagtail, you might find that you want to leave the `Item label` field blank and just use a headline tag in the rich text area instead, but you can play around with it. Also if any of the required fields cause a problem because you find you want to optionally leave them blank for some cases, I can make them optional.

still

JonellaCulmer:
@johnnyporkchops
@dorothyyeager and I took a look at the template and we have a few tweaks:
First, we'd like to make the item icon optional. This would help for the address section. We don't want to use the same icon twice in a row for both addresses and right now we'd be required to use them both. If they are made optional we can avoid this.

Second, can you rename "Sections" to "Help with sections". It wasn't immediately clear which section we were adding information too and whether we needed to include a header. If you modify that, it would help. Particularly in the future when someone else may need to go in and make edits.

Could you also edit "Help With" to "Help with"?

Linking didn't work in the item info section for internal or external links, but it did for email. These need to be functional, particularly for the Contact your analyst section

Finally, if we could limit the "Help with" section to just one section. I think that will help limit and future modification of the template.

Let me know if you have any thoughts or questions.

JonellaCulmer:
@patphongs I was able to publish a page locally without any errors.

johnnyporkchops:
@JonellaCulmer, @patphongs . Thanks for your reviews. I will take a look at these issues.

johnnyporkchops:
Here is a screenshot of the full contact page built using the template.
contact_full_page_composite

dorothyyeager:
Hi John - looks good. A couple of small fixes to request:
1) Help With - make that "Help with" (lower case w)
2) Information Division should be above its phone number and not bold (following the pattern used for the other divisions)

Thanks!

dorothyyeager:
It's a template (so no need to make my small edit as we'll do that once we make the real page). @JonellaCulmer and I paired with @johnnyporkchops and requested an edit to the "Help with" section - to make a separate, editable, section title in case a wording change is desired in the future. That's been made, and everything else looks good to go, so approving from my end (previewed it on John's local).

johnnyporkchops:
I  cleaned up the migration history for this PR so I am not pushing the history of adding/removing/renaming fields etc to the repo, rather just one migration file that represents the current and final state for this template's model. 
I also paired with @jculmer so se could reset her local database and run this locally with just the one migration.

Issue #1983, opened 2018-05-08
Check raising and spending breakdown charts


patphongs:
We need to verify the accuracy per cycle the data that is displayed in these charts:

**Raising breakdown:**
https://fec-stage-proxy.app.cloud.gov/data/raising/

**Spending breakdown:**
https://fec-stage-proxy.app.cloud.gov/data/spending/

Related issue #1949

llienfec:
Received feedback from a user about this graph - saying the numbers looked off and the interaction isn't working correctly.  It looks like Jeff verified the numbers above.  Here are 4 issues that I found as I tried to research the numbers and how the graph worked.  I think these 4 issues are eroding trust in the numbers:

![image](https://user-images.githubusercontent.com/22914750/40258443-a578664a-5abf-11e8-9c25-fb22c8254e6d.png)

1. We've noted that the scale of the map ($4 billion) doesn't work for the off-election year because it looks like no committees are raising money in 2017.  More than that, the singe orange line makes it look like only PACs are represented in the map, even though the key shows that candidates and parties should also show up.
1. When the user moves the mouse over the months of the year on the x-axis, the line moves on the month to the right of the mouse.  The user expected the mouse and the line to move together.
1. The user was looking for a line showing the "all committees" totals on the graph since it was listed in the key first, especially since this number is big enough to appear on the graph's scale in the off-election year.
1.  When trying to verify numbers, I tried to go to the "methodology" button.  When the screen pops up, you can't scroll down.  If your screen is small (verified on my tablet), you can't see all of the methodology information.

llienfec:
I talked to a data user about this chart in depth today.  Looking at the stats for 2016, it seems impossible that almost all of the money raised in the election happened between Dec 2017 and Jan 2018.  On a gut level, the data feels wrong.

I need to keep digging into this data, but may run out of time before the sprint is over.  One theory is a problem with the API data getting pulled into the graph.  I took a snapshot of the current graph to see if the data points decrease once the new filings come in.  It's going to take some time and some conversations about how the data is getting pulled in to figure out what's happening.    

![image](https://user-images.githubusercontent.com/22914750/40747259-bfee19a6-642a-11e8-86a0-3a1667846caf.png)

jwchumley:
I looked at the data coming out of the API and at the code that produces it. I'm pretty sure it's an API issue. For example, I pulled party disbursements by month from FECP and compared to the same thing in the API:


[...]

PR #2044, opened 2018-06-04
[WIP] Line chart with static data

patphongs:
## Summary (required)

- Addresses #1983 
_This will make the line chart on the data landing page static temporarily until the API call for `/totals/by_entity` is fixed._

**To do**
- [ ] Still need to add static data to display in the line chart
- [x] Fix failing JS tests to work with static data

-  Data landing page


patphongs:
This seems like it is no longer needed. If a table is going to be used to push the correct numbers into the line chart, we wouldn't need to pipe the static data into the array within this PR. I am closing for now and will resurrect if it is needed again.

Issue #1888, opened 2018-03-27
Committee addresses missing "Suite" on committee profile "About this committee" pages


JonellaCulmer:
There are a few committees that seem to be missing the Suite from their addresses when listed on the 'About this committee' pages on committee profile pages. 

Suites appear on the committee's Form 1 filing but it doesn't seem to make it to the 'About this committee' page. In order for someone to get the complete address, they need to view Form 1 to make sure they are seeing an accurate address. Moreover, they need to already _know_ to look at Form 1 to get the complete address. They cannot rely on what's listed in the 'About this committee' section alone. 

(check to see if data is pulling address line 2)

### Some examples: 
**Samsung Electronics America Inc Political Action Committee**
About this committee page: https://www.fec.gov/data/committee/C00590315/?tab=about-committee
Form 1: http://docquery.fec.gov/cgi-bin/forms/C00590315/1215949/

**Denali Leadership PAC**
About this committee: https://www.fec.gov/data/committee/C00438291/?tab=about-committee
Form 1: http://docquery.fec.gov/cgi-bin/forms/C00438291/1196606/

**Motor City PAC**
About this committee: https://www.fec.gov/data/committee/C00507574/?tab=about-committee
Form 1: http://docquery.fec.gov/cgi-bin/forms/C00507574/1162211/

AmyKort:
Is this ready to close?

JonellaCulmer:
Yes, the fix has been merged. Closing.

PR #2045, opened 2018-06-05
Include missing "Suite" in committee addresses 

johnnyporkchops:
Include missing "Suite" in committee addresses  on "About this committee" pages

- Addresses https://github.com/fecgov/fec-cms/issues/1888

- Added `Street_2` to template variables on data/views.py 

**[How to test &#187;](#how-to-test)** 

modified:   data/views.py

## Screenshots
suite
 <a id="how-to-test"></a> 

## How to Test:
Check out and run this branch locally: `feature/add-suite-to-committee-address`
Make sure suite numbers show on committee profile information on `About this committee` pages.
Example Pages:
http://127.0.0.1:8000/data/committee/C00507574/?tab=about-committee
http://127.0.0.1:8000/data/committee/C00438291/?tab=about-committee
http://127.0.0.1:8000/data/committee/C00590315/?tab=about-committee

### DIFF
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -319,6 +319,7 @@ def committee(request, committee_id):
         'committee_type': committee['committee_type'],
         'designation_full': committee['designation_full'],
         'street_1': committee['street_1'],
+        'street_2': committee['street_2'],
         'city': committee['city'],
         'state': committee['state'],
         'zip': committee['zip'],


JonellaCulmer:
@johnnyporkchops Looks really good. Seeing lots of suite numbers now!

Did find one weird one though, with just a "." on the second line. It seems the addition of the second address line just adds a "." at the end. This committee listed its suite on the first line.
Production: https://www.fec.gov/data/committee/C00007831/?tab=about-committee
Local: http://127.0.0.1:8000/data/committee/C00007831/?tab=about-committee

johnnyporkchops:
This looks like a one-off error in the data where someone added an errant period when entering the data. So it would not make sense to make a conditional in the template that would resolve this. How do we change the data that was input?

JonellaCulmer:
@johnnyporkchops If it's a one-off I say we leave it as is and focus on the success of the other 99%. No data modification is my recommendation.

JonellaCulmer:
@johnnyporkchops The "." is in the data. Is it possible to include an if/then statement to ignore the period if one is included on line 2?

johnnyporkchops:
@JonellaCulmer , we could but it seems like overkill for one data anomaly and it does not resolve for every possible mistake, just specifically for a period. I would tend to agree with your answer above to live with this small error

JonellaCulmer:
We've decided that for now we will move forward without a fix for the rogue "." 
Merging.

PR #2094, opened 2018-06-13
Rename calendar category 21 to 'Mid-year and year-end', move to bottom

### LINKED ISSUE #1981: Rename "All Reports" filter on calendar
The All Reports filter is not returning results on the calendar. This is partly due to the way the date is coded. However, in investigating this problem, we determined that the filter itself is misleading, since even when it functions properly it does not return "All Reports." We'll start work in a separate ticket to re-think the way this filter works. In the mean time, we will temporarily remove the nonfunctional 
filter from the calendar. 

image

lbeaufort:
## Summary (required)

- Resolves #1981 
_Rename calendar category 21 to 'Mid-year and year-end' and move to the bottom of the list._

- https://www.fec.gov/calendar/
- http://127.0.0.1:8000/calendar/?calendar_category_id=21

### Before
image

### After
image

### DIFF
--- a/fec/fec/constants.py
+++ b/fec/fec/constants.py
@@ -6,10 +6,10 @@ election_types = OrderedDict([
 ])
 
 deadline_types = OrderedDict([
-    ('21', 'Reporting deadlines'),
     ('25', 'Quarterly reports'),
     ('26', 'Monthly reports'),
-    ('27', 'Pre- and post-election')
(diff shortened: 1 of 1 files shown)


AmyKort:
Looks good to me, but I'm going to defer to @johnnyporkchops . Thanks!

PR #2235, opened 2018-08-01
Feature/election page legacy tab redirects

### LINKED ISSUE #2188: Add redirects for old election profile page links
@lbeaufort commented on [Thu Jun 28 2018](https://github.com/fecgov/fec-proxy/issues/88)

**What we're after:** After changing the design of the election profile pages, we changed the way data is organized in tabs on each election page. Unfortunately, the old links are still available and so we need to go back and add redirects for the old election profile page deep links.

The old links that are affected include: 
/?tab=totals
/?tab=contributions
/?tab=spending-by-others

This applies to all past /data/elections/senate, /house, and /president pages. 
Example: https://www.fec.gov/data/elections/president/2012/?tab=contributions

We should add redirects for the old election profile page deep links.

---

@AmyKort commented on [Mon Jul 16 2018](https://github.com/fecgov/fec-proxy/issues/88#issuecomment-405358385)

@patphongs  could you also take a look at this ticket to see if we need to update the work or the link?

---

@AmyKort commented on [Mon Jul 16 2018](https://github.com/fecgov/fec-proxy/issues/88#issuecomment-405358509)

Oh wait. Is this the ticket to update the links?

---

@patphongs commented on [Mon Jul 16 2018](https://github.com/fecgov/fec-proxy/issues/88#issuecomment-405367230)

@AmyKort Yes this is the one, I'm going to move it to fec-cms, since this actually needs to be done within the urls in python since nginx does not have the capabilities to pass arguments through to the resultant URL. 

Completion criteria:
- [ ] Use the URLS.py file to change the route to point to the appropriate view

johnnyporkchops:
## Summary
Redirect election page links that point to the old tab names to the corresponding new named-anchor link
Fix the issue with the page not jumping to the correct location for named-anchors regardless of wether it is a via one of these redirects or just a visit to a direct link with anchor suffix.

The old links that are affected include:
/?tab=totals
/?tab=contributions
/?tab=spending-by-others

- Resolves #2188

modified:   data/urls.py
modified:   data/views.py
modified: data/templates/elections.jinja

- [x] The named anchor links used in the new election page structure do not consistently scroll the page to the correct position because the browser jumps to the correct position on-load and then loads additional assets which pushes  desired location to a random place in page.

## How to test:
- Checkout and run this branch locally: `feature/election-page-redirects`
- Go to the `election pages` and confirm that the page redirects from its legacy  `tab=` argument to the correct  `#` anchor suffix. Here is the mapping of tab to anchor:
  - 'contributions':'#individual-contributions',
  - 'totals' : '#candidate-financial-totals',
  - 'spending-by-others':'#independent-expenditures'
- Also confirm that the page jumps to the correct location on the page based on the named-anchor tag.

Test pages:
http://127.0.0.1:8000/data/elections/house/AL/1/2012/?tab=spending-by-others
http://127.0.0.1:8000/data/elections/house/AL/1/2012/?tab=contributions
http://127.0.0.1:8000/data/elections/house/AL/1/2012/?tab=totals
http://127.0.0.1:8000/data/elections/house/AL/2012/?tab=contributions (no district included)

http://127.0.0.1:8000/data/elections/president/2012/?tab=contributions
http://127.0.0.1:8000/data/elections/president/2012/?tab=spending-by-others
http://127.0.0.1:8000/data/elections/president/2012/?tab=totals


johnnyporkchops:
@patphongs and @apburnes, I added a setTimeout script inside the Jinja template to resolve the problem with the anchors jumping to a location before the page has loaded. I feel that it is not the most elegant solution, but since this is the new Elections page structure(more anchors instead of just tab), it seems justifiable and somewhat necessary.
I tried to make it as unobtrusive as possible... I found that if I set it for 500ms instead of 1000ms, it is usually not enough time, but there is, of course now way to always know how long the page will take to load.

fec/data/templates/elections.jinja
```diff
@@ -60,7 +58,18 @@ var context = {
...
+
+$(document).ready(function() {
+  if(window.location.hash) {
+    setTimeout( function(){
+      $('html, body').animate({
+      scrollTop : $(window.location.hash).offset().top
+      })
+    }, 1000)
+  }
+});
```
patphongs: @johnnyporkchops This works and will probably catch most cases. Can you move this function inside the elections.js, that already has a ready function in it that you can use.

Just room for thought:
An enhancement may be to check inside the elections.js file when all the datatables are loaded using datatables `initComplete` callback https://datatables.net/reference/option/initComplete. There are three draw functions within the elections.js: `drawSizeTable`, `drawComparison`, and `drawStateTable`. Since there's multiple datatables loading, we would have to check if each one has loaded and then trigger the scroll. This may be a bit overkill though because if the number of datatables ever change on the page, the page could break.

johnnyporkchops: @patphongs , I think when I tried it inside `elections.js` it did not perform as well so I opted to put it inside the template where I had the most consistent results. But I tried so many variations, I can go back and check again.

apburnes: @patphongs suggestions on moving the JS out of the jinja template and trying to use asynchronous call is on point.  On problem with our current table functions are they are wrappers around the `datatables` api so we may need to look at how we can implement the asynchronous functionality.

 @johnnyporkchops your on track. We just need to figure out how to move that function from `SetTimeout` to after the tables are built.

fec/fec/static/js/pages/elections.js
```diff
@@ -11,6 +11,7 @@ var electionUtils = require('../modules/election-utils');
 var helpers = require('../modules/helpers');
 var ElectionForm = require('../modules/election-form').ElectionForm;
 var tableColumns = require('../modules/table-columns');
+var tabs = require('../vendor/tablist');
```
apburnes: Let's remove this because we didnt end up using it.

fec/data/urls.py
```diff
@@ -2,18 +2,20 @@
 
 from data import views
 from data import views_datatables
+from django.views.generic import RedirectView
```
apburnes: Is `RedirectView` doing anything?

PR #2254, opened 2018-08-06
[WIP] Feature/add future elections

### LINKED ISSUE #2087: Enable 2019-2020 on election search dropdown
Related issue: #2085 

**What we're after:** If we are to enable 2019-2020 data on election profile pages, then we need to make sure that is an option on the election search pages as well. 

This work is dependent on what happens in #2085.

Affected page: https://www.fec.gov/data/elections/ 

screen shot 2018-06-12 at 4 28 10 pm

lbeaufort:
## Summary (required)

- Resolves #2087 

-  

## Screenshots

- _Include a screenshot of the new/updated features in context (“in the wild”). If it is an interface or front end change, include both a Before and After screenshot._

## Related PRs

API change


fec/data/templates/election-lookup.jinja
```diff
@@ -18,7 +18,7 @@
     <div class="search-controls">
       <form>
         <div class="search-controls__row">
-          {{ select.cycle_select(cycles, location='form', range=True) }}
+          {{ select.cycle_select(cycles, cycle, location='form', id='cycle-1') }}
```
patphongs: Does the `id='cycle-1'` do anything? Looks like the only other time this id is used is within the candidate financial summary jinja file. I don't see that ID being referenced in any scripts though.

Also, do we want to remove `range=True` here? We tend to show the for years on our drop down selections don't we?

patphongs: When we were looking at the About this election tab, the Last general election and general election dates were wrong for future elections (2020 and 2022). Is there an easy way to fix that? Otherwise we can write into the template to remove those dates for future election years?

lbeaufort:
Thanks @patphongs - it looks like the dates show up on `dev` too - do you know what could have changed? https://fec-dev-proxy.app.cloud.gov/data/elections/president/2016/?tab=about-election#state-election-offices

fec/data/templates/election-lookup.jinja
lbeaufort: I believe `id='cycle-1'` is for styling - it makes the year dropdown "long" vs "short":
image
image

We do want to remove `range=True` and just show the election year because the 2-year range doesn't apply to 4/6-year election cycles.

lbeaufort:
@patphongs it looks like this logic is incorrect on production too - it's not calculating properly for Senate either. 

image

lbeaufort:
Update: there were more issues with these dates than just the future elections, so I put in a PR to the release branch to remove them until we can make the API changes to support this feature: https://github.com/fecgov/fec-cms/pull/2268

PR #2258, opened 2018-08-07
Add API caller logs

lbeaufort:
## Summary (required)

**Innovation sprint**
- Add logging for server-side API calls so we can track when the CMS makes server-side (AKA python) API calls. Right now you can see the client-side (AKA JavaScript) calls in the developer tools, but the only way to see the server-side calls that we’ve figured out is to run the API locally and check the API logs. 
- Clean up some formatting and make indentation consistent (thanks @apburnes!)
- fix an unused variable in `call_senate_specials`

This adds info-level logging to `_api_caller` along with which function is calling the API caller.

-  Server-side AKA Python API calls using `_api_caller()`

## Screenshots

screen shot 2018-08-07 at 5 49 01 pm

### DIFF
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -7,7 +7,6 @@ import inspect
 from collections import OrderedDict
 from operator import itemgetter
 from urllib import parse
-
 from django.conf import settings
 from django.http import Http404
 
@@ -16,10 +15,8 @@ from data import constants, utils
 
 MAX_FINANCIALS_COUNT = 4
 
-
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
-
 
 session = requests.Session()
 http_adapter = requests.adapters.HTTPAdapter(max_retries=2)
@@ -181,7 +178,14 @@ def load_single_type(data_type, c_id, *path, **filters):
 
 def load_nested_type(parent_type, c_id, nested_type, *path, **filters):
     # Call API with nested types in load_with_nested
-    return _call_api(parent_type, c_id, nested_type, *path, per_page=100, **filters)
+    return _call_api(
+        parent_type,
+        c_id,
+        nested_type,
+        *path,
+        per_page=100,
+        **filters
+    )
 
 
 def load_with_nested(primary_type, primary_id, secondary_type, cycle=None,
@@ -196,7 +200,13 @@ def load_with_nested(primary_type, primary_id, secondary_type, cycle=None,
     """ Get data for just primary_type
         Example: candidate data for /candidate/* or committee data for /committee/*
     """
-    data = load_single_type(primary_type, primary_id, *path, per_page=1, **query)
+    data = load_single_type(
+        primary_type,
+        primary_id,
+        *path,
+        per_page=1,
+        **query
+    )
 
     cycle = cycle or max(data['cycles'])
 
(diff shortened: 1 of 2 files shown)


fec/data/api_caller.py
```diff
@@ -253,33 +259,33 @@ def result_or_404(data):
...
-        response = _call_api(
-            'candidates', 'totals',
-            sort_hide_null=True,
-            cycle=cycle,
-            election_full=False,
-            office=office,
-            sort=sort,
-            per_page=per_page
-        )
+    response = _call_api(
```
apburnes: Format discrepancy.

fec/data/api_caller.py
```diff
@@ -253,33 +259,33 @@ def result_or_404(data):
...
+    return response if 'results' in response else None
 
 
 def load_top_pacs(sort, cycle=constants.DEFAULT_TIME_PERIOD, per_page=5):
-        response = _call_api(
-            'totals', 'pac',
-            sort_hide_null=True, cycle=cycle, sort=sort, per_page=per_page
-        )
-        return response if 'results' in response else None
+    response = _call_api(
```
apburnes: Format discrepancy.

fec/data/api_caller.py
```diff
@@ -253,33 +259,33 @@ def result_or_404(data):
...
+    return response if 'results' in response else None
 
 
 def load_top_parties(sort, cycle=constants.DEFAULT_TIME_PERIOD, per_page=5):
-        response = _call_api(
-            'totals', 'party',
-            sort_hide_null=True, cycle=cycle, sort=sort, per_page=per_page
-        )
-        return response if 'results' in response else None
+    response = _call_api(
```
apburnes: Format discrepancy.

fec/data/api_caller.py
```diff
@@ -333,14 +339,12 @@ def call_senate_specials(state):
         given state. Returns a list of dictionaries
         Example: [{details for election1}][{details for election2}]
     """
-    api_response = _call_api('election-dates',
-                             election_type_id='SG',
-                             office_sought='S',
-                             election_state=state)
-
-    special_results = api_response['results']
+    special_results = _call_api('election-dates',
```
apburnes: Format discrepancy.

lbeaufort:
Thanks @apburnes! I pushed up changes to formatting and made a change to use `.get()` when possible to improve readability.

apburnes:
Getting an error on `/data` due to the use of `.get` in the `load_top_candidates`function.

https://github.com/fecgov/fec-cms/blob/2652189fd4c03e88992be94c2ce2c49f6f24bc04/fec/data/api_caller.py#L283-L293

screen shot 2018-08-08 at 5 01 45 pm

apburnes:
May have something to do with the `_call_api` function returning json and not a dictionary.

https://github.com/fecgov/fec-cms/blob/2652189fd4c03e88992be94c2ce2c49f6f24bc04/fec/data/api_caller.py#L43-L52

lbeaufort:
@apburnes thank you - I misread this - the original returns the whole response, not just the results. I'll put back the changes. 

I started to add some API caller tests because our test coverage is 0% but it seemed like a big task. Maybe something worth doing this sprint.

lbeaufort:
@apburnes thanks again for taking a look. While using this feature to diagnose an issue, I realized I wasn't logging the parameters. I changed the logging to output `response.url` (includes parameters) instead of just `url` (just the base URL)

Issue #2264, opened 2018-08-09
Wagtail dealbreakers (before release Penny Robinson 6.5 on 8/14)


patphongs:
## Priority need to fix
- [x] Resource and collection template: the left nav is not scrolling to anchors. **(Andrew)**
- [x] Press release, digests, tips, and record all need to show up on the following below. If they don't, we need to fix it. **(John)** **Tested and working properly, please see** [2266](https://github.com/fecgov/fec-cms/issues/2266)
    - Homepage (all)
    - Updates page (all)
    - Services landing page (for tips only)
- [x] image and document upload tool is not working. The path seems correct but probably never executes upload to bucket. **(Pat)**
- [x] Contact page renders incorrectly, **Fix for this: Change `block.value` to `block` here:** https://github.com/fecgov/fec-cms/blob/a0cf3714a312ce29107d4cad8cbfbd66f78d7659/fec/home/templates/home/contact_page.html#L18 
![contact_error](https://user-images.githubusercontent.com/5572856/43935942-44b68dea-9c24-11e8-9cbe-81a124272320.jpg)

## Less severe, will need to fix, maybe not before release?
- Document feed OIG page is not working at all
- Copying text in RTE is not copying all of the content elements like glossary and links. Possibly need to upgrade react to v16.0.0 in order to fix this problem. Draftail 0.17.2 is the latest tagged release which has a fix on copying and pasting, but we can't install the latest copy fix without upgrading react first. 
- Contact page is not editable in Wagtail (Error: End of block reached without closing entity elements)
**Fix- Recreate the Contact page once Wagtail 2.0 is live on prod to resolve conflicts with existing html in RTE. Copy/Paste text from RTE sections directly from live page in browser.** Regular links are preserved on copy/past but email links need to be relinked in RTE
This is in addition to making the template block change in the "Priority Need to Fix" section above.

patphongs:
All of the priority issues have been addressed in this issue. The less severe items have been moved to this issue: https://github.com/fecgov/fec-cms/issues/2274

PR #2265, opened 2018-08-09
Remove filter remove_digits from section titles and anchors

apburnes:
## Summary (required)

- One piece of #2264 

_Resource and collection template: the left nav is not scrolling to anchors.  Looks like the templates and the section nav partial were using the `filter remove_digit` function. Not sure why that was used in the creation of the anchors.  If numbers are **not** allowed in section title anchors, this should not be merged._

- `fec/home/templates/blocks/section.html`
- `fec/home/templates/partials/section-nav.html`

### DIFF
--- a/fec/home/templates/blocks/section.html
+++ b/fec/home/templates/blocks/section.html
@@ -1,7 +1,7 @@
 {% load wagtailcore_tags %}
 {% load filters %}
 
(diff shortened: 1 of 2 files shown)


patphongs:
Thanks for catching this @apburnes! Not sure if this was a business requirement to begin with or not, so I've pinged the content team to get their thoughts. This might be an old rule that is either not known or now unwanted, will merge if they say it's good to go.

patphongs:
And content just confirmed that there is no business req for numbers to be stripped out of anchors, so we should be good to merge. @JonellaCulmer, since you're on the review list, can you review and merge?

PR #2281, opened 2018-08-14
Removed <a> from anchor

patphongs:
## Summary

_Needed to update the draftail anchor component to remove the <a> tag from the anchor so that this component is not a link by default. Also updated the styles for the anchor when published and within the editor._


patphongs:
> Should we remove the helpers.anchorify function from the modules/helpers.js and its tests/helpers.js?

Let's leave it in there for now and we can remove it later if we find we are not using it in the future.

PR #2317, opened 2018-08-28
[WIP] Adds Recaptcha capabilities on website forms

### LINKED ISSUE #2205: Form human validation tool for feedback tool
**What we're after:** 

**Completion criteria:**
- [ ] Test for 508 compliance

patphongs:
## Summary

- Resolves #2205 

We needed a way to safeguard against bots submitting to our forms on the website. There are currently 2 forms that this effects: Feedback drawer (globally across our website) and RAD Contact analyst form.

This uses invisible reCAPTCHA, which requires no additional work on the user's side to submit a form. The reCAPTCHA will only be invoked after the user clicks on the submit/post button. By default only the most suspicious traffic will be prompted to solve a captcha. Read more here: https://developers.google.com/recaptcha/docs/versions#invisible

This PR utilizes the "automatically bind the challenge to a button" to invoke the invisible reCAPTCHA.

1) Feedback drawer implementation: https://developers.google.com/recaptcha/docs/invisible#auto_render. This is used for the feedback drawer, since it is used on every page on our site and can be easily invoked with the `submitFeedback`callback function. 

2) RAD Contact form implementation: https://developers.google.com/recaptcha/docs/invisible#programmatic_execute. This was needed for the RAD Contact form. Since the RAD form is inside of a separate template and the JS is tied via a module export, the submit callback function had to be handled within the HTML form itself and the invoke recaptcha function is defined within the JS file for the call back to work.

Both forms handle server side verification through the `views.py` in data and home. Uses the `g-recaptcha-response` POST parameter. https://developers.google.com/recaptcha/docs/verify

To do: Need to fix tests

- Feedback drawer (On every page of the site)
- RAD Contact form (http://localhost:8000/help-candidates-and-committees/question-rad/)

### DIFF
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -6,6 +6,7 @@ from django.conf import settings
 
 from distutils.util import strtobool
 
+import requests
 import datetime
 import github3
 import json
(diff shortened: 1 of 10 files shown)


patphongs:
Decided to split up the work for implementing recaptcha to the feedback drawer and the RAD contact form. The feedback drawer implementation is now in this separate PR: https://github.com/fecgov/fec-cms/pull/2367

PR #2322, opened 2018-08-30
Feature/enhance utils tests

vrajmohan:


fec/data/tests/test_utils.py
```diff
@@ -49,6 +50,14 @@ def test_election_url(self):
         assert (filters.get_election_url(candidate, 2012)
                 == '/data/elections/house/NJ/02/2012')
 
+    def test_election_title(self):
```
lbeaufort: This is a great addition 👍 better than my "here be dragons" comment...

fec/data/utils.py
```diff
@@ -13,61 +9,6 @@ def current_cycle():
     year = datetime.datetime.now().year
     return year + year % 2
 
-class LRUCache(cachecontrol.cache.BaseCache):
```
lbeaufort: This code always looked intimidating - thanks for investigating into whether we actually used it.

PR #2324, opened 2018-08-31
Feature/use pytest

vrajmohan:


vrajmohan:
IMHO, the API decision to split it out was misguided. It helps only in the _very rare_ case that someone wants to run the application without wanting to run tests on it. I can be persuaded to fix it here as well.

PR #2345, opened 2018-09-10
Fix/audit search subcats

### LINKED ISSUE #2342: audit search sub-categories are not populating on dev and staging
Something I noticed over the weekend that I don't think is to be expected on dev/stage environments: audit search sub-categories are not populating on dev and staging. 
https://fec-stage-proxy.app.cloud.gov/legal-resources/enforcement/audit-search/

johnnyporkchops:
## Summary 

- Resolves #2342
`var = require` statements were removed by the format-linter because they were not being used as variables. However, the require statement was still making the script available to the page and running it. So removing them broke part of the datatable. Worked with @apburnes to refactor these scripts so they can be required in the app without failing the linters.

## Impacted areas of the application:
modified:   fec/static/js/modules/audit-category-sub-category.js
modified:   fec/static/js/modules/audit_tags.js
modified:   fec/static/js/pages/datatable-audit.js

To test:
Run this branch locally and 
1) Make sure that the Sub-category dropdown works and that the filter tag in the head of the table populates with `Findings and issue category: All` upon page-load 
2) And then this filter tag updates as `Findings and issues` and `Sub categories` dropdown options are selected. 
You can compare with  broken behavior here:
https://fec-stage-proxy.app.cloud.gov/legal-resources/enforcement/audit-search/

### DIFF
--- a/fec/fec/static/js/modules/audit-category-sub-category.js
+++ b/fec/fec/static/js/modules/audit-category-sub-category.js
@@ -9,38 +9,39 @@
 var $ = require('jquery');
 var helpers = require('./helpers');
 
-$('#primary_category_id').change(function() {
-  var $select = $('#sub_category_id');
-  $.getJSON(
-    helpers.buildUrl(['audit-category'], {
-      primary_category_id: $('#primary_category_id').val(),
-      per_page: 100
-    }),
-    function(data) {
-      $select.html('<option selected value="all">All</option>');
-      $('.sub--filter--indent')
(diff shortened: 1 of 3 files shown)


apburnes:
@johnnyporkchops everything worked and then ended up putting the `auditCategorySubcategory` function just below the `auditTag` function but before the `table` function.  Try it out and let me know if it works the way you expected.

johnnyporkchops:
@apburnes , good call. I tested locally and works well.

PR #2367, opened 2018-09-14
Feedback drawer RECAPTCHA implementation

### LINKED ISSUE #2205: Form human validation tool for feedback tool

patphongs:
## Summary

- Resolves #2205 

Use invisible RECAPTCHA to safeguard against bots submitting to our global feedback drawer form on the website. 

This uses invisible reCAPTCHA, which requires no additional work on the user's side to submit a form. The reCAPTCHA will only be invoked after the user clicks on the submit/post button. By default only the most suspicious traffic will be prompted to solve a captcha. Read more here: https://developers.google.com/recaptcha/docs/versions#invisible

This PR utilizes the "automatically bind the challenge to a button" to invoke the invisible reCAPTCHA: https://developers.google.com/recaptcha/docs/invisible#auto_render. This is used for the feedback drawer, since it is used on every page on our site and can be easily invoked with the `submitFeedback`callback function. 

Form handles server side verification through the `/data/views.py`. Uses the `g-recaptcha-response` POST parameter. https://developers.google.com/recaptcha/docs/verify

- Feedback drawer (On every page of the site)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,6 +8,7 @@
     'setTimeout': true,
     'clearTimeout': true,
     'ga': true
+    'grecaptcha': true
   },
   "extends": [
     "plugin:react/recommended",
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,7 +7,7 @@
     'module': true,
     'setTimeout': true,
     'clearTimeout': true,
-    'ga': true
+    'ga': true,
     'grecaptcha': true
   },
   "extends": [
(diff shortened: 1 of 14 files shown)


patphongs:
> This will break the `reaction-box.js` if we do not add the recaptcha logic to it because it uses the same endpoint `/data/issue`

Thanks for catching that @apburnes! I've added the recaptcha to the reaction boxes too so they should go through now

Issue #2375, opened 2018-09-17
Insert HTML code block for Press Contact after 9/25 deploy


patphongs:
We needed to update the press landing page so that content owners can update the press contact information without needing a code change. The following HTML block needs to be inserted into the contact intro section of that page. 

Page: https://www.fec.gov/press/

Use the contact block to create this section on the Press landing page. Update the office hours to 7:00am to 5:30pm.

The above HTML block will render like this:

<img width="747" alt="screen shot 2018-09-17 at 11 33 06 am" src="https://user-images.githubusercontent.com/12799132/45633341-89267280-ba6d-11e8-9f0d-d5fcb235616d.png">

patphongs:
Press page has been updated with most recent contact information: https://www.fec.gov/press/#contact

PR #2376, opened 2018-09-17
Removes hardcoded contact page information

### LINKED ISSUE #2374: Update press landing contact section
We received a request to change the contact section information on the press landing page (See below). The press landing page has a hard coded contact section, so we'll need to remove that hardcoded section and replace it with an HTML block in Wagtail so that we can edit without needing to modify the code.

**Current:**
Local
(202) 694-1220
7:00 a.m. to 6:00 p.m.
M-F, Eastern Time
 
**Update  to:**
7:00 a.m. to 5:30 p.m.

patphongs:
## Summary

- Resolves #2374 
_This PR removes the hardcoded contact page information from the press landing page and leaves only the intro section. Needed to do this so that we can insert an HTML block into the contact intro for easier changes without needing a code change._

Follow-up issue to insert the HTML block is here: https://github.com/fecgov/fec-cms/issues/2375

-  Press landing page

### DIFF
--- a/fec/home/models.py
+++ b/fec/home/models.py
@@ -58,7 +58,8 @@ stream_factory = functools.partial(
         ('html', blocks.RawHTMLBlock()),
         ('image', ImageChooserBlock()),
         ('table', TableBlock()),
-        ('custom_table', CustomTableBlock())
+        ('custom_table', CustomTableBlock()),
+        ('contact', ContactInfoBlock())
     ],
 )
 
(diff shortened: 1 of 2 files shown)


johnnyporkchops:
@pat, I tested and this works great. We we could add `('contact_info', ContactInfoBlock())` to the [streamfactory](https://github.com/fecgov/fec-cms/blob/60d148980703b7be57dabf791946abd4f75cde46/fec/home/models.py#L53-L63) and one could build the same content and it renders visually, exactly the same.(no migration required). I guess the only positive to that is that the users would have less HTML to edit and they would be editing in RTE or single field areas. But either way it works. 
**Here is a screenshot:**
screen shot 2018-09-17 at 6 09 27 pm

patphongs:
Excellent idea @johnnyporkchops, thanks for helping me with that, the contact block has been added.

PR #2391, opened 2018-09-19
Fixed election dropdown on candidate profile page

### LINKED ISSUE #2341: Fix election dropdown on candidate profile page
**What we're after:** Need to correct the election data dropdown on candidate profile pages. It is not displaying correctly on the Raising and Spending tabs. 

**Completion criteria:**
- ~Make clearer options for size of election dropdown box~ _Found out that the `id` is actually used for the label to focus on the correct select box. So this criteria isn't needed anymore_
- [x] Fix sizing of election drop down box
- [x] Change dropdown label to `Two-year period` to remain consistent with implementation of two year cycles

**Before:**
screen shot 2018-09-10 at 9 13 05 am

**After:** 
screen shot 2018-09-10 at 9 14 09 am

patphongs:
## Summary

- Resolves #2341 
_Inserts additional div container with cycle classes to fix election dropdown width._

http://localhost:8000/data/candidate/S2MA00170/?tab=raising
http://localhost:8000/data/candidate/S2MA00170/?tab=spending

-  Candidate profile pages: Raising & Spending tab

### DIFF
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -30,15 +30,17 @@
 {% macro candidate_cycle_select(cycles, cycle=none, id='') %}
 {% set cycle = cycle | int %}
   <div class="content__section">
-    <label for="{{id}}-cycle" class="label cycle-select__label">Election</label>
-    <select id="{{id}}-cycle" class="js-cycle" name="cycle" data-cycle-location="query" data-election-full="False">
-      {% for each in cycles | sort(reverse=True) %}
-        <option
-            value="{{ each }}"
-            {% if cycle and cycle <= each and cycle > (each - 2) %}selected{% endif %}
-          >{{ each }}</option>
-      {% endfor %}
-    </select>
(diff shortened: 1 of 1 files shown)


JonellaCulmer:
@patphongs What do you think about also eliminating the time period from these pages as well. If we changing the dropdown, the additional text is redundant. 

screen shot 2018-09-19 at 2 10 36 pm

patphongs:
> What do you think about also eliminating the time period from these pages as well? If we are changing the dropdown, the additional text is redundant.

@JonellaCulmer Since there's no indication at first glance that 2014 is the only available 2 year time period within that election year, I think it would be best that we keep this there. I could be persuaded othewise if you find a more compelling reason to remove it.

JonellaCulmer:
@patphongs I'm noting here that there are some responsive issues with the filters at the top. This is behaving this way on production, so it's not new, but I'd like to explore this asap if possible.

patphongs:
@JonellaCulmer Thanks for flagging that. Since this is a production issue, I'd like to push that into a new issue to fix and get this merged.

PR #2409, opened 2018-09-28
[WIP] Election search page redesign

### LINKED ISSUE #2353: Implement updated election search page design
History: https://github.com/fecgov/fec-cms/issues/2286

**What we're after:** It's been observed that users have had difficulty finding search results on the election search page. Often the results populate below the fold and users cannot tell that the search has worked.

To address this issue, we have redesigned the election search page, to make more room for the results to populate on the page. Allowing users to better see and validate when a search has completed.

**Design:**
screen shot 2018-09-11 at 2 42 55 pm

**Completion criteria:**
- [x] Modify the display of search filters into gray row. This is similar to the way filters are display on the [Latest Updates](https://www.fec.gov/updates/) page.
- [x] Remove the default Results language "This page shows...show up here."
- [x] Modify the search button for ZIP code. Buttons should become one, search box and button combined. Similar to the way it's displayed on datatable search filters. 
- [x] Change page title from "Search" to "Find elections"
- [ ] Show upcoming presidential election profile page by default. For example, we want to show 2020 for for this current time period.
- [ ] Add to release notes

johnnyporkchops:
## Summary 

- Resolves #2353
Implement redesigned election search page

modified:   data/templates/election-lookup.jinja
	modified:   fec/static/scss/components/_search-controls.scss

### DIFF
--- a/fec/data/templates/election-lookup.jinja
+++ b/fec/data/templates/election-lookup.jinja
@@ -10,73 +10,66 @@
 
 {% block body %}
   {{ header.header(title) }}
-<section class="main" >
+<section class="main election-search" id="election-lookup" >
   <div class="container">
     <header class="heading--main">
       <h1>Search</h1>
     </header>
-</section>
-<section class="main__content--full data-container__wrapper election_search" id="election-lookup">
-  <div class="search-controls filters--horizontal slab--home">
-    <form class="js-form-nav container">
-
-      <div class="search-controls__either">
-        <div class="row">
-          <div class="search-controls__cycle">
-          {{ select.cycle_select(cycles, cycle, location='form', id='cycle-1') }}
-          </div>
-
-
-          <div class="search-controls__zip combo combo--search__medium combo--search--inline">
-            <label for="zip" class="label">Find by ZIP code</label>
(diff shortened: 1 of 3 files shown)


JonellaCulmer:
Based on the incredible conversation we just had about how this page will function, please see the below modifications:
- [ ] Presidential elections will appear in the results list, only in presidential years, regardless of whether or not other filters are selected. ex. select 2020 - presidential election will appear. Select 2018, nothing will appear. Only after a state or ZIP are selected will House and Senate elections appear in the results list. 
- [ ] All other filter functionality stays the same.
- [ ] Gray bar with filters will be broken up into two, with election in one section and the other filters in another. These gray boxes will not span the width of the page, but will contain only those filters and take up that specific space.
- [ ] Display "Results" text only if selections to the filters have been made and there are in fact results to display
- [ ] Create static section for upcoming Presidential election below the map with the text: "You might also like"

**Design:**
screen shot 2018-10-03 at 12 45 09 pm

patphongs:
This PR has been replaced by https://github.com/fecgov/fec-cms/pull/2432

PR #2424, opened 2018-10-04
Fix recaptcha for Internet Explorer

### LINKED ISSUE #2205: Form human validation tool for feedback tool

johnnyporkchops:
## Summary 
Recaptcha badge placement was breakimg in  Internet Explorer
- Resolves #2205
- Related PR #2353, #2421

modified:   fec/static/scss/components/_feedback.scss
This appears on all pages of site inside feedback tool

## Screenshots
recaptcha_ie

### DIFF
--- a/fec/fec/static/scss/components/_feedback.scss
+++ b/fec/fec/static/scss/components/_feedback.scss
@@ -109,17 +109,23 @@
 }
 
 .feedback__button {
-  margin-top: u(1rem);
+  margin-top:1rem;
 }
 
 .grecaptcha-badge {
   position: relative;
   bottom: u(-10rem);
+}
 
+.feedback__button {
+  margin-top:1rem;
 }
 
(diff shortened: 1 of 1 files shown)


fec/fec/static/scss/components/_feedback.scss
```diff
@@ -109,17 +109,23 @@
 }
 
 .feedback__button {
-  margin-top: u(1rem);
+  margin-top:1rem;
```
patphongs: @johnnyporkchops I thought that `u( )` was standard for rems. Just curious why it was removed here, was it an IE problem?

johnnyporkchops: Thanks Pat,this just got changed when I was testing different setting and overlooked. I have restored  it to the way it was before.

PR #2432, opened 2018-10-10
NEW Election search redesign

### LINKED ISSUE #2353: Implement updated election search page design

johnnyporkchops:
## Summary
Election search page redesign after additional grooming. 
This PR replaces : [WIP] Election search page redesign #2409 

- Resolves #2353
 This PR replaces : [WIP] Election search page redesign #2409

Design and Functionality changes to the  /data/elections (elections search page) as outlined mockup and in last conversation in PR 2409 : https://github.com/fecgov/fec-cms/pull/2409#issuecomment-426711308
This includes changes to SCSS and JS files. 

## Screenshots
oct-10-2018 00-01-19

### DIFF
--- a/fec/fec/static/js/templates/upcomingPresidential.hbs
+++ b/fec/fec/static/js/templates/upcomingPresidential.hbs
@@ -1,0 +1,17 @@
+<div class="slab--inline slab--neutral callout" data-office="{{office}}">
+  <span class="t-note t-small">You might also like:</span>
+  <h3 class="u-margin--bottom">
+    {{#if district}}
+    <span data-color="{{color}}"></span>
+    {{/if}}
+    <a href="{{url}}">{{electionName}}</a>
+  </h3>
+  <div class="row">
+    <span class="result__term">{{electionType}}:</span>
+    <span class="result__definition">{{electionDate}}</span>
+  </div>
+  <div class="row">
+    <span class="result__term">Current candidates:</span>
+    <a class="result__definition" href="{{url}}">Show all &raquo;</a>
+  </div>
+</div>
--- a/fec/fec/static/js/templates/upcomingPresidential.hbs
+++ b/fec/fec/static/js/templates/upcomingPresidential.hbs
@@ -1,9 +1,6 @@
 <div class="slab--inline slab--neutral callout" data-office="{{office}}">
   <span class="t-note t-small">You might also like:</span>
   <h3 class="u-margin--bottom">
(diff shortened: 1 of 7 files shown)


johnnyporkchops:
There are at least two initial issues unresolved with this. Regarding the checklist  item, I would argue that is should show the results area titles "Results" when there are no results. It might be  confusing for it to be completely blank when there are no results.
- [ ] Display "Results" text only if selections to the filters have been made and there are in fact results to display

The gray cycle area does not maintain equal height to the zip/state/district area at all sizes, but it has been given an explicit height that keeps it uniform at all but one breakpoint. This is very tricky given all the resizing and stacking that takes place on this new design.
oct-10-2018 00-10-39

johnnyporkchops:
oct-11-2018 23-00-46

fec/fec/static/js/templates/upcomingPresidential.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="slab--inline slab--neutral callout" data-office="{{office}}">
+  <span class="t-note t-small">You might also like:</span>
+  <h3 class="u-margin--bottom">
+    {{#if district}}
+    <span data-color="{{color}}"></span>
+    {{/if}}
```
patphongs: Since this template is only used for just presidential elections, I am going to remove this district conditional

patphongs: Going to remove this conditional since it doesn't apply for the presidential template

PR #2451, opened 2018-10-15
Show exact counts under 500K

### LINKED ISSUE #2450: Use exact counts for datatable results
The CMS function `getCount`[(link)](https://github.com/fecgov/fec-cms/blame/develop/fec/fec/static/js/modules/tables.js#L97-L108) rounds datatable counts that are over 1,000 to the nearest thousand. 

Because the API now returns accurate counts above 500,000, we should:

- [x] provide the exact counts where results are 500,000 or less.
- [x] remove the "about" when the results are 500,000 or less.

lbeaufort:
## Summary (required)

Resolves #2450: **Use exact counts for datatable results**

- Because the API now returns accurate counts above 500,000, provide the exact counts where results are 500,000 or less.
-  Remove the "about" when the results are 500,000 or less.
- Remove the audit check because we have  around 1K audits total,  I didn't think we'd ever get over 500,000, and it might slightly improve performance not to do the extra check.

-  Data table counts

## Screenshots
screen shot 2018-10-15 at 4 38 32 pm
screen shot 2018-10-15 at 4 38 21 pm

### DIFF
--- a/fec/fec/static/js/modules/tables.js
+++ b/fec/fec/static/js/modules/tables.js
@@ -661,11 +661,15 @@ DataTable.prototype.fetchSuccess = function(resp) {
   var changeCount = this.newCount - this.currentCount;
 
   var countHTML =
-    this.newCount > 0
-      ? 'about <span class="tags__count">' +
+    this.newCount > 0 && this.newCount <= 500000
+      ? '<span class="tags__count">' +
         this.newCount.toLocaleString('en-US') +
         '</span>'
-      : '<span class="tags__count">0</span>';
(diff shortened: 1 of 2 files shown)


fec-jli:
I tested on my local cms, works well. Thanks
one question. do we want to show exact total page number at the bottom?

pkfec:
Good work @lbeaufort . counts appear as expected.

lbeaufort:
@fec-jli I think showing page numbers would be helpful, but as far as I can tell we've never shown page numbers.

Issue #2117, opened 2018-06-21
Dropdown filter disappears after options are selected then deselected


JonellaCulmer:
**What's we're after:** In the audit search and filings data tables, selecting and then deselecting an option makes the option disappear entirely. We need to correct this so that options can be selected/deselected multiple times without requiring the user to refresh the page. 

Once you deselect the committees you no longer want to use and close them out using the "X" the category disappears but does not reappear in a dropdown, unless you refresh the page. Please see the below gif and screenshot. 

![audit_filters](https://user-images.githubusercontent.com/31663028/41689480-c3da0ec2-74be-11e8-9e02-2a8543fe8d44.gif)
<img width="300" alt="screen shot 2018-06-20 at 7 12 14 pm" src="https://user-images.githubusercontent.com/31663028/41689486-cd134864-74be-11e8-9db5-037405fec0dc.png">

PR #2471, opened 2018-10-24
Fix disappearing dropdown

johnnyporkchops:
## Summary (required)
Currently, once you choose all the checklist items in a filter dropdown, the dropdown disappears. Although there is a legitimate UX argument for this behavior, the fact that it does not reappear in the  unlikely (but possible edge-case ) event that the user removes all chosen filters means the user ultimately no longer has access to those filters. This PR makes the dropdowns persist, regardless of filters chosen.

Resolves: #2117

fec/static/js/modules/dropdowns.js

## How to test:
- Checkout and un this branch
- Make sure a checklist-dropdown does not disappear after choosing all the items and  adding them to the selected filters list above.
- http://127.0.0.1:8000/data/disbursements/ is a good one to test because it has a two-item dropdown.

# Screenshots
dropdown-disappear

### DIFF
--- a/fec/fec/static/js/modules/dropdowns.js
+++ b/fec/fec/static/js/modules/dropdowns.js
@@ -228,7 +228,6 @@ Dropdown.prototype.selectItem = function($input) {
         .focus();
     }
   } else {
-    this.removePanel();
     this.$selected.find('input[type="checkbox"]').focus();
   }
 };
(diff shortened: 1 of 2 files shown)


johnnyporkchops:
@JonellaCulmer, On a geekier note--Interestingly this was not actually a bug, or a broken behavior. It was intended to work that way and there were even JS tests to confirm it was removed when no more choices were available. I can see the UX argument for it, but also wonder if it was just not thought all the way through to consider edge cases of user interaction.

JonellaCulmer:
Okay, gotcha. Then you're right. It wasn't built to afford users the opportunity to change their mind and make different selections and go back again. Requiring users to refresh a page to make filters available again doesn't seem the best use of that kind of functionality.

PR #2502, opened 2018-11-09
Fix districts map zoom-to-state

### LINKED ISSUE #2447: Election search map not zooming in on selected states properly
**What we're after:** The election search map is one way that users can filter and search for elections within a particular state or district. To help users accomplish that goal, the map should zoom in on relevant areas. 

However, when selecting specific states from the dropdown list, the zoom in function is not working appropriately, sometimes zooming out instead of zooming in. 

**What it looks like when it's working:** 
screen shot 2018-10-15 at 11 30 44 am

**What it looks like when it's _not_ working:**
screen shot 2018-10-15 at 11 30 54 am

On first select the state appears correctly, but then each subsequent state selection is effected. It doesn't appear that the glitch or bug is affecting the same "states" consistently, but here are some that were observed having this issue:
- American Samoa
- Arizona
- Arkansas
- Colorado
- Delaware
- District of Columbia
- Florida
- Guam
- Hawaii
- Idaho
- Iowa
- Kentucky
- Maine
- Maryland
- Massachusetts
- Michigan
- Minnesota
- Mississippi
- Missouri
- Montana
- Nevada
- New Hampshire
- New Jersey
- New Mexico
- North Carolina
- North Dakota
- Northern Mariana Islands
- Oklahoma
- Oregon
- Pennsylvania
- Puerto Rico
- South Carolina
- South Dakota
- Texas
- U.S. Virgin Islands
- Vermont
- Virginia
- Washington
- Wisconsin
- Wyoming

**Completion criteria:**
- [ ] Check across browser types including but not limited to Chrome, IE, and Firefox
- [ ] If external library is in use, check to make sure we're using the latest version
- [ ] Debug first by going to page, toggle between different states to see if there are any JS console errors
- [ ] Investigate and correct the flawed and inconsistent zooming function on the map
- [ ] Add to release notes

johnnyporkchops:
## Summary
Selecting states on /data/elections was inconsistent, sometimes zooming in, sometimes out

- Resolves #2447

modified:   fec/static/js/modules/election-map.js
This solution uses the leaflet.js flyTo() function, using `this.overlay.getBounds(),this.opts` as arguments,  instead of fitBounds().

## Screenshots
fly_to_zoom

## How to test
  - Checkout and run this branch locally: `feature/fix-districts-map-zoom`
 - `npm run build`
  - go to http://127.0.0.1:8000/data/elections/ and test selecting states and districts to ensure that the map zooms into each chosen state and/or districts without zooming way out like it is on production.

### DIFF
--- a/fec/fec/static/js/modules/election-map.js
+++ b/fec/fec/static/js/modules/election-map.js
@@ -142,7 +142,9 @@ ElectionMap.prototype.updateBounds = function(districts) {
   if (rule) {
     this.map.setView(rule.coords, rule.zoom);
   } else if (districts) {
-    this.map.flyToBounds(this.overlay.getBounds(), { duration: 0.15 });
+    window.setTimeout(() => {
+      this.map.flyToBounds(this.overlay.getBounds(), { duration: 0.5 });
+    }, 500);
   }
 };
 
--- a/fec/fec/static/js/modules/election-map.js
+++ b/fec/fec/static/js/modules/election-map.js
@@ -141,7 +141,7 @@ ElectionMap.prototype.updateBounds = function(districts) {
   if (rule) {
     this.map.setView(rule.coords, rule.zoom);
   } else if (districts) {
-    this.map.fitBounds(this.overlay.getBounds());
+    this.map.flyToBounds(this.overlay.getBounds(),this.opts)
   }
 };
 
--- a/fec/fec/static/js/modules/election-map.js
+++ b/fec/fec/static/js/modules/election-map.js
@@ -141,7 +141,7 @@ ElectionMap.prototype.updateBounds = function(districts) {
   if (rule) {
     this.map.setView(rule.coords, rule.zoom);
   } else if (districts) {
-    this.map.flyToBounds(this.overlay.getBounds(),this.opts)
+    this.map.flyToBounds(this.overlay.getBounds(), this.opts);
   }
 };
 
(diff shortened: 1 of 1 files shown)


johnnyporkchops:
@JonellaCulmer. This solution addresses the problem in the issue, and I think makes the map less cumbersome to navigate for users. Although the time it takes to zoom between states that are far apart is an issue, it still seems better than the current implementation, and requires fewer futile clicks.

The time in MS that it takes to zoom from Florida to Washington , for example,  is probably equal to or less than the time it takes on the current map to do this , by either:
1) Zooming into a state that is state on the map by clicking the plus button and also panning with you mouse to get it centered. 
--OR--
2) Clicking to zoom into that state and then having to zoom back out because it takes you to the district you happened to have clicked instead of the whole state view

JonellaCulmer:
@johnnyporkchops Thanks for the additional information. 
Would love to hear what @AmyKort and @jwchumley think about this solution.

AmyKort:
@JonellaCulmer @johnnyporkchops @jwchumley That solution works for me.

fec/fec/static/js/modules/election-map.js
```diff
@@ -125,13 +125,15 @@ ElectionMap.prototype.drawDistricts = function(districts) {
   }).addTo(this.map);
   this.updateBounds(districts);
   this.drawBackgroundDistricts(districts);
+  //this.overlay.bindPopup('popupContent').openPopup();
```
patphongs: Looks like this is leftover commented out code that should be removed

johnnyporkchops: oops, comment removed

JonellaCulmer:
Still looks great. Merging.

PR #2582, opened 2018-12-11
[WIP] Customize record, tips, and release feed order

### LINKED ISSUE #1694: Ensure Record articles appear in feeds with the most recent article on top
In recent weeks, there were a couple of occasions when more than one Record news article was published on a given day. On those days, the article that was published earliest in the day ended up being the one at the top of the feed. While some of the feeds have since been tweaked to make the last article published the top one on the feed, not all of them have been.

For example: The home page feed and the [legal resources "legal news" feed](https://www.fec.gov/legal-resources/) have been tweaked, with the last article posted showing up first (which is what we're after).
legalresources-recordfeed1

On the other hand, the Record page at https://www.fec.gov/updates/?update_type=fec-record and the feed on the latest updates page at https://www.fec.gov/updates/ show Record articles appearing in the order they were posted on a given day, but with the earliest article posted above articles posted later in the day.
recordpage-feed

To ensure consistency, we'd like to make sure all feeds for Record news articles show the articles with the most recent article posted appearing at the top.

Completion criteria
- [ ] Ensure that all feeds of Record news articles are configured so that the most recent article posted is on top, and if more than one article is posted on a given day, so that articles are posted with the last one posted showing first (similar to the first screencap above).

patphongs:
## Summary

- Resolves #1694 
_This updates record articles, tips, and releases to all use DateTimeField so that a time can be specified and the order can be based on that instead of just date._

- Getting an error on the updates page "datetime.datetime to datetime.date". Need to see how to convert that
- It would be good to check if this will effect the date fields for these articles, because we are introducing a time element that was not previously there.

- Homepage - News and announcements section
- Service landing page - Tips feed
- Updates page

## How to test
1. `./manage.py makemigrations`
2. `./manage.py migrate`
3. Change the date field of record articles, tips, and releases to see if they shift in the homepage feed and


patphongs:
This is outdated and I haven't been able to revisit this issue. Closing for now

Issue #2066, opened 2018-06-06
Raising and spending breakdown back button issue


patphongs:
When on the raising or spending breakdown charts with anything other than presidential selected. If you click on a candidate and navigate away from the spending and raising breakdown page and then press the back button, the office displayed in the dropdown does not match the actual data that is displayed on the chart.

https://www.fec.gov/data/raising-bythenumbers/
https://www.fec.gov/data/spending-bythenumbers/

![breakdown-bug](https://user-images.githubusercontent.com/12799132/41065071-756eff68-69ab-11e8-97de-2594eed84f66.gif)

AmyKort:
In re-implemented chart, back button doesn't function in this way. Closing.

JonellaCulmer:
@AmyKort I'm looking at the by the numbers pages and this issue and I'm still observing this issue. So reopening.

PR #2634, opened 2019-01-29
Raising / spending back button fix

rfultz:
Added functionality to reflect the user's current choices in the URL, and restore those when using the browser history to return to the page.

## Summary
The pull-downs' states weren't being pushed to the URL so changing the cycle or office, clicking a candidate, then going back to the previous page (from the candidate details) reloaded the default version of the raising or spending data and didn't reflect the user's most recent settings.

Changing either the cycle or office pull-downs now pushes those updated values to the URL.

- Resolves [#2066](https://github.com/fecgov/fec-cms/issues/2066)

`top-entities.js`:
- I added `uri.js` to handle url updates.
- added `office` to `baseQuery` in `init`
- borrowed `pushStateToURL()` from elsewhere; would like to streamline
- added `this.pushStateToURL()` to both `handleCycleChange()` and `handleCategoryChange()`

`views.py`:
- near-identical updates to both `raising()` and `spending()`
- now overriding `top_category` if the URL contains a (valid) value for `list`
- since cycle can now come from the URL, forced it/them to `int`

## Screenshots
Check out the URL changes:
default
with new url vars

## How to test
- Go to http://localhost:8000/data/raising-bythenumbers
- Change the cycle (remember what it was and how you changed it)
- Change the office pull-down (remember what it was and how you changed it)
- Look at the list of candidates (remember general names, order, and values)
- Click a candidate's name
- Go back (browser back button, keyboard cmd ⬅️, Android ⬅️, etc)
- Verify the cycle, office, and candidates are exactly how they were when you clicked the candidate's name
- ~~*Bonus*: change both the cycle and office values multiple times, click a candidate, and then go back several times to make sure each step through history is correct~~ Coming in a following ticket
- Do the same for http://localhost:8000/data/spending-bythenumbers/?cycle=2012&list=P

TESTING CONCERNS:
I'd like to make sure this change doesn't break something anywhere else these components are used. Where all might that be?
____

### DIFF
--- a/fec/data/views.py
+++ b/fec/data/views.py
@@ -34,6 +34,8 @@ report_types = {
     'I': 'ie-only',
 }
 
+validListUrlParamValues = ['P', 'S', 'H', 'pac', 'party']
+# INITIALLY USED BY raising() AND spending() FOR VALIDATING URL PARAMETERS, THE list URL PARAM
 
 def to_date(committee, cycle):
     if committee['committee_type'] in ['H', 'S', 'P']:
@@ -503,10 +505,8 @@ def elections(request, office, cycle, state=None, district=None):
 def raising(request):
     top_category = request.GET.get('top_category', 'P')
     
-    # DOES THIS KIND OF LIST EXIST ELSEWHERE?
-    validListValues = ['P', 'S', 'H', 'pac', 'party']
     # IGNORING INVALID list URL PARAMETERS
-    if request.GET.get('list') and request.GET.get('list') in validListValues:
+    if request.GET.get('list') and request.GET.get('list') in validListUrlParamValues:
         top_category = request.GET.get('list')
         # IF A VALID list VALUE EXISTS, WE'LL LET IT OVERRIDE top_category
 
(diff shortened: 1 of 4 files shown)


fec/data/views.py
```diff
@@ -500,11 +500,18 @@ def elections(request, office, cycle, state=None, district=None):
         },
     )
 
-
 def raising(request):
     top_category = request.GET.get('top_category', 'P')
+    
+    # DOES THIS KIND OF LIST EXIST ELSEWHERE?
+    validListValues = ['P', 'S', 'H', 'pac', 'party']
```
patphongs: So within /data/constants.py we define a lot of constants there. I did see one for [`pac_party`](https://github.com/fecgov/fec-cms/blob/125e190f3485427190a13f48dbf7429afc3c10fd/fec/data/constants.py#L228) and [`house_senate_types`](https://github.com/fecgov/fec-cms/blob/125e190f3485427190a13f48dbf7429afc3c10fd/fec/data/constants.py#L245) within that file. But not necessarily for 'P'. I wonder if we should change `house_senate_types` to `candidate_office_types` to add P in there. And then fix the logic for where house_senate_types is currently being called.

rfultz: We could… Should we?
We could spin up a tech debt ticket to handle (and test) that, too. I'm not sure how tangled that might get.

lbeaufort: @rfultz I would at least share the same list between the `raising` and `spending` functions.

rfultz: @lbeaufort, I've moved the list outside of `raising` and `spending` so they both reference it.

Issue #2644, opened 2019-01-31
Map tooltip does not contain text


patphongs:
When hovering over the maps on candidate and election profile pages, the tooltip no longer contains the text within it. 

https://www.fec.gov/data/candidate/S2MA00170/?cycle=2018&election_full=true&tab=raising
https://www.fec.gov/data/elections/senate/MA/2018/?tab=election-data#individual-contributions

<img width="1057" alt="screen shot 2019-01-31 at 3 59 34 pm" src="https://user-images.githubusercontent.com/12799132/52085124-a7e75a80-2571-11e9-80e0-e1ffa6bd46fb.png">

patphongs:
PR merged today, closing

PR #2574, opened 2018-12-06
Update boolean search box and add tooltips for AO,MUR, Statutes 

johnnyporkchops:
## Summary
Make design/layout edits to keyword modal html and jinja templates
Add tooltip for keyword text-box field on AO, MUR, Statues
 - AO Search is React.js, so I made the tooltip in here, a reusable react component (Could now be 
    used on any filter type inside AO search with custom tooltip message and positioning classes)
 - MUR and Statutes datatables are both  rendered from legal.jinja, so the tooltip is hardcoded there

- Resolves #2283

## Impacted areas of the application
AO Search react app
	- modified:   fec/static/js/legal/Filters.js
	- modified:   fec/static/js/legal/filters/TextFilter.js
	- added:       fec/static/js/legal/filters/TooltipHelp.js
MUR and Statutes datatables:
	- modified:   legal/templates/macros/legal.jinja
        - modified:   fec/static/scss/components/_tooltips.scss
Keyword Modal popup:
        - modified:   home/templates/partials/legal-keyword-modal.html
	- modified:   legal/templates/partials/legal-keyword-modal.jinja
        - modified:   fec/static/scss/components/_search-controls.scss
        - modified:   fec/static/scss/components/_modals.scss

### Tested in Internet Explorer below, per checklist item 5 in issue:
gif2
<hr>

## How to test
Checkout this branch locally: feature/update-boolean-search-box
NPM run build
Runserver and test the tooltips on Keyword text filter on AO, MUR and Statutes searches
Test keyword modal box on the same pages as well as the legal landing page (this uses a different template)
If one were ambitious they could also test the reusable tooltip react component by adding it to `DateFilter.js`* by:
- Copying the tooltip code from the `TextFilter.js` file:
  - Lline 3 
  - Lines 21-27 
  - Line 69 and line 79
...and pasting it into `DateFilter.js`.
- Then enable it in` Filters.js`  for that filter-type (DateFilter), like I did for the  TextFilter (@line 56). Maybe add a different message.
- Don't forget to  run build,  and see if it works! ( I tested it but would still like to see if someone else thinks it works well)

*For filters that depend on other filters, like CheckboxFiler, the process is different, but it can still be added to any filter in the panel.

PR #2657, opened 2019-02-05
Restyle tooltip so content is visibly contained within

rfultz:
## Summary
Being absolutely positioned, the tooltip was ignoring the size of its content.

- Resolves # [2644](https://github.com/fecgov/fec-cms/issues/2644)

The changes should only affect the map tooltips, and only with tooltips that appear above the pointer but we should make sure we check maps and tooltips elsewhere on the site.

`fec/fec/static/js/modules/maps.js`:

- Expanded js to set tooltip height to accommodate that of its content + padding.
- Adjusted the vertical distance between the tooltip point and the pointer so the tooltip no longer appears under the pointer.

`fec/fec/static/scss/components/_tooltips.scss`:

- Adjusted tooltip `width` settings to be `min-width` so wider content (like "Pennsylvania") don't overflow the right edge of the tooltip
- Horizontally centered the downward arrow (it has two pieces, the `:before` and the `:after`) under the tooltip rather than leave it hard-coded a specific distance from the left
- Removed a duplicated chunk of code for `.tooltip--above &::before` (and `after`)
- Decreased the `line-height` of the larger/title text in the tooltip from 4rem to 2rem because the distance between title lines was greater than the distance between the title and the value/text under it

## Screenshots
Current (with the pointer over the northwestern point of Washington)
before
Proposed (with the pointer in the same place)
after

## How to test
- Find a [candidate page on localhost](http://localhost:8000/data/candidate/S2MA00170/?cycle=2018&election_full=true&tab=raising)
- Choose Raising from the left nav
- Scroll down to Individual Contributions
- Choose the State tab
- Move your cursor over the map, back and forth over the states, especially those with names of different widths (I think Iowa and Massachusetts are the extremes)
- The tooltip text should appear inside the tooltip box and the widths and tooltip position should adjust properly
- **Check the maps and tooltips elsewhere on the site, including tooltips with two-line titles**
- **Double-check functionality on IE, older browsers, and touch devices**
- **Test in Safari after it's on dev**

____

### DIFF
--- a/fec/fec/static/js/modules/maps.js
+++ b/fec/fec/static/js/modules/maps.js
@@ -185,7 +185,17 @@ function stateTooltips(svg, path, results) {
 function moveTooltip(tooltip) {
   var x = d3.event.pageX - tooltip[0][0].offsetWidth / 2;
   var y = d3.event.pageY - tooltip[0][0].offsetHeight;
-  tooltip.style('left', x + 'px').style('top', y + 'px');
+
+  var bottomPointerHeight = '.8rem';
+
+  var contentHeight = $('#map-tooltip .tooltip__title').innerHeight();
+  contentHeight += $('#map-tooltip .tooltip__value').innerHeight();
+  contentHeight += 30; // (padding)
+
+  tooltip
+    .style('left', x + 'px')
+    .style('top', 'calc(' + y + 'px - ' + bottomPointerHeight + ')')
+    .style('height', contentHeight + 'px');
 }
 
 function highlightState($parent, state) {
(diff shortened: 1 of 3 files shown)


johnnyporkchops:
@rfultz , Awesome job on getting these map tooltips to position correctly! The tooltip styles I added to flesh out  the `tooltip.scss` stylesheet in PR #2574 changed some styles on tooltip classes for the calendar. Since this PR (#2657) makes some changes there, I thought I would go ahead and add one line to your commit for that stylesheet rather than create a whole new PR.
### Current Calendar Tooltips:
screen shot 2019-02-05 at 10 57 39 pm
<hr>

### Fixed Calendar tooltip:
screen shot 2019-02-05 at 10 58 23 pm

PR #2702, opened 2019-02-26
Remove double slashes from spending urls

lbeaufort:
## Summary (required)

- Resolves issue found in testing: **Remove double slashes from spending urls**
- `helpers.buildUrl` is expecting the format `['elections', 'search']`
- URL was being formatted as localhost:5000/v1//schedules...

"Spending by others to support or oppose" tab: http://localhost:8000/data/candidate/P80001571/?tab=other-spending&cycle=2016&election_full=false

## Before
screen shot 2019-02-26 at 3 19 46 pm

## After
screen shot 2019-02-26 at 3 18 32 pm

### DIFF
--- a/fec/fec/static/js/modules/other-spending-totals.js
+++ b/fec/fec/static/js/modules/other-spending-totals.js
@@ -8,9 +8,9 @@ var _ = require('underscore');
 var helpers = require('../modules/helpers');
 
 var pathMap = {
-  independentExpenditures: '/schedules/schedule_e/by_candidate/',
-  communicationCosts: '/communication_costs/by_candidate/',
-  electioneering: '/electioneering/by_candidate/'
+  independentExpenditures: ['schedules', 'schedule_e', 'by_candidate'],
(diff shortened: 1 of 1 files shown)


patphongs:
> The code is fine. I'm curious what changed (and broke this) and if we should look for this issue occurring anywhere else.

@rfultz I'm not sure if this has always been a problem, but I saw the double slashes show up through all the different environments [prod, stage, dev]. But prod is somehow more forgiving? Not sure why though

lbeaufort:
Thanks @patphongs and @rfultz! 

Robert - I looked at all the places we're using `helpers.buildUrl` and they all looked like they were using the proper format.

Pat - I wonder if the API umbrella is handling the `//` in URLs. It looks like `stage` and `prod` handle the `//` but `dev` doesn't. `Dev` doesn't use the API umbrella and the others do.

lbeaufort:
@patphongs I confirmed the API umbrella removes extra slashes: https://github.com/NREL/api-umbrella/blob/master/test/proxy/test_url_special_characters.rb#L25-L34

Issue #2671, opened 2019-02-07
Update Election Data and other pages that are linking to the wrong pages for 2019-2020 deadlines


dorothyyeager:
**What we're after**
I'm using the "compare candidates" pages, and noticed the links to the various dates and deadlines pages. The top box is correct. However, when clicking the buttons for specific items after each section, they are pointing to transition pages for 2020. Starting with this year, we are no longer putting these pages in transition, and instead are using Wagtail pages, so those links need to be changed as they return a 404.

For an example, see https://www.fec.gov/data/elections/president/2020/. However, every single page that comes out of this for 2020 most likely needs modification (or the template that generates these pages).

![image](https://user-images.githubusercontent.com/24437369/52441372-69f8b200-2aee-11e9-8263-7e5389f08bfb.png)
This button needs to point to https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/ 

![image](https://user-images.githubusercontent.com/24437369/52441453-9b717d80-2aee-11e9-9a87-7c3f23025ef2.png)

This button needs to point to https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/2019-reporting-dates/24-and-48-hour-reports-independent-expenditures-2019-20/ (or alternatively, you could just point to dates and deadlines and not have to edit the link each year).

![image](https://user-images.githubusercontent.com/24437369/52441504-c6f46800-2aee-11e9-9289-fd55f927fd92.png)
This button needs to point to https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/2019-reporting-dates/electioneering-communications-periods-2019/  (or alternatively, you could just point to dates and deadlines and not have to edit the link each year).

Likewise, there are many pages reference the compliance map; however, the map was not updated for 2019 and older links to it just return a dead map. The Dates and deadlines page has the same information, and contains archived information for past years.

**To do:**
- [ ] Replace the hard-coded links at the bottom of the election data pages to link to the Dates and Deadlines page at https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/ - front end work
- [x] Links to the Compliance Map (https://transition.fec.gov/info/ElectionDate/index.shtml) and its subdirectories and files within (/info/ElectionDate/) from past years should be redirected to  https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/ (Front end or content team work) - This work has been moved to #2699

dorothyyeager:
[...]

PR #2716, opened 2019-03-04
Change button hrefs from transition to dates-and-deadlines

rfultz:
## Summary
- Resolves # [2671](https://github.com/fecgov/fec-cms/issues/2671)
Updated the button links on https://www.fec.gov/data/elections/president/2020

Only the Jinja template was updated but it may affect https://www.fec.gov/data/elections/president/*. Luckily, the links were made more generic so as not to require updates every year. Also, the buttons no longer link to transition.

## Screenshots
Skipping the screenshots as there was no visible change, only URL changes
Some shots showing the URLs in the address bar and the button HREF in the status bar at the bottom of the window:
image
image
image
image
image
image
image
image
image
image
image

## Related PRs
No related PRs, though there will be additional work to add the redirects specified in the To-do list in the ticket.

## How to test
For these pages, verify the dark blue buttons' href values go to the new address—both the subdomain and the paths.
- http://localhost:8000/data/elections/president/2020/
- http://localhost:8000/data/elections/house/VA/11/2020/
- http://localhost:8000/data/elections/house/MD/05/2020/
- http://localhost:8000/data/elections/senate/VA/2020/
(any district/area will work—these are just examples)
____

### DIFF
--- a/fec/data/templates/partials/elections/election-data-and-compliance-tab.jinja
+++ b/fec/data/templates/partials/elections/election-data-and-compliance-tab.jinja
@@ -85,7 +85,7 @@
                 'Quarterly / monthly filing and pre-/post-election reports',
                 [{
                   'text': 'Reporting deadlines and notices',
-                  'href': 'https://transition.fec.gov/info/report_dates_'+cycle|string+'.shtml'
+                  'href': 'https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/'
                 }]
               )
             }}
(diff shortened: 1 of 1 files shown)


rfultz:
> This is the correct URL to redirect the old transition links to.
> 
> I don't have a local set up so couldn't review that way.
> 
> Last thing to check - this affected pages for House/Senate candidates too, so we'd want to ensure that this code change was put in place for their election data pages as well. For example, here's the VA Senate page - https://www.fec.gov/data/elections/senate/VA/2020/ There's probably a page like this for each Senate and House race, so we'll want to ensure the link is modified on all of them.

Oh… Then this ticket isn't ready yet. Any chance of getting a full list of URLs that need to be updated and reviewed?

It looks like /data/elections/senate/VA/2020/ needs updates to the "Convention and Primary deadlines" and "General deadlines" buttons under "Candidate financial totals" but the other two look right (on localhost).

PR #2725, opened 2019-03-07
[DO NOT MERGE - feature space only] Test/raising spending bythenumbers update

patphongs:
This is the branch that is currently deployed on our feature space for user testing of raising/spending by the numbers pages.

### DIFF
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -347,12 +347,13 @@ def result_or_404(data):
     return data['results'][0]
 
 
-def load_top_candidates(sort, office=None, cycle=constants.DEFAULT_TIME_PERIOD, per_page=5):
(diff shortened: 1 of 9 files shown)


JonellaCulmer:
@patphongs Just a couple things we probably want to fix before this is deployed. 

Since we're removing the line charts from the data landing page, we should relink the Spending and Raising overview links at the tops of the left-nav to instead go to the Raising and Spending line charts on the advanced data pages. If we can use an anchor to those spots, that is. If not the overall page should do. Or we can remove. But I like there'd be a way to get back to Raising and Spending pages, at least for now.

And second, we briefly discussed changing the coverage dates from the 2-4-6 year periods to the current through dates of the data we have. Can we revisit that?

PR #2752, opened 2019-03-19
Use cache key prefix for finer control of cache

lbeaufort:
I believe the build on `stage` was weird because Circle was caching dependencies. This change forces Circle to get fresh dependencies for the production deploy.

Use cache key prefix for finer control of cache. See the tip under https://circleci.com/docs/2.0/configuration-reference/#save_cache: "Given the immutability of caches, it might be helpful to start all your cache keys with a version prefix v1-.... That way you will be able to regenerate all your caches just by incrementing the version in this prefix."

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,9 +37,9 @@ jobs:
 
       - restore_cache:
           keys:
-          - fec-cms-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "package.json" }}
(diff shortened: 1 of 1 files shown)


patphongs:
Looks like the CMS is functioning and javascript is working again. Still investigating why that is, it's possible that rebuilding without cache must have fixed it but we didn't notice until after our browser cache cleared?

PR #2834, opened 2019-04-15
Add download only API key to CMS 

### LINKED ISSUE #2833: Use a downloads-only key for datatable exports
Follow-up from https://github.com/fecgov/openFEC/issues/3651

So we can more closely monitor and limit usage of the downloads endpoint, we should make a separate `FEC_DOWNLOAD_API_KEY` that is used only for exports for the CMS. I've already created these keys in the API umbrella admin interface.

patphongs:
## Summary

- Resolves #2833 
- Add FEC_DOWNLOAD_API_KEY as environment variable for data templates
- Add DOWNLOAD_API_KEY as a parameter to the download query
- Remove default api_key if a separate api_key is defined in the query

-  Export button when querying /download/ API calls

## Related PRs

branch | PR
------ | ------
feature/limit-downloads-access | https://github.com/fecgov/openFEC/pull/3695

## How to test
1. Set FEC_DOWNLOAD_API_KEY as environment variables
2. Test exporting a dataset like the candidates data: http://localhost:8000/data/candidates/. The `/download/` api call should only contain the `api_key` defined inside FEC_DOWNLOAD_API_KEY
3. Make sure all other client side API calls still uses the default key defined in FEC_WEB_API_KEY

### DIFF
--- a/fec/data/templates/layouts/main.jinja
+++ b/fec/data/templates/layouts/main.jinja
@@ -20,6 +20,7 @@
     API_LOCATION = '{{ FEC_API_URL }}';
     API_VERSION = 'v1';
     API_KEY = '{{ FEC_API_KEY }}';
+    DOWNLOAD_API_KEY = '{{ FEC_DOWNLOAD_API_KEY }}';
     DEFAULT_TIME_PERIOD = '2018';
     START_YEAR = '1979';
     END_YEAR = '2018';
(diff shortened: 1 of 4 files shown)


lbeaufort:
@patphongs noting here that we need to update the key `env var` in production. It's probably best to do that before the deploy

PR #2847, opened 2019-04-18
Consistent  menu bar indicator

### LINKED ISSUE #2760: Menu bar indicators need to be fixed on website
**What we're after:** It has come to our attention that the menu bar indicators for the buckets are not remaining up consistently for all sections. 
- [x] Campaign finance data (blue/teal bar) - should show blue bar once clicked
- [x] Help for candidate and committees (red/orange bar) - should show red once clicked 
- [x] Legal resources (blue/teal bar) - should show blue once clicked
- [x] About (red/orange bar) - should show red bar once clicked

To do: Needs to be fixed to be consistent

johnnyporkchops:
## Summary
Make menu bar indicators consistent site-wide

- Resolves #2760

modified: home/models.py
modified: partials/navigation.html
modified: home/views.py
modified: data/views.py

## How to test
- checkout `feature/2760-fix-menu-bar-indicator` or see `https://fec-feature-cms.app.cloud.gov`
- Test that menu bar indicator shows the proper parent section for all pages
- A few pages at the root level that have no parent section and will not get indicator: Example: `test-- new-features-our-design-team/` and `/freedom-information-act/`

### DIFF
--- a/fec/home/models.py
+++ b/fec/home/models.py
@@ -83,7 +83,6 @@ def get_content_section(page):
     }
 
     ancestors = page.get_ancestors()
-
     content_sections = [
         slugs.get(ancestor.slug) for ancestor in ancestors
         if slugs.get(ancestor.slug) != None
@@ -92,6 +91,17 @@ def get_content_section(page):
         return content_sections[0]
     else:
         return ''
+
+
+
+    # content_sections = [
+    #     slugs.get(ancestor.slug) for ancestor in ancestors
+    #     if slugs.get(ancestor.slug) != None
+    # ]
+    # if len(content_sections):
+    #     return content_sections[0]
+    # else:
+    #     return ''
 
 class UniqueModel(models.Model):
     """Abstract base class for unique pages."""
@@ -127,7 +137,8 @@ class ContentPage(Page):
         index.SearchField('body')
     ]
 
-    #Default content section for determining the active nav
+    """Returns no content section so the active nav can be set in the page-type \
+    that extends the ContentPage model """
     @property
     def content_section(self):
        return ''
@@ -354,7 +365,7 @@ class RecordPage(ContentPage):
 
     @property
     def content_section(self):
-        return 'about'
+        return get_content_section(self)
 
     @property
     def get_update_type(self):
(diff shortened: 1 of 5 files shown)


johnnyporkchops:
@kathycarothers , @dorothyyeager, Thereare a few pages at the root level that do not have a parent section and will not get a menu br indicator. Examples: `test-new-features-our-design-team/` and `/freedom-information-act/`.
For the FOIA page, it is titled `About:FOIA` in the Wagtail explorer , but is not inside the about section in wagtail, I assume this was intentional ? Or should we move it?
For other pages like `test-new-features-our-design-team/` and any future pages created at the `home (root)` level, we could add logic to make them all default to 'about', but I don't think it's  necessary.

dorothyyeager:
Hi @johnnyporkchops - I asked on Slack, and there was no particular intent for why it was put in Wagtail under the root rather than the /about/. It's fine to move it, and I tried, BUT that caused a hard-coded link on the home page to break. My inclination is to leave it where it is (I moved it back so that the link on home page will work). I think for other random pages, we should not have  a default setting, but definitely need to be careful when creating them to make sure they are in the right directory.

It was good you asked about it -- there were some other issues on the FOIA page that we fixed, like updating contact information and adding external buttons where needed. The page looks better now!

fec/home/models.py
```diff
@@ -88,6 +92,17 @@ def get_content_section(page):
     else:
         return ''
 
+
+
+    # content_sections = [
```
rfultz: If these are commented, should we just remove them?

johnnyporkchops: Thanks @robert, I will remove

johnnyporkchops:
@dorothyyeager -- Yet one more question: For `intro to campaign finance`  section. Does this section claim its parent as `Campaign Finance Data` or will it one day have its own top level menu item? I can give it a parent of  `Campaign Finance Data` in this PR if we want or just leave it with no indicator. Here is how a page there would look. (BTW, I have an older version of this page on my local)


intro-camp_finance-parent-indicatorsA

dorothyyeager:
@johnnyporkchops No need. Its parent page is https://www.fec.gov/introduction-campaign-finance/. We aren't linking that particular page on the home page top menu at this point so no need for an indicator (plus it might be confusing, since 1/3 of that section is data-related, 1/3 is H4CC-related and 1/3 is public records and non-FEC election info). Thanks for checking.

PR #2852, opened 2019-04-22
Handle over-limit downloads

### LINKED ISSUE #2842: Handle over rate limit message for downloads
https://github.com/fecgov/openFEC/issues/3651

API message for over rate limit is code 429. Test with more than one request with this key: 
`curl -I "https://api.open.fec.gov/v1/candidate/P80001571/totals/?api_key=8Bhfz2eaOM0KVmfulpzCUgLnTeYn60oJQDZp4tzy"`

```
HTTP/1.1 429 Too Many Requests
Server: openresty
Date: Wed, 17 Apr 2019 15:04:14 GMT
Content-Type: application/json
Connection: keep-alive
X-RateLimit-Limit: 1
X-RateLimit-Remaining: 0
Access-Control-Allow-Origin: *
X-Cache: MISS
Strict-Transport-Security: max-age=31536000; preload
```

lbeaufort:
## Summary

- Resolves #2842 
_Add specific 429 over rate limit error message when a post request is received. This will only work on first 429 error with a POST xhr call. Otherwise, print a generic error message that also warns of over rate limits when there is an error._

Also see this detailed explanation about error message handling for downloads: https://github.com/fecgov/fec-cms/issues/2842#issuecomment-497367683

-  Datatable page downloads

## Screenshots
Below you will see that the initial error may give you the 429 error message. The subsequent request gives you the generic error message.

Screen Shot 2019-05-30 at 12 31 04 PM

## How to test
- Check out this branch and `npm run build`
- Query stage space for FEC_DOWNLOAD_API_KEY and set it in your environment
- Change FEC_API_URL to point to stage
- `./manage.py runserver`
- Check a datatable page, like candidates http://localhost:8000/data/candidates
- Do a bunch of exports until you hit the rate limit for the stage API. Right now it's set to 25 hits per IP per hour.
- Check error messages and make sure you're getting either the updated generic error message, seen in the screenshot above, or the exceeded limit error message, also seen in screenshot above.
____

### DIFF
--- a/fec/fec/static/js/modules/download.js
+++ b/fec/fec/static/js/modules/download.js
@@ -194,7 +194,7 @@ DownloadItem.prototype.handleServerError = function(xhr) {
   } else {
     this.$body.html(
       '<div class="message message--alert message--mini">' +
-        'Sorry, there was a server error. Please try again later.' +
+        'Sorry, you have either exceeded your maximum downloads for the hour or encountered a server error. Please try again later.' +
         '</div>'
     );
   }
--- a/fec/fec/static/js/modules/download.js
+++ b/fec/fec/static/js/modules/download.js
@@ -157,7 +157,7 @@ DownloadItem.prototype.handleSuccess = function(response) {
 
 DownloadItem.prototype.handleError = function(xhr, textStatus) {
   if (textStatus === 'error') {
-    this.handleServerError();
+    this.handleServerError(xhr);
   } else if (textStatus !== 'abort') {
     this.schedule();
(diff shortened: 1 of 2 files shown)


patphongs:
Thanks @rfultz. The CORS error will happen regardless of environment. Reason being is because we're sending OPTIONS HTTP requests methods as a preflight request. I think our API umbrella does not allow for this and that's why we're getting those errors. We'll need to figure out a way to allow OPTIONS request methods I think in order to get rid of this CORS error. So for now, let's save that for future investigation.

PR #2848, opened 2019-04-19
[MERGE WITH GITFLOW] Hotfix for rendering bugs and responsive issues with home bar charts

johnnyporkchops:
## Summary 
This is a hotfix to fix responsive issues and rendering bugs with the homepage bar charts. 

## Impacted areas of the application
Home page raising/spending bar charts

## Screenshots
**Currently on production**

Screen Shot 2019-04-19 at 12 25 54 AM

**This hotfix**

Screen Shot 2019-04-18 at 10 22 55 PM

### Also fixes this:
med-size-bug

## Related PRs
#2835
#2748

## How to test
Checkout and run hot fix branch and test that above bugs are fixed at small/med screen sizes
____

Issue #2588, opened 2018-12-14
Relabel "election" dropdown for raising/spending breakdown pages


lbeaufort:
**User story**
Because we're looking at 2-year totals on the raising/spending breakdowns, we should relabel the "election" dropdown to "two-year-period"

https://www.fec.gov/data/raising-bythenumbers/ 
https://www.fec.gov/data/spending-bythenumbers/
<img width="1054" alt="screen shot 2018-12-14 at 9 55 24 am" src="https://user-images.githubusercontent.com/31420082/50011083-21f35e00-ff89-11e8-8f71-d2e9fb010fb7.png">

**Completion criteria**
- [ ] Relabel "election" dropdown to "two-year-period" on raising breakdown
- [ ] Relabel "election" dropdown to "two-year-period" on spending breakdown

PR #2854, opened 2019-04-23
[MERGE BEFORE RELEASE] Raising-Spending barcharts - disable non-pres election years for Safari-and-IPhone

### LINKED ISSUE #2851: Raising/Spending: by the numbers: Not showing correct election years for Prez. candidates on iPhone
**What we're after:** Just like everywhere else, we need to only see Presidential election years on the raising/spending by the numbers pages. Unfortunately, we're seeing additional election years on iPhone, such as 2014, 2018, and 2022 as options. And it's returning bad data. 

**Completion criteria:**
TBD

johnnyporkchops:
## Summary
- Fixes Raising/Spending barcharts bug on IPhone and Safari desktop which was showing non-presidential years as select option for presidential  searches. If a non-pres years was chosen, it would present the user with  a meaningless, zero-sum bar chart.
-  Adds `disabled(true/false)` property to accommodate Safari Desktop/Mobile's lack of support for hide/show --- or any JS or CSS display property manipulation on `select options`

```
if (optValue % 4 !== 0) {
$(this).hide().prop('disabled', true);
...
else {
$('#election-year option').show().prop('disabled', false);
}
```
So I think the best solution is to `hide/show` and `disable/enable` them. This way Chrome and FF do not show the options, while Safari/IPhone show them as disabled and unclickable.

**BTW:** If we REALLY needed to remove the presidential options completely from view, this actually  works using `jQuery wrap(), unwrap()` but I did not think it was worthwhile to wrap it in them in all browsers just to accommodate one:
```
if (optValue % 4 !== 0) {
$(this)
.hide()
.wrap( "<span style='visibility:hidden'></span>" )

    ...

else {
$('#election-year option')
.show()
.unwrap("span")
}

(we have to keep the show/hide in here to trigger the default-to-2020 expression when switching from H/S to P while on non pres year)
```
<hr>
Because of Safari's vigilant adherence to  its user-agent stylesheets for select options, we have no ability to  enhance UX by adding conditional CSS to more clearly delineate visible, yet disabled options. 

- Resolves #2851

modified:   fec/static/js/modules/top-entities.js_
Impacts stand-alone and home raising/spending barcharts 

### IPhone:
ezgif com-resize
### Safari:
Apr-22-2019 22-46-55

## Related PRs
#2848 : Hotfix for rendering bugs and responsive issues with home bar charts 

## How to test
Checkout feature/disable-non-pres-elections-choice-safari-and-iphone
Run npm run buld
Test in Safari to make sure that non-pres-years are disabled for presidential searches and available for house/senate searches

Note:  I tested this on IPhone by pointing iPhone to local servers IP address and temporarily setting ALLOWED_HOSTS to ['*'] in settings/dev.py. I think you also need to have your Mac and your iphone using the same WIFI network to do this. 
____

### DIFF
--- a/fec/fec/static/js/modules/top-entities.js
+++ b/fec/fec/static/js/modules/top-entities.js
@@ -103,9 +103,11 @@ TopEntities.prototype.updateElectionYearOptions = function(office) {
     // only show presential options
     $('#election-year option').each(function() {
       var optValue = parseInt($(this).val());
-      // hide all of the non-presidential election years
+      // hide all of the non-presidential election years/disable for Safari
       if (optValue % 4 !== 0) {
-        $(this).hide();
+        $(this)
+          .hide()
+          .prop('disabled', true);
       } else {
         // track the nearest future presidential election
         if (optValue > currentYear) {
@@ -122,8 +124,10 @@ TopEntities.prototype.updateElectionYearOptions = function(office) {
         .change();
     }
   } else {
-    // show all options!
-    $('#election-year option').show();
+    // show all options/enable for Safari!
+    $('#election-year option')
+      .show()
+      .prop('disabled', false);
   }
 };
 


rfultz:
I'm digging the animated images!

Safari is objecting to a CORS issue between localhost the dev API; I've created an issue for it ( #2588 )

PR #2942, opened 2019-06-05
Schedule_a datatable search with criteria

### LINKED ISSUE #2811: Implement schedule_a filter interaction changes for all receipts, indiv. contrib.
**User story:** As a data user, I need to be able to increase my time period selection filter larger than a single 2 year time period for receipts and disbursements data so that I can view and filter data for that expanded time span.

**What we're after:** In order to successfully remove the two-year restriction from the all receipts, individual contributions and disbursements data tables, we need to consider what changes to the front-end interaction need to look like as well. 

As an MVP, we're going to make the two-year period search consecutive, and not allow for users to choose random two-year periods. This changes the front-end interaction from its current implementation, although we do hope to retain both the two-year period selection as whole as well as allow for users to select between two year periods. 

Example: User selects2015-2016 and 2017-2018. This search should return results. The user should also be able to refine this search to also specify specific dates with these full two-year periods such as 03/01/2016-10/01/2018 and should get results. 

**Completion criteria:**
- [x]  Consider the necessary front-end interaction changes for the following datatables
  - All receipts https://www.fec.gov/data/receipts/
  - Individual contributions https://www.fec.gov/data/receipts/individual-contributions/
  - Disbursements https://www.fec.gov/data/disbursements/
- [x] Discuss in design sync and make updates, if necessary
- [x] ~Update [front-end implementation issue, 2812](https://github.com/fecgov/fec-cms/issues/2812) with mockup and completion criteria based on mockup~ Making this issue the front end ticket as well, since a lot of the thread about design and implementation is in this ticket.
- [x] Finish up error message handling
- [x] Notify @JonellaCulmer when we know the work will be finished so she can reach out to potential power users for usability testing

patphongs:
## Summary

- Resolves #2811 
We're introducing a lot of new functionality for schedule_a datatable search, basically opening it up across all time periods as long as secondary criteria are defined.

- [x] Updated heading labels for fields to include minor heading styling
- [x] Uses multi-year select for report year box
- [x] Adds report year to the filter panel
- [x] Added custom error handling specifically for individual contributions and receipts datatables. 
        - [ ] Error message when no secondary criteria are defined: "You're searching a large dataset. Filter by recipient name or ID, contributor details, or image number."
- [x] Restricted fields are: Time periods, More contributor details, Recipient committee type, and Receipt details. These fields will lock or become disabled if multiple time periods are selected without any recipient name or ID, contributor details, or image number defined.
- [x] Path redirects with two_year_transaction_period=2020&min_date=01%2F01%2F2019&max_date=12%2F31%2F2020 in the path for both schedule_a datatables

- Receipts datatable
- Individual contributions datatable

## Screenshots

Screen Shot 2019-06-13 at 10 14 51 AM

## How to test
- Check out the branch
- `npm run build`
- Go to receipts datatable: http://localhost:8000/data/receipts/ and Individual contributions datatable, http://localhost:8000/data/receipts/individual-contributions/
- Test items listed above in the Summary section of the PR. Also test for bugs and design issues and note in this PR.
____


fec/fec/static/js/modules/tables.js
```diff
@@ -692,14 +688,26 @@ DataTable.prototype.fetchSuccess = function(resp) {
...
+  if (textStatus == 'abort') {
+    // Pending message occurs when the previous query was cancelled due to
+    // the user adding or removing filters
+    errorMessage =
+      '<div class="filter__message filter__message--delayed"><strong>Just a moment while we process your new request. You are searching a large dataset.</strong></div>';
+  } else if (jqXHR.status == 400) {
+    // Narrow search results message occurs when multiple time periods are requested
+    errorMessage =
+      '<div class="message filter__message message--error">' +
+      '<p>To expand your search to multiple time periods, filter by recipient name or ID, contributor details, or image number.</p>' +
```
patphongs: Update message to say: "You’re searching a large dataset. Filter by recipient name or ID, contributor details, or image number."

patphongs: Done

fec/data/templates/partials/individual-contributions-filter.jinja
```diff
@@ -14,16 +15,23 @@ Filter individual contributions
 {% block filters %}
 <div class="filters__inner">
   {{ typeahead.field('committee_id', 'Recipient name or ID') }}
-  {{ text.field('contributor_name', 'Contributor name') }}
-  {{ date.partition_field('date', 'Receipt date', dates ) }}
+  <legend class="label  u-margin--top">Contributor details</legend>
+  {{ text.field(name = 'contributor_name', title = 'Contributor name', minor_label_text = True) }}
+  {{ text.field(name = 'contributor_zip', title = 'ZIP code', attrs = {'maxlength':5}, minor_label_text = True) }}
+  {{ text.field(name = 'contributor_city', title = 'City', minor_label_text = True) }}
```
JonellaCulmer: The spacing here between the filter and the filter label should match the `Contributor name or ID` filter label spacing.

fec/data/templates/partials/individual-contributions-filter.jinja
```diff
@@ -14,16 +15,23 @@ Filter individual contributions
 {% block filters %}
 <div class="filters__inner">
   {{ typeahead.field('committee_id', 'Recipient name or ID') }}
-  {{ text.field('contributor_name', 'Contributor name') }}
-  {{ date.partition_field('date', 'Receipt date', dates ) }}
+  <legend class="label  u-margin--top">Contributor details</legend>
+  {{ text.field(name = 'contributor_name', title = 'Contributor name', minor_label_text = True) }}
+  {{ text.field(name = 'contributor_zip', title = 'ZIP code', attrs = {'maxlength':5}, minor_label_text = True) }}
```
JonellaCulmer: The spacing here between the filter and the filter label should match the `Contributor name or ID` filter label spacing.

fec/data/templates/partials/individual-contributions-filter.jinja
```diff
@@ -14,16 +15,23 @@ Filter individual contributions
...
+  {{ text.field(name = 'contributor_city', title = 'City', minor_label_text = True) }}
+  {{ text.field(name = 'contributor_occupation', title = 'Occupation', minor_label_text = True, helper_text = 'Use the employer field to search by occupation in reports filed before 2003') }}
+  {{ text.field(name = 'contributor_employer', title = 'Employer', minor_label_text = True) }}
```
JonellaCulmer: The spacing here between the filter and the filter label should match the `Contributor name or ID` filter label spacing.

JonellaCulmer:
@patphongs 
### For Receipts:
- For when a user adds an additional time period without secondary criteria and the other filters lock, would it be possible to still close the time period dropdown filter? (See screenshot)
- Also, can we not gray out the error message? It all happens so quickly, but I still would like the error message to be easy to read. (See screenshot)
- The text "When searching multiple time periods..." should be size 12 and unitalicized. It should resemble the text at the top of the filter panel for the Processed/Raw toggle.
- In Contributor details section change "Contributor name or ID" to "Name or ID"
- In Time period section, change "Receipt date" to "Receipt date range"

Screen Shot 2019-06-13 at 3 12 15 PM

### For Individual Contributions:
- Same style changes as above.
- Change "Contributor name" to "Name"
- "When searching multiple time periods, choose one or more fields: recipient name or ID, contributor name ~or ID~, city, ZIP code, occupation or employer, or image number.
- In Time period section "Receipt date" should be "Contribution date range"
- Change panel section "Transaction details" to "Contribution details"

patphongs:
@JonellaCulmer I believe I have resolved all of the comments you had above

JonellaCulmer:
@patphongs Everything looks great!

The only thing I'm seeing now is that if the other sections (not required filters) are open and the checkboxes are active and usable and users will get another error. Not saying that's a blocker to move forward. Just noting as a potential issue, but the users should let us know.

fec/data/templates/macros/filters/text.jinja
```diff
@@ -1,6 +1,13 @@
-{% macro field(name, title, attrs={}, id_suffix='') %}
+{% macro field(name, title, attrs={}, id_suffix='', minor_label_text=False, helper_text=False) %}
 <div class="filter js-filter" id="{{ name }}-field{{id_suffix}}" data-filter="text">
-  <label class="label" for="{{ name }}{{id_suffix}}">{{ title }}</label>
+  {% if minor_label_text %}
+    <label for="{{ name }}{{id_suffix}}" class="u-margin--bottom--small">{{ title }}</label>
+  {% else %}
+    <label class="label" for="{{ name }}{{id_suffix}}">{{ title }}</label>
```
rfultz: This is in the original in several places, where we use spaces around variables inside the double curly brackets in some instances but not others, even on the same line. Just a linting thing

fec/data/templates/macros/filters/typeahead-filter.jinja
```diff
@@ -1,11 +1,15 @@
...
+{% macro field(name, title, instructions=False, dataset='committees', allow_text=False, id_suffix='', minor_label_text=False) %}
 <div
   class="filter js-filter typeahead-filter"
   data-filter="typeahead"
   data-name="{{ name }}"
   id="{{ name }}-field{{id_suffix}}"
   data-dataset="{{ dataset }}" {% if allow_text %}data-allow-text{% endif %}>
-  <label class="label" for="{{ name }}{{id_suffix}}">{{ title }}</label>
+  {% if minor_label_text %}
+    <label for="{{ name }}{{id_suffix}}">{{ title }}</label>
```
rfultz: Since <label> are supposed to describe their parent elements, should we wrap this inside something like a <fieldset>? It may change the layout, though, so maybe it's a follow-up ticket? Looking at the page, it looks like we're using visual decoration to separate groups rather than using semantic separations. Could it be worth a follow-up ticket to use semantic grouping rather than only superficial separations?

fec/data/views_datatables.py
```diff
@@ -112,7 +112,15 @@ def independent_expenditures(request):
     })
 
 
+# Temporarily adding min and max date redirect until we can
+# handle null disbursement dates in a future implementation
```
rfultz: Should we add a TODO note? Or just create a follow-up reminder ticket?

fec/fec/static/js/modules/tables.js
```diff
@@ -278,6 +278,12 @@ function filterSuccessUpdates(changeCount) {
 
     $('.is-successful').removeClass('is-successful');
     $('.is-unsuccessful').removeClass('is-unsuccessful');
+    // Enable all restricted fields on success
+    $(
+      '.restricted-fields input, .restricted-fields button, .restricted-fields legend'
+    )
+      .css('opacity', '1')
+      .css({ pointerEvents: 'auto' });
```
rfultz: I don't know if there's a performance difference either way between this way, `.css({ pointerEvents: 'auto', 'opacity', 1 });`, adding/removing a class name for that functionality, using css to define opacity and pointerEvents based on some other state of fields.

fec/fec/static/js/modules/tables.js
```diff
@@ -338,12 +344,10 @@ function filterSuccessUpdates(changeCount) {
 
     message = '<strong>' + filterAction + '</strong><br>' + filterResult;
 
-    if ($filterMessage.length) {
-      $filterMessage.fadeOut().remove();
-      // if there is a message already, cancel existing message timeout
-      // to avoid timing weirdness
-      clearTimeout(messageTimer);
-    }
+    $filterMessage.fadeOut().remove();
```
rfultz: Will we get an error if $filterMessage doesn't exist for some reason?

fec/fec/static/js/modules/tables.js
```diff
@@ -439,7 +441,11 @@ var defaultOpts = {
   },
   pagingType: 'simple',
   title: null,
-  dom: browseDOM
+  dom: browseDOM,
+  error400Message:
+    '<strong>We had trouble processing your request,</strong><br>' +
```
rfultz: Any reason we shouldn't remove the forced line break?

fec/fec/static/js/pages/datatable-individual-contributions.js
```diff
@@ -20,6 +20,8 @@ $(document).ready(function() {
     useFilters: true,
     useExport: true,
     rowCallback: tables.modalRenderRow,
+    error400Message:
+      "<p>You're trying to search across multiple time periods. Filter by recipient name or ID, contributor details, or image number for results.</p>",
```
rfultz: Any concerns with changing the double quotes to single?

fec/fec/static/js/pages/datatable-receipts.js
```diff
@@ -22,12 +22,14 @@ $(document).ready(function() {
     useFilters: true,
     useExport: true,
     rowCallback: tables.modalRenderRow,
+    error400Message:
+      "<p>You're trying to search across multiple time periods. Filter by recipient name or ID, contributor details, or image number for results.</p>",
```
rfultz: Could we use single quotes to be more consistent with linter rules?

fec/data/templates/macros/filters/typeahead-filter.jinja
patphongs: Thanks @rfultz for bringing this up. Since these jinja templates are used for multiple different datatable filters, I'd rather make these changes separate from this PR. Also thanks for agreeing to make a separate tech debt ticket for this!

fec/data/views_datatables.py
patphongs: @rfultz So this is related to the sort_nulls_last performance improvements here: https://github.com/fecgov/openFEC/issues/3775

fec/fec/static/js/modules/tables.js
patphongs: Thanks for the suggestion! I've added  `is-active-filter` and `is-disabled-filter` classes to the form styles and updated the JS accordingly.

fec/fec/static/js/modules/tables.js
patphongs: The $filterMessage is set already above as a selector. So we don't really need to check the number of elements, we want to remove them regardless of if there's 0 or many.

fec/fec/static/js/modules/tables.js
patphongs: Looks like the intention of this error message was to have the bold be somewhat of a title and then the text below being kind of sub text. See text below, I updated it to match what the original message had in the screenshot.

Screen Shot 2019-06-14 at 8 16 59 PM

fec/fec/static/js/pages/datatable-individual-contributions.js
patphongs: Thanks, updated this to use HTML char entities within the quotes and applied single quotes.

fec/fec/static/js/pages/datatable-receipts.js
patphongs: Updated this to use HTML char entities within the quotes and applied single quotes also

fec/data/templates/macros/filters/text.jinja
patphongs: Fixed it for this jinja file. I know there's other instances, but I don't want to put too many of these code linting changes to clutter up this PR. We can look into fixing the other ones later.

patphongs:
Thanks @rfultz for your thoughtful review. I believe I have addressed all of your comments. In addition, we talked about how the datatable was working in internet explorer. The local set up needs to connect to stage API for better performance.

johnnyporkchops:
I noticed on thing first load of Receipts datatable.
if I choose another time period (other than the aleady-chosen 2019-2020):
- I get the error message, but the new time period filter is shown as applied (checked).
- The pagination is now disabled so I could not paginate through 2019-2020. If I click the pagination, the error message goes away, but the filter still shows as applied, but with red-x.
Screen Shot 2019-06-16 at 7 08 12 PM

Maybe this is not possible, but:
Should we make it so either that dropdown only allows one choice, unless other criteria are applied
--OR--
Just unapply the newly chosen time period so the datatable is still operable with just the initial (2019-2020) time period filter applied. And the user does not have to remove it.

johnnyporkchops:
On both Contributions and Receipts, should we change `Recipient name or ID` to `Recipient name or committee ID`. An uninitiated user, I might click through from the home page barcharts to a candidate's page and grab their ID from the candidate profile page and try to use that on these datatables, which does not work.

patphongs:
> I noticed on thing first load of Receipts datatable.
> if I choose another time period (other than the aleady-chosen 2019-2020):
> 
> * I get the error message, but the new time period filter is shown as applied (checked).
> * The pagination is now disabled so I could not paginate through 2019-2020. If I click the pagination, the error message goes away, but the filter still shows as applied, but with red-x.

Thanks @johnnyporkchops for your comments. In regards to the pagination not working, later down the road, we'll want to add some kind of indication that the table has not updated and perhaps disable interaction until the error goes away. The red-x and the error may not be enough after we do some user testing. In order to make this work, we'll make an additional ticket to iron out the tableSwitcher logic. We don't want to initialize the datatable within the same logic table switch logic. The tableSwitcher is what is causing the additional bad request (400 call) to happen.

JonellaCulmer:
> On both Contributions and Receipts, should we change `Recipient name or ID` to `Recipient name or committee ID`. An uninitiated user, I might click through from the home page barcharts to a candidate's page and grab their ID from the candidate profile page and try to use that on these datatables, which does not work.

@johnnyporkchops This is a good point. And something I'll pay attention to during testing.

PR #2957, opened 2019-06-10
Address polyfills for Internet Explorer (IE)

### LINKED ISSUE #2870: Calendar not showing properly in Internet Explorer
**What we're after** A working calendar. Some users, including FEC staff, must use IE as their default browser, so we need the calendar to work in IE.

**Background** 
From an OGC staff member; I also experienced the same result when I tested it so that's my screenshot.

> Good morning,
I’ve been meaning to let you know about an issue I’ve encountered lately when accessing the FEC site using IE: the Calendar link (upper right, next to the glossary and search box) doesn’t work.  When I click I get this:
 
image

> In Firefox I have no problem getting the calendar page and all the filters.

From Dorothy: I noticed that the updates feed on the home page does not show the next event for each category in IE. This is also drawn from the calendar and possibly related.

image

**To do:**
- [ ] Find the problem with IE showing the calendar and populating the updates feed on home page..
- [ ] Fix the problem so IE shows the calendar and populates the updates feed on home page.

### LINKED ISSUE #2958: Fix issues with Internet Explorer
Some tweaks to make our site/app work a little better for those still using outdated browsers (mostly Internet Explorer 10 & 11). This is a test ticket because we're changing the X-UA meta tag on every page of the site and we're changing the build process for every datatable page.

rfultz:
## Summary
Some tweaks to make our site/app work a little better for those still using outdated browsers (mostly Internet Explorer 10 & 11). This is a test ticket because we're changing the X-UA meta tag on every page of the site and we're changing the build process for every datatable page.

- Resolves #2958 
- Resolves #2870 
- Resolves https://github.com/fecgov/FEC/issues/5751

- **Updated our `X-UA-Compatible` meta tag to target `edge` instead of `IE11`** ([Microsoft](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/dev-guides/bg182625(v=vs.85)): "edge mode is the preferred document mode; it represents the highest support for modern standards available to the browser." 
- **Added `require('babel-polyfill');`** to aggregate-totals*.js and datatables-*.js files.
- `package.json` has been updated to include our targeted browsers & versions. The values are pretty standard, though we added IE 10, too.
- Compensated for IE not supporting `document.currentScript` (aggregate-totals-box).

## Screenshots
None

## Related PRs
None

## How to test
- Pull the branch.
- `npm run build`.
- `npm run build-production`.
- `npm run test-single`.
- Any errors?
- Do the JavaScript-powered parts of the site still work in non-IE?
- Do the JavaScript-powered parts of the site work in IE 11? IE 10?
- If there are no errors, it's likely things will work normally with modern browsers. Then it's just up to IE 10 & 11.
High priority JavaScript-powered components to check: datatables, aggregate totals (behind a feature flag).
- Want to make sure we didn't re-break #2424 

## Possible next steps
- Roll the change into develop and/or release.
- Update any other pages that reference polyfills to use the same `require('babel-polyfill');` method.
- Check the backlog for IE-specific bugs and whether this addresses those.
____


rfultz:
Aggregate Totals works in IE 11!
image001

rfultz:
@dorothyyeager, I wasn't able to re-create the issue with the menu starting in an open state and then closing so I don't know what's going on there. Seems like it would be a jQuery issue but that that should affect every IE user every time.

Inside the aggregate totals box, the vertical divider isn't going to be there for IE, which doesn't like flexbox. The spacing is still there but the divider isn't. IE also doesn't like the meter elements on the right side so I've switched those off.

The data issues are indeed differences in environments (dev, stg, prd).

Same thing as the Submit a Question accordion item.

The Pages column head was because of a discrepancy between develop and this branch; that's been addressed.

dorothyyeager:
Just tested again @rfultz - Am still getting the About menu dropping down and blinking off every time I click on something within the data or the legal resources part. It's noticeable enough that we might need to figure it out. I can demo for you via Slack call if you want to see what's happening with that.

I'm good with the other things you mentioned. The pages thing and the calendar issue are resolved, so that's two issues fixed! Thanks.

PR #2977, opened 2019-06-18
Adjust z-index of .data-table

### LINKED ISSUE #2952: Download options are not clickable on candidate filings page
### Summary

A candidate attempted to view the PDF of their Form 2 but the drop down menu (for CSV, .fec, PDF download) is behind another HTML element and the user cannot click to view the PDF.

It looks like the Div w/ class #results-info overlays the download drop down options.

### Expected Behavior

User should be able to click all available options when they expand the drop down for CSV/.fec/PDF.

### Actual Behavior

The options are not clickable.

### Frequency

- [ ] This is the first time it's happened
- [x] This is an ongoing, predictable issue
- [ ] It seems to be random
- [ ] Other: …

### How to Reproduce

URL: https://www.fec.gov/data/candidate/H0FL13125/?tab=filings

1. Expand the download options (CSV / .fec / PDF should become visible)
1. Attempt to click PDF

### Browser

_Which version of which browser are you using? Does this issue occur with other browsers or with other browser sizes?_

- [x] Chrome
- [x] Firefox
- [ ] Safari, (didn't test)
- [ ] Edge (didn't test)
- [x] Internet Explorer
- [ ] Other: …

### Device

_Tell us what you can about your device and its operating system._

- [x] Laptop/desktop
- [ ] Tablet
- [ ] Phone

### Screenshots

image

### Misc

If you resize the browser window (make it less wide relative to the height) you can expose the PDF option and download the PDF.

rfultz:
## Summary

Assigning a (new) z-index to parent objects of download options lists.

Pull-downs inside tables were visible but couldn't receive click/tap actions when the row was such that its pull-down could appear behind the table's footer—the footer's transparent background was receiving the click/tap action.

Because elements load and stack in front of preceding elements by default, we're giving a z-index to tables so the pull-down elements inside them will appear in front of the tables' footer elements.

- Resolves #2952 

The only change was to apply a z-index to .data-table at line 68, which will only affect those tables. As a side effect, though, there could be a situation where something that was (and should have been) in front of the table will now be behind the a data-table element.

Adjusted the `top` value for `.dropdown__panel` from 3rem (which doesn't adjust for rows that are more than one line tall) to a calculation of [middle - 1.5rem (half of 3, the height of the `<button>` that toggles the `<ul>`) - 1 px].

The other changes in the file are from the linter and are "nothing"—zeroes and whitespace adjustments.

## Screenshots

The overall look should be unaffected, but the .fec and .pdf options should now receive hover/rollover and click/tap events and react accordingly:
image

## Related PRs
None

## How to test
- Pull the branch
- `npm run build`
- Check tables with download options that collide with the table footers to make sure other options than the first receive mouse events.
  - At http://localhost:8000/data/committee/C00000885/?tab=filings, the last entries' download pull-downs' .fec and .pdf options should be fully functional. The other pulldowns on the page should still be unaffected and fully functional.
  - Same with http://localhost:8000/data/candidate/P00011486/?tab=filings, the last row in the table, its download pull-down should be fully functional while the table's footer elements—and every other element on the page—should function as expected.
- Other places across the site where the data-table class is used should be unaffected—basically every data table.
____

### DIFF
--- a/fec/fec/static/scss/components/_datatables.scss
+++ b/fec/fec/static/scss/components/_datatables.scss
@@ -255,7 +255,7 @@
 
   .dropdown__panel {
     right: 0;
-    top: u(3rem);
+    top: calc(50% + 1.5rem - 1px);
     min-width: 70px;
   }
 
--- a/fec/fec/static/scss/components/_datatables.scss
+++ b/fec/fec/static/scss/components/_datatables.scss
@@ -60,16 +60,17 @@
 .data-table {
   @include transform(translateX(0%));
   @include transition-property(transform);
-  @include transition-duration(.5s);
+  @include transition-duration(0.5s);
   height: 100%;
   font-family: $sans-serif;
   position: relative;
   table-layout: auto;
+  z-index: $z1;
 
   th {
     line-height: u(1.4rem);
-    letter-spacing: -.3px;
-    padding: u(.5rem 1rem);
+    letter-spacing: -0.3px;
+    padding: u(0.5rem 1rem);
     border-bottom: 1px solid $primary;
 
     &.sorting,
@@ -116,7 +117,7 @@
 
     &.row-color-contrast,
     &:nth-child(even) {
-      background-color: rgba($gray-lightest, .5);
+      background-color: rgba($gray-lightest, 0.5);
     }
 
     &.row-color-normal {
@@ -124,12 +125,12 @@
     }
 
     &.row-active {
-      border-left: .5rem solid $primary;
+      border-left: 0.5rem solid $primary;
     }
   }
 
   td {
-    @include transition(padding-left, .2s);
+    @include transition(padding-left, 0.2s);
     padding: u(1rem);
     border-left: 1px solid $neutral;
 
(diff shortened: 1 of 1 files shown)


rfultz:
@JonellaCulmer, I added you as optional if you'd like to check random data-table pages. Or feel free to remove yourself as a reviewer.

rfultz:
@JonellaCulmer, the issue where the bottom of the "pull-down" (it's not actually a `<select>` but a `<ul>`) is so close to the bottom of the table that their bottom borders kinda merge? I saw that! …have no idea why Chrome's doing that—Firefox and Safari draw the two borders as expected.

The scrollbar is happening because the `<ul>` technically makes the table taller. I'll see if I can play with the overflow setting, if there's a quick fix.

Not sure there's anything I can do about the alignment between the bottom of the button and the top of the `<ul>`. It wouldn't be an issue if we were using `<select>`. I suspect what's happening is the top (y) is being set a certain distance from the top of the row, which would explain why it's off for the two-line rows but fine for one-line rows. I'll try to set its y to a calculation without it getting too hacky or losing support for some browsers (IE + `calc` worry me).

rfultz:
@JonellaCulmer I just pushed an update to adjust the top position of those download lists

PR #2990, opened 2019-06-21
[WIP] Update removeAllTags to have default values

### LINKED ISSUE #2987: Fix Clear all filters link on receipts, indiv. contributions datatables
**What we're after:** On the Receipts and Individual contributions datatables, when clicking the `Clear all filters` link in the top-right corner of the table, it just looks like the filters are removed, and the table reverts to `All report years`, which it shouldn't do. This is a problem because if the user does not have any specific "required" filters applied when searching across all years then the filter panel is locked. So if the user tries to filter by time period at this point, they will not be able to. There is also no error message telling the user there's something they need to do. 

When the user clears all filters it should return to the default search state of the 2019-2020 time period, and the date range of 01/01/2019-12/31/2019. 

The same problem exists on the disbursement datatables, but it's not as urgent since we don't require any fields on this table.

**Completion criteria:**
- [ ] Change function of `Clear all filters` link to return to default state instead of `All report years`

Jun-21-2019 11-06-05

rfultz:
## Summary

- Resolves #2987 

Added some default values to removeAllTags inside filter-tags.js and triggered fake events on those elements to more reliably integrate with the current structure.

## Screenshots
None

## Related PRs
None

## How to test
- Pull the branch
- `npm run build`
- `npm run test-single`
- Make sure no tests failed
- Check the receipts, individual contributions, and other datatables'
  - 'clear all filters' link text (should read "Reset filters" for Receipts and Individual Contributions but remain "Clear all filters" for every other use)
  - default states (data present, no errors, no stalls)
  - state as data is filtered down (use some non-date filters) (data and filters update appropriately)
  - state when the date filters are removed from the top of the tables
    - after the filters are cleared [1]
  - state when the date filters are deselected/removed from the left/filters
    - after the filters are cleared [1]
  - state when the Report Time Period tag/label is totally removed from the filters
    - after the filters are cleared [1]
- *Anywhere else "Clear all filters" is used that we may have missed?*

[1]: Receipts and Individual contributions tables should revert to their default state, i.e. with 2019-2020 and 01/01/2019-12/31/2020 filters applied. Other tables should clear all filters.

# Question / Known issue
[Receipts](http://127.0.0.1:8000/data/receipts/?data_type=processed&two_year_transaction_period=2020&min_date=01%2F01%2F2019&max_date=12%2F31%2F2020) and [disbursements](http://127.0.0.1:8000/data/disbursements/?data_type=processed&two_year_transaction_period=2020&min_date=01%2F01%2F2019&max_date=12%2F31%2F2020) default with an "All report years" filter applied, but clicking Clear all filters now changes it to "2019-2020" (calculated from DEFAULT_ELECTION_YEAR)

Is this okay? Do we need a more solid option?

____


johnnyporkchops:
@laura, it may not be possible to make it to certain datatabled, because I believe it’s set globally in filters-tags.js

rfultz:
Yeah, that one piece of code is used for all the filters site-wide. But we could find something else to check, like a URL substring, template name, something... I'm open to suggestions.

lbeaufort:
@johnnyporkchops @rfultz I had a conversation with @JonellaCulmer in the #dev channel in Slack about the expected behavior for the datatables. Let's all get on the same page on Monday, sorry to cause confusion!

rfultz:
*NOTE*
While reviewing the PR with @JonellaCulmer, we discovered something we want to follow up on.
Going to [Receipts](http://127.0.0.1:8000/data/receipts/), the default for localhost is 01/01/2017-12/31/2020 And "All report years". The "All report years" is a problem in that it should be "2019-2020" instead. If there's a discrepancy and there was a change to the functionality around that, we should check why that change was made and what it might mean to undo it

JonellaCulmer:
Based on @rfultz note above I'm approving with the caveat that we should pay close attention to how it behaves once it's on stage. With stage acting strangely yesterday and with what Robert noted above i have concerns about this going out.

patphongs:
Thanks @rfultz for your initial work on this fix! I have used your work as a base for my new PR here: https://github.com/fecgov/fec-cms/pull/3062

PR #3100, opened 2019-08-07
[DO NOT MERGE-feature space only] Test/fix inadequate headers

patphongs:
This is to test the content-security-policy headers on the CMS and how it interacts with the staging API. The goal is to get no errors.

### DIFF
--- a/fec/data/templates/elections.jinja
+++ b/fec/data/templates/elections.jinja
@@ -53,6 +53,7 @@ var context = {
   election: {
     cycle: '{{ cycle }}',
     election_full: true,
+    duration: '{{ election_duration }}',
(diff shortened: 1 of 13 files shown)


lbeaufort:
Based on some research, it appears that adding the CSP as an HTTP header is more secure. In addition, the vulnerability alert notes the missing HTTP header, not the `meta` tag. 

Since @patphongs is away, I can draft up a branch that includes this CSP as an HTTP header rather than `meta` tags to see how that works.

Testing tool: https://securityheaders.com/?q=https%3A%2F%2Ffec-feature-cms.app.cloud.gov%2F&hide=on&followRedirects=on

PR #3124, opened 2019-08-22
Add security response headers

### LINKED ISSUE #3069: [Tenable] CMS missing/inadequate headers (due 8/31/19)
### 1. Missing 'Cache-Control' Header 
The HTTP 'Cache-Control' header is used to specify directives for caching mechanisms. The server did not return a 'Cache-Control' header which means page containing sensitive information (password, credit card, personal data, social security number, etc) could be stored on client side disk and then be exposed to unauthorised persons. This URL is flagged as a specific example: fec-dev-cms.app.cloud.gov 
Remediation path: Configure your web server to include a 'Cache-Control' header with appropriate directives. If page contains sensitive information 'Cache-Control' value should be 'no-cache, no-store' and 'Pragma' header value should be 'no-cache'. 
References: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma; https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control; https://www.owasp.org/index.php/Testing_for_Browser_cache_weakness_(OTG-AUTHN-006)
### 2. Missing Content Security Policy 
Content Security Policy (CSP) is a web security standard that helps to mitigate attacks like cross-site scripting (XSS), clickjacking or mixed content issues. CSP provides mechanisms to websites to restrict content that browsers will be allowed to load. No CSP header has been detected on this host. This URL is flagged as an specific example: fec-dev-cms.app.cloud.gov 
Remediation path: Configure Content Security Policy on your website by adding 'Content-Security-Policy' HTTP header or meta tag http-equiv='Content-Security-Policy'. 
References: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy; https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP; https://csp-evaluator.withgoogle.com/; https://developers.google.com/web/fundamentals/security/csp/; https://content-security-policy.com/

**To do**
- [ ] Front end investigates use of Engine X or the CMS application itself to investigate headers. (Also, are there core settings?)

lbeaufort:
## Summary

Resolves #3069 
- Adds content security policy response headers to every page. See original issue for background information.
- Add `report-uri` that sends violation reports to API for now (ran into csrf issues with a Django route)
- Add `CSRF_TRUSTED_ORIGINS` setting to address csrf errors

-  This will impact everything on our website. Need to make sure that there are no loading errors for anything we pull into our site like api calls, scripts, images, fonts, etc. 

## Screenshots

None, this is pure code

## Related PRs

branch | PR
------ | ------
openFEC/feature/3876-add-response-headers | [link](https://github.com/fecgov/openFEC/pull/3891)

## How to test
Check each section of the site to make sure that the content security policy is not interfering with anything we're pulling into the site.
- [x] Homepage
- [x] Data landing
- [x] Election search page
- [x] All datatable pages
- [x] Candidate profile pages
- [x] Committee profile pages
- [x] Election profile pages
- [x] Wagtail admin (creating new pages, updating pages, saving pages, adding new block components)
- [x] Wagtail generated pages
- [x] Legal resource pages

## How to test the changes on stage

- I pushed a deploy to `stage` with Circle (as of 8/28) (stage.fec.gov)
- Test API for console errors
- Curl API to check headers
- Run scan like https://securityheaders.com/ and https://csp-evaluator.withgoogle.com/ results
- Make sure CMS can still access what it needs

### DIFF
--- a/fec/fec/middleware.py
+++ b/fec/fec/middleware.py
@@ -3,23 +3,26 @@ from django.conf import settings
 
 
 class SecurityHeaders(MiddlewareMixin):
+    """Add secure headers to each response"""
 
     def process_response(self, request, response):
 
         if settings.FEC_CMS_ENVIRONMENT == settings.ENVIRONMENTS.get('local'):
-            content_security_policy = ''\
-                'default-src \'self\' data: *.fec.gov localhost:* http://127.0.0.1:* http://*.app.cloud.gov https://*.app.cloud.gov; '\
-                'frame-src \'self\' https://www.google.com; '\
-                'img-src \'self\' data: http://*.fastly.net;'\
-                'script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' https://www.google.com https://www.gstatic.com https://polyfill.io; '\
-                'style-src \'self\' data: \'unsafe-inline\'; '
+            content_security_policy = \
+                "default-src 'self' " \
+                "data: *.fec.gov localhost:* http://127.0.0.1:* *.app.cloud.gov; " \
+                "frame-src 'self' https://www.google.com; " \
+                "img-src 'self' data: http://*.fastly.net; " \
+                "script-src 'self' 'unsafe-inline' 'unsafe-eval' https://www.google.com https://www.gstatic.com https://polyfill.io; " \
+                "style-src 'self' data: 'unsafe-inline'; "
         else:
-            content_security_policy = ''\
-                'default-src \'self\' data: *.fec.gov; '\
-                'frame-src \'self\' https://www.google.com;'\
-                'img-src \'self\' data: http://*.fastly.net; '\
-                'script-src \'self\' \'unsafe-inline\' \'unsafe-eval\' https://www.google.com https://www.gstatic.com https://polyfill.io https://dap.digitalgov.gov; '\
-                'style-src \'self\' data: \'unsafe-inline\'; '
-        response['Content-Security-Policy'] = content_security_policy
-        response['cache-control'] = "max-age=600"
+            content_security_policy = \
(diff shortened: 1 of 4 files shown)


fec/fec/middleware.py
```diff
@@ -0,0 +1,31 @@
...
+    def process_response(self, request, response):
+
+        # Report violations to the API due to CSRF issue with Django route
+        REPORT_URI = "{0}/report-csp-violation/?api_key={1}".format(
+            settings.FEC_API_URL, settings.FEC_API_KEY_PUBLIC
+        )
+        content_security_policy = {
+            "default-src": "'self' *.fec.gov *.app.cloud.gov https://www.google-analytics.com",
+            "frame-src": "'self' https://www.google.com/recaptcha/",
+            "img-src": "'self' data: https://*.ssl.fastly.net https://www.google-analytics.com *.app.cloud.gov",
```
patphongs: It appears we're calling to the fastly CDN for some of the map graphics via http, not https. I'm not sure how to change that to https yet, so in the meantime, can you please change that to http? That will fix the errors on the election search landing page and the homepage.

lbeaufort: @patphongs thanks for taking a look! It looks like the `leaflet-providers` package has a fix for this in a later version (see https://github.com/leaflet-extras/leaflet-providers/issues/227). I updated to the latest version and it looks like the errors went away. Can you take another look? Updated on `stage`.

patphongs: Yes that did work! Looks great, thank you @lbeaufort !

Issue #3072, opened 2019-07-22
Build wagtail template for organization chart


JonellaCulmer:
**What we're after:**
As part of the next steps of implementing the organization chart on .gov, we need to build the wagtail template and make it easy for content to edit and change names and fields. It should be fairly straightforward to implement the html that @rfultz  has created into editable Wagtail template. Given the fairly predictable structure of the table and the changes editors might make, this is a reasonable candidate for a Wagtail streamfield block. I can work on the Wagtail template.

### Related issues
- https://github.com/fecgov/fec-cms/issues/2877
- https://github.com/fecgov/fec-cms/issues/3071

### Completion criteria

_List action items that need to be done to complete this task. List in checklist formatting._
- [ ] Build new wagtail template for the organization chart for .gov
- [ ] Meet/discuss with content and design on what's built to make sure it meets the needs of those who will implement changes as necessary.

PR #3130, opened 2019-08-27
3071 Org chart styles

### LINKED ISSUE #3071: HTML/responsive-side work for organizational chart
**What we're after:** Move forward with the on-going work done by @rfultz, re-styled as part of the next steps to implement the updated org chart. The smaller widths will be a stacked list (single column), though we're still thinking about ways to represent the non-hierarchal relationship between the commissioners and IG (maybe two columns for them?). We'll probably lose the connecting lines for the smaller widths; still need to engineer how they'll work for the larger widths.

We're researching options to efficiently administer the information inside the CMS, keeping in mind the upcoming work for streamlining the admin for tables that are currently html blocks.

### Related issues
- https://github.com/fecgov/fec-cms/issues/2877
- https://github.com/fecgov/fec-cms/issues/3072

### Completion criteria
- [ ] Update existing work to match styling 
- [ ] Consider solution for connecting lines (possible MVP stacking columns left to right without lines)
- [ ] Meet/discuss work with @AmyKort and design for review

### Design
Organization chart

**Future work**
- #3072 follows this ticket

rfultz:
## Summary

- Resolves #3071 

Outlines the structure for a responsive org chart and styles it according to comps

Right now, the only change is a new scss file that will apply to an upcoming org chart.
That file is included in fec/fec/static/scss/base.scss for now (see the TODOs at the bottom).

## Screenshots

image
image

## Related PRs

None

## How to test

- Pull the branch
- `npm run build` just in case (if you get an error, you may need to run `npm rebuild node-sass` then build again)
- Copy the sample html from the bottom of fec/fec/static/scss/components/_fec-org-chart.scss
- I'm still new to Wagtail; I think this is what I did:
  - In local Wagtail, find a page to edit (I chose Home > About > Leadership and structure)
  - Add a new html section
  - Paste the copied html into the new section
  - Save draft
  - Preview
- Scale your browser window from its smallest width across all the sizes to 980px+

### TODO:

- @JonellaCulmer, is the hover too subtle—should we make it stronger?
- #3072 will work out how the org chart lives on the site; we may need to move the css then.
- We need a wider page template
- Does it still make sense to include _fec-org-chart.scss in base.scss or would it be better somewhere else?
____

### DIFF
--- a/fec/fec/static/scss/components/_fec-org-chart.scss
+++ b/fec/fec/static/scss/components/_fec-org-chart.scss
@@ -1,12 +1,16 @@
 // Org Chart Styles
 //
 // Everything related to the org chart
+// Should be included inside the css of whatever page it's included
 //
 
+$colorStandard: #b2b1b6;
+$colorGreyHover: #d6d7d9;
+
 $gutterSize: 10px;
-$lineDashed: thin dotted grey;
-$lineSolid: 2px solid grey;
-$lineSolidDark: thin solid black;
+$lineDashed: thin dashed $colorStandard;
+$lineSolid: 2px solid $colorStandard; // line for the wider layout
+$lineSolidDark: thin solid $colorStandard;
 
 #fec-orgchart {
   display: block;
@@ -56,29 +60,39 @@ $lineSolidDark: thin solid black;
       }
     }
     
-    div {
+    a {
+      align-items: center;
       border: $lineSolidDark;
+      border-radius: 4px;
+      display: flex;
+      flex-direction: column;
+      flex-wrap: wrap;
+      justify-content: center;
       list-style-type: none;
       margin: 0;
       min-height: 80px;
-      padding: 2rem;
+      padding: 1rem;
       position: relative;
       text-align: center;
+      transition: border-color .1s;
 
-      display: flex;
-      align-items: center;
-      justify-content: center;
-      flex-wrap: wrap;
-      flex-direction: column;
+      &:hover {
+        border-color: $colorGreyHover;
+        transition: border-color .25s;
+      }
+      &:not([href*="/"]) {
+        // if the link doesn't contain a slash (e.g. if it's blank), let's remove its clicks/taps
+        pointer-events: none;
+      }
 
       span {
         display: block;
-        line-height: 1em;
+        line-height: 1.2em;
         
         &:first-child {
           font-size: 14px;
           font-weight: bold;
-          line-height: 1em;
+          margin-bottom: .1em;
         }
         &:nth-child(2) {
           font-family: Karla, sans-serif;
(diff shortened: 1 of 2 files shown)


JonellaCulmer:
@rfultz In your instructions, we also need to run `npm run build-sass`, or at least I had to.

rfultz:
> @rfultz In your instructions, we also need to run `npm run build-sass`, or at least I had to.

Oops. Sorry. I've updated it.

JonellaCulmer:
@dorothyyeager 

> are the footnotes incorporated into this?

I assumed we'd retain the current process we use for footnotes. There would just need to be another text block after the org chart, which is something @johnnyporkchops will be able to contend with on the org chart template. When we create a template for footnotes, I'd like them to be the same across the whole site.

fec/fec/static/scss/components/_fec-org-chart.scss
```diff
@@ -0,0 +1,589 @@
...
+            width: $gutterSize;
+          }
+        }
+      }
+    }
+  }
+}
+
+
+@media (min-width: 768px) {
```
johnnyporkchops: @robert, Can we use existing breakpoint here ( `@include media($med) ` , which is 640px/40em ) .It seemed to work for me locally, but maybe I am missing an in-between spot where one or more elements just don't respond correctly

johnnyporkchops:
@rfultz . To answer your question, I think it makes sense to `import_fec-org-chart.scss` in `base.scss`
Also, for the `full-width page template` --That template will be part of the reporting-dates tables PR which is getting closer and closer to being complete ... if it is not, before they are ready to make the org chart html block, I can do a small separate PR to get that available in Wagtail ahead of time.

fec/fec/static/scss/components/_fec-org-chart.scss
rfultz: Yeah, we should use the standard ones. I figured I'd come back to it after we got the full-width template. The problem is that media queries are based on the window width (instead of element widths) and, at the breakpoint where the left nav comes in, one pixel smaller and the module is wider than if the page were one pixel larger. (e.g. when the window is 767px wide, the module is 767px wide, but when the window is 768px, the left nav comes on so the module only has 600px to work with.)

rfultz: @johnnyporkchops New code on the way

johnnyporkchops: However, since this is a standalone scss component, we do have precedent for using customized styles that depart from the established, especially when its a necessity to get something to work the way e want. For [example](https://github.com/fecgov/fec-cms/blob/c58d40870a9fdb3139ea012e46c7ba8ad20cb605/fec/fec/static/scss/components/_mega-menu.scss#L225), _mega-menu.scss used a `max-width` media query. Also the upcoming reporting-dates tables use custom media queries (although that may change)

patphongs:
All reviewers have approved, merging...

Issue #3125, opened 2019-08-23
Vulnerability: ESLint


rfultz:
Got a new **critical** vulnerability report from [npm audit](https://www.npmjs.com/advisories/1118)

PR #3134, opened 2019-08-28
#2840 Implement Tag Manager

### LINKED ISSUE #2840: Implement Tag Manager
Switch the site(s) to Tag Manager from the current direct analytics implementation

Acceptance Criteria
- [ ] Decide (as a team) which account ID to use
- [ ] Duplicate current analytics inside Tag Manager (page views, downloads, etc)
- [ ] Current analytics and tag manager can be chosen with a feature flag
- [ ] Analytics performance on dev, stage mirror that on prod
- [ ] No breaking errors from ad-blockers, across browsers
- [ ] Prod analytics with Tag Manager is at least as robust as ga directly

rfultz:
## Summary

- Resolves #2840 

Progress:

- [x] Add feature flag
- [x] Wrap current analytics inside the feature flag
- [x] Create Tag Manager (GTM) container for non-Production
- [x] Add Production GTM container to pages behind feature flag
- [x] Add non-Production container to pages as an alternate to the Production container
- [x] Add `<noscript>` GTM elements to templates
- [x] Add code to non-Production container to confirm it's loading
- [ ] **Confirm proper GA/DAP/GTM elements are loading**
- [ ] Comb site to migrate any necessary GA events to GTM data layer
- [ ] Confirm successful migration of GA events to GTM
- [ ] Confirm GSA DAP is unaffected (all GSA DAP events are still coming through)
- [ ] Remove current/then-retired GA and DAP implementation
- [ ] Remove feature flag?
- [ ] Add new events?

Every page of the site should be affected.

I was getting some linting messages so I let it add some line breaks. Created the env var.

The major changes were to the templates:
- fec/data/templates/layouts/[main, widgets].jinja
- fec/fec/templates/[base, home_base].html

## Screenshots

None

## Related PRs

None

## How to test
- Pull and build the site like usual
- Check the homepage for the GSA DAP tag
  1. Update DEBUG settings to false
  1. Run server
  1. check the page source (in the browser)
  1. cmd-f for `_fed_an_ua_tag` (the GSA DAP element)
  1. confirm that the tag is in the page
  1. cmd-f for 'googletagmanager`
  1. confirm that it does **not** exist
- Visit a data page and confirm that the tag is in the page source
- Go to the calendar to confirm that the GSA DAP element (`_fed_an_ua_tag`) is in the page.
- Interrupt the server
- Set the feature flag `export set FEC_FEATURE_USE_TAG_MANAGER=true`
- Restart the server
- Check the homepage source again. Confirm that:
  1. `_fed_an_ua_tag` is gone
  1. `googletagmanager` **does** appear
  1. GTM is using the `GTM-PMSLNL3` container (search the page source, or check near the end of the GTM `<script>`)
  1. GTM is **not** using the production container of `GTM-T5HPRLH`
  1. the Console in inspector is showing a message like "loaded the Tag Manager dev container"
- Check those same things for a data page
- Check those same things for a calendar page
- Celebrate 🎉

## Next steps

See the Progress list in the Summary at the top
____


fec/fec/templates/base.html
```diff
@@ -104,37 +173,50 @@ <h2 style="color: #ffffff">Your browser is outdated</h2>
     <footer class="footer">
       <div class="container">
         <div class="seal">
-          <img class="seal__img" width="140" height="140" src="{% static "img/seal--inverse.svg" %}" alt="Seal of the Federal Election Commission | United States of America">
+          <img class="seal__img" width="140" height="140" src="{% static 'img/seal--inverse.svg' %}" alt="Seal of the Federal Election Commission | United States of America">
```
patphongs: Looks like the change here is effecting the load of the white seal at the bottom footer. It's giving this error `GET http://localhost:8000/%7B%%20static 404 (Not Found)`. It may be having issue with the single quote rather than the double quote?

rfultz:
I've just pushed an update to include fixes from #3125; does this help? (`%7B%%20static` is `{% static`; an over-eager linter broke some `{% … %}` by adding line breaks, which should be fixed now)

fec/data/constants.py
```diff
@@ -5,8 +5,8 @@
 
 START_YEAR = 1979
 END_YEAR = 2020
-DEFAULT_TIME_PERIOD = 2020  # Change after the April quarterly report (4/15/21)
-DEFAULT_ELECTION_YEAR = 2020  # Change after election day (11/3/20)
+DEFAULT_TIME_PERIOD = 2020  # Change after the April quarterly report (2020-04-15)
```
lbeaufort: This should be `2021`

patphongs:
> @rfultz Not sure if I'm missing something, but after I apply the feature flag to my environment vars, none of the images, css, or scripts are loading. Looks like all 404s?

So the comment that I had about 404s, we found out it's because we're running locally as debug False as if we are in production, so when JS and CSS dependencies won't be mapped correctly on local. This is fine because the settings are compiled differently for our app environments vs local dev.

PR #3256, opened 2019-10-09
add condition for amendment_chain != null

### LINKED ISSUE #3060: Null values for reports tables
### Summary

_When there is a null value for the `amendment_chain` JSON key, we need to do a catch on the front end for what should be returned. Otherwise there is a `.length` error and the table does not return at all._

Related issue: https://github.com/fecgov/fec-cms/issues/3057

### Screenshots

Screen Shot 2019-07-18 at 12 34 34 PM

## Completion Criteria ##
- [ ] Handle Null error

jason-upchurch:
## Summary (required)

- Resolves #3060 
- add condition to `helpers.js` to ensure `amendment_chain.length` is not called before checking to see that `amendment_chain` is not null.

- `helpers.js`

## How to test
- check out this branch
- `pytest`
- run local cms and check [example case](http://localhost:8000/data/reports/house-senate/?data_type=processed&committee_id=C00498683) where `amendment_chain` is null behaves as expected

### DIFF
--- a/fec/fec/static/js/modules/helpers.js
+++ b/fec/fec/static/js/modules/helpers.js
@@ -444,7 +444,7 @@ function amendmentVersionDescription(row) {
   // This checks for terminatino reports and if the amendment number is greater
   // than 0 or the amendment chain is longer than 1, it is an amendment
   else if (row.amendment_indicator === API.amendment_indicator_terminated) {
-    if (row.amendment_chain.length > 1) {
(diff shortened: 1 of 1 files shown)


jason-upchurch:
> Changes look good, thanks @jason-upchurch! I can't seem to replicate the old issue, but the code changes look good and the results look the same.
> 
> I don't think we want to delete the `package-lock.json` file though, can you remove that change please?

👍 thanks @lbeaufort!

PR #3265, opened 2019-10-10
Feature/3246 wccf in ie

### LINKED ISSUE #3246: WCCF internet explorer compatibility issues
### Summary

_Afte testing in internet explorer, it can be seen that the map colors and the table cell number alignment is not working. In addition, when you hover over the states, the pop up box appears but never disappears when you roll off the state map._

### Screenshots

Image from iOS (5)

rfultz:
## Summary

- Resolves #3246 
- Many fixes/patches for Internet Explorer 11

- Fixed a couple typos in the jinja files
- Corrected the `this.` scopes to `instance.`
- Changed `mouseover` to `mouseenter`
- Changed `mouseout` to `mouseleave`
- Now assigning `mouseleave` to DOM elements instead of letting d3 do it (it was assigning it as `mouseout` for IE, where I most needed it to be `mouseleave`)
- Added a `mouseleave` to the svg itself, too—the country in addition to the individual states
- Added a `table-display` class that's only assigned for IE to revert tables' components' display properties back to their defaults instead of a flex layout (because IE doesn't seem to like flexy tables)
- Added `min-height` and `min-width` values to the svg because IE doesn't like to guess/calculate svg dimensions so wasn't scaling the map at all
- Adjusted z-index values so the the tooltip is above the map but below the typeahead, all of which are under the overlay

## Screenshots
Internet Explorer 11:
image

## Related PRs
Several other WCCF PRs floating around but they're independent of this one

## How to test
- Check Feature with Internet Explorer 11 (it's there as of this writing)
- Make sure WCCF looks and works okay ( /data/raising-bythenumbers )
- Check it in IE 10? It's not a target browser but it might be good to review just in case (Looks and works as it should in Chrome, Firefox Developer Edition, Safari Technology Preview, Vivaldi, Android Chrome, and the Android browser (though the site isn't great when it dark mode).

### Known differences / issues
- Rolling out of the states, the tooltip won't disappear until rolling out of the bounding box of the group of states (i.e., higher than WA, left of AK, lower than TX, right of ME)
- The table of states may not be perfectly neat but should still be visible—it may be less than 100%
- IE doesn't allow skinning native scroll components so the scrollbar for the list of states will be native to the browser
____

### DIFF
--- a/fec/fec/static/js/modules/data-map.js
+++ b/fec/fec/static/js/modules/data-map.js
@@ -404,7 +404,8 @@ function buildStateTooltips(svg, path, instance) {
   svg
     .selectAll('path')
     .on('mouseenter', function(d) {
-      if (instance.getStateValue(d.id) && instance.getStateValue(d.id) !== 0) {
+      let thisValue = instance.getStateValue(d.id);
+      if (thisValue && thisValue !== 0) {
         this.parentNode.appendChild(this);
         let html = tooltipTemplate({
           name: fips.fipsByCode[d.id].STATE_NAME,
@@ -414,9 +415,17 @@ function buildStateTooltips(svg, path, instance) {
         moveTooltip(tooltip);
       }
     })
-    .on('mousemove', function() {
-      if (tooltip.style('display') != 'none') {
+    .on('mousemove', function(d) {
+      let thisValue = instance.getStateValue(d.id);
+      if (thisValue && thisValue !== 0) {
+        let html = tooltipTemplate({
+          name: fips.fipsByCode[d.id].STATE_NAME,
+          total: '$' + Math.round(instance.getStateValue(d.id)).toLocaleString()
+        });
         moveTooltip(tooltip);
+        tooltip.style('display', 'block').html(html);
+      } else {
+        tooltip.style('display', 'none');
       }
     });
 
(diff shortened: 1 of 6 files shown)


fec/fec/static/js/modules/data-map.js
```diff
@@ -406,7 +403,7 @@ function buildStateTooltips(svg, path, instance) {
   // Go through our svg/map and assign the mouse listeners to each path
   svg
     .selectAll('path')
-    .on('mouseover', function(d) {
+    .on('mouseenter', function(d) {
```
patphongs: @rfultz Looks like IE is not removing the tooltip on `mouseenter` of a state with `colorZero`. The image below shows the bug. The tooltip still shows the data for Nebraska, the previous state that was entered into before going into South Dakota, which has zero contributions.

Image from iOS (6)

rfultz: …and that's not Nebraska that's white…

dorothyyeager:
Am testing in IE at @patphongs request. It all seems to be working. I noticed the list of contributors when you click a state's total usually, but not always, returns in reverse alphabetical order. I also noticed the search box for the map defaults to President Trump (rather than a blank box). Wasn't sure if these things were intentional or not, so mentioning. The important thing is that it's working.

rfultz:
@johnnyporkchops, it's everybody's favorite task!

PR #3302, opened 2019-10-27
Reaction box for WCCF, simplify reaction box implementation

### LINKED ISSUE #3237: Implement reaction box for WCCF feature
**What we're after:** To support the release and track usability needs, we should implement the reaction box below the WCCF to solicit for feedback. 

This feedback will be tracked and recorded by @JonellaCulmer 

**Completion criteria:**
- [ ] See previous implementation and confirm that questions/form boxes used apply to this feature. Last used on line chart on data landing page and raising and spending browse data pages. Confirm questions with @JonellaCulmer 
- [ ] Implement reaction box that solicits for and sends feedback to the FEC github repo. 
- [ ] Open up follow-up ticket to take down reaction box approximately 2 months from the time of the launch.

### Reaction box screenshot
Screen Shot 2019-10-07 at 11 09 36 AM

johnnyporkchops:
## Summary
- Resolves #3237 

- Implement the reaction box  under WCCF widget on data/raising-by-the-numbers
- Simplify the process for adding/ removing reaction boxes which now only requires including the macro and a reference to `reaction-box.js` in a template.
- Add comments with instructions in `reaction-box.js`
- Remove residual code leftover from since-removed reaction boxes
- Put WCCF-reaction-box under feature flag (the flag is nested in the WCCF flag, so they do not have to be turned on/off separately. 

- [ ] Add `FEC_FEATURE_WCCF_REACTION` env var to all environments using CF Update-User-Provided-Service (UUPS)  **Note:** The env var has been set on feature-space only for testing
- [ ] Should we consider implementing a simple `templatetag` for reaction boxes in templates other than Jinja templates (html Django templates). With this PR, a reaction box could be added to non Jinja tempates by simply hardcoding the macro's html into a template and replacing its `{{ }}` tags with hardcoded values . But a Django templatetag would be more convenient for developers and a more permanent solution.

Contributions by State widget (WCCF) on `data/raising-bythenumbers/`

	modified:   data/templates/browse-data.jinja
	modified:   data/templates/landing.jinja
	modified:   data/templates/partials/browse-data/raising.jinja
	modified:   data/templates/partials/browse-data/spending.jinja
	deleted:    fec/static/js/modules/reaction-box.js
	modified:   fec/static/js/pages/bythenumbers.js
	modified:   fec/static/js/pages/data-browse-data.js
	modified:   fec/static/js/pages/reaction-box.js

## How to test
  - checkout branch `feature/3237-reaction-box-to-wccf`
 -  `npm run build`
 -  `export FEC_FEATURE_WCCF_REACTION=1`
 -  `export FEC_FEATURE_CONTRIBUTIONS_BY_STATE=1`
 -  `./manage.py runserver`
-  Go to http://localhost:8000/data/raising-bythenumbers/ and try to submit a reaction
   - You must use `localhost:8000`, because the recaptcha does not work on `127.0.0.1:8000`
   - Submissions will return a 500 error locally and you will get the, "Oops, something went wrong..." message", but this means the form is actually working
- You can test submitting a real test reaction on feature space at:
 https://fec-feature-cms.app.cloud.gov/data/raising-bythenumbers/ . 
  - See https://github.com/fecgov/fec/issues to confirm submission and close the test issue

In addition to implementing the reaction-box for WCCF, this PR also simplifies the process for adding reaction-boxes to Jinja templates. 
- You can test this by:
  - Adding one or more reaction-box macros to a template
  - Including `reaction-box.js` in the template (preferably in an the extra js block at the bottom)

One last important test is to confirm that, with this new technique,  two (or more) reaction boxes on one page, work properly. This can be tested  at:
 - https://fec-feature-cms.app.cloud.gov/data/browse-data/?tab=raising
 - https://fec-feature-cms.app.cloud.gov/data/browse-data/?tab=raising

### DIFF
--- a/fec/data/templates/raising-bythenumbers.jinja
+++ b/fec/data/templates/raising-bythenumbers.jinja
@@ -1,5 +1,6 @@
 {% extends "layouts/sidebar-page.jinja" %}
 {% import "macros/bythenumbers.jinja" as breakdowns %}
+{% import 'macros/reaction-box.jinja' as reaction %}
 
 {% block css %}
 <link rel="stylesheet" type="text/css" href="{{ asset_for_css('data-landing.css') }}" />
@@ -50,6 +51,9 @@
 {# <script defer id="gov_fec_contribs_by_state_script" src="{{ asset_for_js('widgets/contributions-by-state.js') }}"></script> #}
 {# <iframe src="/widgets/contributions-by-state" id="fec-gov-contribs-by-state"></iframe> #}
 {% include 'partials/widgets/contributions-by-state.jinja' %}
+{% if FEATURES.wccf_reaction %}
+{{ reaction.reaction_box('contributions_by_state', 'raising-by-the-numbers') }}
+{% endif %}
 </section>
 {% endif %}
 {% endblock %}
@@ -82,4 +86,5 @@
   }
   </script>
   <script src="{{ asset_for_js('bythenumbers.js') }}"></script>
+  <script src="{{ asset_for_js('reaction-box.js') }}"></script>
 {% endblock %}
--- a/fec/fec/static/js/pages/reaction-box.js
+++ b/fec/fec/static/js/pages/reaction-box.js
@@ -1,16 +1,14 @@
 'use strict';
 
-// THIS MODULE IS CURRENTLY NOT IN USE.
-//
-// This requires modifications to `/data/views.py` feedback view
-// to post as a Github issue.
-//
-// Previously implemented here (needs a port to this Django project):
-// https://github.com/18F/openFEC-web-app/blob/develop/openfecwebapp/views.py#L302
+// // This required modifications to `/data/views.py`
+// // to post as a Github issue.
+// //
+// // Previously implemented here (ported to this Django project):
+// // https://github.com/18F/openFEC-web-app/blob/develop/openfecwebapp/views.py#L302
 
 var $ = require('jquery');
-var helpers = require('./helpers');
-var analytics = require('./analytics'); // TODO - move this to Tag Manager?
+var helpers = require('../modules/helpers');
+var analytics = require('../modules/analytics'); // TODO - move this to Tag Manager?
 
 function ReactionBox(selector) {
   this.$element = $(selector);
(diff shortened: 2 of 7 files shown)


fec/data/templates/raising-bythenumbers.jinja
```diff
@@ -82,4 +86,5 @@
   }
   </script>
   <script src="{{ asset_for_js('bythenumbers.js') }}"></script>
+  <script src="{{ asset_for_js('reaction-box.js') }}"></script>
```
rfultz: I'm interested in your (and others') thoughts on whether it makes sense to include both macros/reaction-box.jinja at the top of this file _and_ the reaction-box.js at the bottom, or if it would be cleaner (and easier later) to have the macro also include the script. It may mean including a <script> in the middle of the page, but 🤷‍♂? One downside would be if we have multiple reaction boxes on one page, we'd then be including the script inside each instance of the macros (i.e., multiple times in the same page)

fec/fec/static/js/pages/reaction-box.js
```diff
@@ -111,4 +109,34 @@ ReactionBox.prototype.handleReset = function() {
   this.$textarea.val('');
 };
 
-module.exports = { ReactionBox: ReactionBox };
+/* To implement a reaction box:
```
rfultz: If these comments are in JSDoc format, editors will show them to us when we hover and enter a new one. (I can show you)
`/**`
` * Text`
` * @example Text for the example`
` */`

fec/fec/static/js/pages/reaction-box.js
```diff
@@ -1,16 +1,14 @@
 'use strict';
 
-// THIS MODULE IS CURRENTLY NOT IN USE.
-//
-// This requires modifications to `/data/views.py` feedback view
-// to post as a Github issue.
-//
-// Previously implemented here (needs a port to this Django project):
-// https://github.com/18F/openFEC-web-app/blob/develop/openfecwebapp/views.py#L302
+// // This required modifications to `/data/views.py`
```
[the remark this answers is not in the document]
johnnyporkchops: The reaction-boxes.js in modules is deleted in this branch. So we could put it there. But if we exported it with `module exports`, then it would have to be required in  script to access it, not in a page as a script src, like it is now....right?

fec/data/templates/raising-bythenumbers.jinja
johnnyporkchops: I like that idea because it makes it even more simple to add a reaction box. Pat and I were talking about that but were not sure how we would go about bundling the macro and the script--this idea might make sense--I am not sure how often we have more than one on a page so it might be rare.

fec/fec/static/js/pages/reaction-box.js
johnnyporkchops: @rfultz Thanks for the demo on that, I will implement JSdocs comments for this

johnnyporkchops:
@rfultz , I made a first pass at adding JS docstrings to `reactionBox.js`, let me know how I did and I will refine.

fec/fec/static/js/pages/reaction-box.js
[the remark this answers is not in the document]
rfultz: Maybe? We could use it either way. In modules would make it more consistent with our other scripts, etc. Exporting it makes it more like a Class (as far as JavaScript has classes) If we go the exports route, we'd remove the part at the bottom where it creates a new one of itself in the same file.

fec/data/templates/raising-bythenumbers.jinja
patphongs: It is possible that the reaction box may be used multiple times on a page. Generally it's used to gage the reaction of a new feature but only up for a few weeks before we turn it off. Putting the JS inside the jinja template would make it more portable, but I'd like to avoid the JS redundancy. I think we should just leave it as is, my two cents :)

fec/fec/static/js/pages/reaction-box.js
```diff
@@ -111,4 +124,38 @@ ReactionBox.prototype.handleReset = function() {
...
-module.exports = { ReactionBox: ReactionBox };
+/**
+ * To implement a reaction box:
+ * Add a reaction-box jinja macro to a template
+ * Include a reference to this JS file in the parent template(preferably in extra JS block)
+ * (The below function will use the name/location values of any
+ *  reaction box on the page to initiate it as a new ReactionBox())
+ */
+
+/**
```
patphongs: Nice docs @johnnyporkchops !

Issue #3149, opened 2019-09-05
Create Example Page template in Wagtail


dorothyyeager:
### Summary

To assist with our upcoming work in making disclaimer examples and keep Wagtail tidy and easy for content managers, we'd like to make a new template similar to the reporting example template, but called "Example page"

### Proposed Behavior / Feature

We'd like the template to work like this one (https://fec-prod-proxy.app.cloud.gov/admin/pages/add/home/reportingexamplepage), but be called "Example page." We'd like to have a few other features added, such as embeddable video and internal/external buttons.

Here's a screenshot of the reporting example template:

![image](https://user-images.githubusercontent.com/24437369/65260926-31ce3e80-dad5-11e9-85be-ccd65118792c.png)

For this template, we'd like to add the following options to what's already in the reporting example template:
- image into the body portion (similar to what's available in custom/resource page templates)
- video embed into the body portion  (similar to what's available in custom/resource page templates)
- internal button into the body portion (See #2935 for background on this button)
- external button into the body portion  (See #2935 for background on this button)
- "Example" added as an option into "pre-title" drop down

### Use Case

Users who will love the examples we are making of proper disclaimers. (Currently being drafted in #3148 )

## Completion criteria ##
- [ ] Add the requested features listed above into the requested new template (based off the reporting example template)
- [ ] Testing by content team to ensure it has what we need for the disclaimer examples in #3148

patphongs:
@dorothyyeager Can you please highlight the elements you want to be present or added to this new "Example page" template?

dorothyyeager:
I've updated the ticket with a screenshot and list of the elements we're looking for in this new template.

johnnyporkchops:
@dorothyyeager. Would it make sense to add the additional features to the ReportingExamples template and that could then be used for the disclaimer examples as well?

dorothyyeager:
I think so, yes.
Sent from my iPhone

On Oct 27, 2019, at 11:41 PM, John Carroll <notifications@github.com<mailto:notifications@github.com>> wrote:


@dorothyyeager<https://github.com/dorothyyeager>. Would it make sense to add the additional features to the ReportingExamples template and that could then be used for the disclaimer examples as well?

dorothyyeager:
Clarifying - If we can make the choice of "Example" available along with "Example Scenario" and "Reporting Example" that would be great. We must have the ability to embed videos too. Thanks @johnnyporkchops

johnnyporkchops:
[...]

PR #3338, opened 2019-11-07
Add internal/external links to RecordPage template

### LINKED ISSUE #2935: Wagtail wish list: Add internal and external link buttons to Record and Example page templates
### Summary

Often the content team needs to place links on reporting examples and in Record articles. It would be nice to have the internal and external button features available on those two page templates. 

### Use Case

Doing this would make the links stand out more.

### Completion Criteria

Add the internal button and external buttons to these templates in Wagtail: 
- [ ] Record page
- [x] New example page from (#3149)

Here's a screenshot from a different Wagtail template (Resource page) that has these options:
image

When opened in Wagtail, here's what they look like:
image

image

Here's what they look like when rendered (both buttons look the same):
image

johnnyporkchops:
## Summary 
The `RecordPage` model   inherits from `ContentPage` model with uses `stream_factory` as its body section. So I added internal and external link buttons here.

- Resolves #2935

This will effect all page models inheriting from ContentPage model, but only to the extent that they will now also have the new internal and external link buttons.
 - modified:   home/models.py

## Screenshots

Screen Shot 2019-11-07 at 12 11 55 PM

## Related PRs
https://github.com/fecgov/fec-cms/issues/3149:

## How to test
- Checkout and run branch
- Create page of type, RecordPage and confirm and test the external and internal link buttons
- Publish or Preview to confirm it works 
____

### DIFF
--- a/fec/home/models.py
+++ b/fec/home/models.py
@@ -64,7 +64,9 @@ stream_factory = functools.partial(
         ('image', ImageChooserBlock()),
         ('table', TableBlock(table_options=core_table_options)),
         ('custom_table', CustomTableBlock()),
-        ('contact', ContactInfoBlock())
+        ('contact', ContactInfoBlock()),
+        ('internal_button', InternalButtonBlock()),
+        ('external_button', ExternalButtonBlock()),
     ],
 )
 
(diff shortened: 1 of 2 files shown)


johnnyporkchops:
@dorothyyeager , this is on feature for testing

dorothyyeager:
Loving it! The Record articles are going to look so much better. I tested both buttons on one and here's how they turned out... (perfect!)
image

PR #3347, opened 2019-11-12
Corrected candidate mentioned query param

### LINKED ISSUE #3247: Candidate name lookup not working for Schedule E raw data
We should use candidate name lookup for Schedule E raw data not working as expected. 

https://www.fec.gov/data/independent-expenditures/?data_type=efiling&candidate_name=warr&is_notice=true

**Example:**
Screen Shot 2019-11-01 at 1 53 34 PM

**Completion criteria:**
- [ ] Correct the candidate mentioned filter to search and return appropriate results by candidate's name
- [ ] Open follow-up ticket to review the business logic for how we want users to use this filter, or create an additional filter to also search by candidate ID. Will our data support this change?

patphongs:
## Summary

- Resolves #3247 
_Fixes candidate mentioned search field by correcting query param._

While doing this PR, I found a bug that the `candidate_search` query param is not returning all possible results. Although this PR fixes the query on the front end, we should investigate why we don't see all possible results. Issue logged: https://github.com/fecgov/openFEC/issues/4086. 

-  Independent expenditures raw filings search

## Screenshots

Screen Shot 2019-11-12 at 12 11 08 PM

## How to test
- [ ] Go to independent expenditures raw efiling: http://localhost:8000/data/independent-expenditures/?data_type=efiling&is_notice=true
- [ ] Enter search term in "Candidate mentioned". It should filter results by the search term entered.

### DIFF
--- a/fec/data/templates/partials/independent-expenditures-filter.jinja
+++ b/fec/data/templates/partials/independent-expenditures-filter.jinja
@@ -22,7 +22,7 @@ Filter independent expenditures
 {% block efiling_filters %}
 <div class="filters__inner">
   {{ typeahead.field('committee_id', 'Spender name or ID', id_suffix='_raw') }}
-  {{ text.field('candidate_name', 'Candidate mentioned', id_suffix='_raw')}}
+  {{ text.field('candidate_search', 'Candidate mentioned', id_suffix='_raw')}}
   {{ support_oppose.checkbox('_raw') }}
   {{ date.field('expenditure_date', 'Expenditure date', id_suffix='_raw') }}
 </div>


patphongs:
> It appears to work as expected. The only potential problem I'm seeing is the search button. It appears to be disabled the entire time, when it should "activate" once the user starts typing in a name.

@JonellaCulmer I see the button activate when I start typing (see gif below). Maybe try doing a rebuild of your npm stuff again `npm run build`.

Nov-14-2019 14-39-02

patphongs:
This has been approved, merging

PR #3355, opened 2019-11-14
Feature/3353 fix statute thumbnail

### LINKED ISSUE #3353: Fix link to URL for image of FECA used on statutes page
### Summary

**What we are after:**
When I uploaded the new thumbnail of the FECA, I didn't realize that the statutes page was pulling off a static link. Right now the image is not working. @patphongs has figured it out and knows how to fix this.

### Related issues
- PR for changing out image - https://github.com/fecgov/fec-cms/pull/3334/files

Note: Image we want has been uploaded to cms at https://www.fec.gov/resources/cms-content/images/52usc.original.png and to dev at https://fec-dev-proxy.app.cloud.gov/resources/cms-content/images/52usc.original.original.png (I don't know why dev is adding original twice into the file name)

I'll attach here too in case you need to reupload

52usc original

### Completion criteria

- [ ] Adjust the "document thumbnail" coding in the jinja file to take the "static" out of the image URL
- [ ] Test this on stage after release

patphongs:
## Summary

- Resolves #3353 
_Removes `static` dir path from image thumbnail from document macro. These thumbnails should be controlled via uploads to Wagtail CMS rather than through static images._

-  Statutes landing page: http://localhost:8000/data/legal/statutes/

### Before
Image path is `/static/resources/cms-content/images/52usc.original.png`
Screen Shot 2019-11-14 at 4 59 21 PM

### After
Image path is `/resources/cms-content/images/52usc.original.png`
Screen Shot 2019-11-14 at 4 58 59 PM

## How to test
- [ ] Run the branch
- [ ] Go to the statues landing page: http://localhost:8000/data/legal/statutes/
- [ ] Make sure that the image path for the Federal Election Campaign Laws by the FEC thumbnail is going to `/resources/...`. It should not have `/static/` as part of the path since we are not using static assets for this thumbnail.
____

### DIFF
--- a/fec/data/templates/macros/document.jinja
+++ b/fec/data/templates/macros/document.jinja
@@ -1,7 +1,8 @@
 {% macro thumbnail(title, url, img="", size='', type="PDF") %}
   <div class="grid__item document-thumbnail">
     <a class="document-thumbnail__img" href="{{ url }}" target="_blank">
-      <img src="{{ static(img)}}" alt="{{ title }}">
+    {# Image resource is pulling from CMS image uploads #}
+      <img src="{{ img }}" alt="{{ title }}">
     </a>
     <span class="document-thumbnail__description">
       <a href="{{ url }}" target="_blank">{{ title }}</a>
--- a/fec/fec/settings/production.py
+++ b/fec/fec/settings/production.py
@@ -14,8 +14,10 @@ SESSION_COOKIE_HTTPONLY = True
 CSRF_COOKIE_SECURE = True
 CSRF_COOKIE_HTTPONLY = True
 
-# TODO(jmcarp) Update after configuring DNS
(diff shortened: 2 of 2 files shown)


patphongs:
> Should we be using relative paths or absolute? (For search engines, absolute)
> 
> For the link you provided, two of the images' src values started with "img/" instead of "/":

@rfultz So the image thumbnail files will be relative paths. But the path to the PDF documents will probably be absolute because these files are stored on govinfo. Technically, I would like to move these thumbnail images to being stored inside of Wagtail so that we can update them without a code change. Later on we can also update the other thumbnails to be on Wagtail as well.

patphongs:
Accidentally coupled these changes with other commits. Going to close this PR and do a new one

PR #3369, opened 2019-11-19
[MERGE WITH GITFLOW]Release/public 20191126

fec-jli:
This is for the sprint (10.5) release.


rfultz:
@patphongs should you be a reviewer here, too?

rfultz:
@pkfec should we add you as a reviewer since you contributed, too?

patphongs:
@rfultz Since I will be deploying the release, I don't need to be added as a reviewer. Thanks!

PR #3409, opened 2019-12-11
Add error message close button for downloads

### LINKED ISSUE #3339: Can't close the downloads box if there's an error message
**What we're after:** Usually there's a little X marks the spot to close the downloads box once a user is done with their download and wants to move on. It's just been observed that if there's an error message like in the screenshot below that there is no way to close the box. The user is forced to refresh the page. 

**Completion criteria:**
- [ ] Allow users to close the downloads box regardless of the content within it

**Before:** 
Screen Shot 2019-11-07 at 4 03 22 PM

**After:**
Screen Shot 2019-11-21 at 1 27 56 PM

rfultz:
## Summary

- Resolves #3339 
Added a close button for download window error messages.

Added it to the error message in the template for the download error message text and styled it

## Screenshots
image

## Related PRs
None

## How to test
- Pull the branch
- `npm i` `npm run build`
- Change your API key to one with limits that you can hit, or otherwise ensure you get an error message
- `./manage.py runserver`
- Check an [Individual Contributions page](http://127.0.0.1:8000/data/receipts/individual-contributions/?committee_id=C00577130&committee_id=C00696948&contributor_employer=Amazon&two_year_transaction_period=2020), limit the results so the Export button is active
- Click the Export button at the top right
- The error message should appear at the bottom
- Clicking the X to the right of the message should close the window _and_ re-enable the export button
- We should check elsewhere where that kind of error may be found
- We should check other areas of the site that are similar to the downloads window and error message to make sure they're unaffected
____

### DIFF
--- a/fec/fec/static/js/modules/download.js
+++ b/fec/fec/static/js/modules/download.js
@@ -189,12 +189,14 @@ DownloadItem.prototype.handleServerError = function(xhr) {
     this.$body.html(
       '<div class="message message--alert message--mini">' +
         'Sorry, you have exceeded your maximum downloads for the hour. Please try again later.' +
+        '<button class="js-close button--cancel download__cancel"><span class="u-visually-hidden">Remove</span></button>' +
         '</div>'
     );
   } else {
     this.$body.html(
       '<div class="message message--alert message--mini">' +
         'Sorry, you have either exceeded your maximum downloads for the hour or encountered a server error. Please try again later.' +
+        '<button class="js-close button--cancel download__cancel"><span class="u-visually-hidden">Remove</span></button>' +
         '</div>'
     );
   }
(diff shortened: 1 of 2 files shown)


JonellaCulmer:
@rfultz Can you please include instructions on how to change my API key and to which one has a limited that would easily trigger the error message?

fec/fec/static/js/modules/download.js
```diff
@@ -209,6 +211,14 @@ DownloadItem.prototype.handleServerError = function(xhr) {
   // Tell the container to subtract an item, but preserve the DOM itself
   // so that the message stays visible
   this.container.subtract(true);
+
+  // Add error close button functionality
+  $(this.$body)
+    .find('.js-close')
+    .on('click', this.handleCloseErrorClick.bind(this));
+};
```
patphongs: Can you have the error message removed each time the click happens? Right now it's adding itself multiple times when you click the export button.

Screen Shot 2019-12-16 at 5 36 45 PM

JonellaCulmer: @patphongs That's how it works for other exports, each time, the assumption is that the user changed some things and so it allows a new "export" to appear. Perhaps we can limit this specific error to only appear once? That is, of course, unless the user closes it and hits export again, then it would appear pop-up again.

rfultz: Commenting after our conversation last week: we decided to let the functionality stand where a user could trigger a stack of error messages (click, error message 1, click error message 2, click, error message 3...), biggest reasons being that
- it's the current functionality and
- with identical messages, a user may trigger an additional error that has the same text as the previous messages. Removing the first message to immediately replace it with an identical message would have the appearance to the user that nothing had happened with every successive click.

Issue #3426, opened 2019-12-23
Server error on C00140590 committee profile page 


pkfec:
**What happened :** C00140590 committee profile page is throwing a server error.
- url : https://www.fec.gov/data/committee/C00140590/
- Note : C00140590is a party committee and should not have an associated candidate. 

**Where:**  Found in production. This server error in consistent across all environments

**Screen shot:**  : 

<img width="982" alt="Screen Shot 2019-12-23 at 6 54 39 PM" src="https://user-images.githubusercontent.com/11650355/71385437-b6899680-25b5-11ea-9f82-32b60361f17f.png">

**Reference PR**: https://github.com/fecgov/fec-cms/pull/3299

jason-upchurch:
@pkfec @fec-jli this is the stack trace I encountered:

```
File "/Users/jupchurch/dev/cms/fec-cms/fec/data/templates/macros/cycle-select.jinja", line 53, in template
    {% for each in cycles | sort(reverse=True) %}
  File "/Users/jupchurch/.pyenv/versions/cms-dev/lib/python3.6/site-packages/jinja2/filters.py", line 278, in do_sort
    return sorted(value, key=key_func, reverse=reverse)
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```

fec-jli:
The reason is cycles_has_activity return a 'None' value in ofec_committee_history_mv
wrong example query: select * from ofec_committee_history_mv where committee_id='C00140590'
cycles_has_activity=[1982, 1984, 1986, 1988, 1990, 1992, 1994, 1996, 1998, 2000, 2002, 2004, 2006, 2008, 2010, 2012, 2014, 2016, 2018, 2020, NULL]

the correct query: select * from ofec_committee_history_mv where committee_id='C00105668'
cycles_has_activity=[1978, 1980, 1982, 1984, 1986, 1988, 1990, 1992, 1994, 1996, 1998, 2000, 2002, 2004, 2006, 2008, 2010, 2012, 2014, 2016, 2018, 2020]

solution:
1)We will create an separated issue to fix ofec_committee_history_mv.
2)modify cms load_committee_history function in views.py
to clean cycles_has_activity, remove 'None' value in cycles_has_activity
`    committee['cycles_has_activity'] = list(filter(None, committee.get('cycles_has_activity')))
`

fec-jli:
for some reasons, my local cms could not pass `npm run test-single`,
so I have to force push my branch. 
[...]

PR #3427, opened 2019-12-24
Remove 'None' value from list cycles_has_activity in ofec_committee_h…

fec-jli:
## Summary (required)
C00140590 committee profile page is throwing a server error.

- Resolves #[[_3426_](https://github.com/fecgov/fec-cms/issues/3426)]
modify cms load_committee_history function in views.py
to clean cycles_has_activity, remove 'None' value in cycles_has_activity
committee['cycles_has_activity'] = list(filter(None, committee.get('cycles_has_activity')))

committee profile page
 

## How to test
1) checkout branch
2)pytest
./manage.py test
3)npm run test-single
4) test url, compare with prod.
(wrong one)
prod: https://www.fec.gov/data/committee/C00140590/
local: http://127.0.0.1:8000/data/committee/C00140590/

(correct one should not affected)
prod: https://www.fec.gov/data/committee/C00105668/
local: http://127.0.0.1:8000/data/committee/C00105668/

We should create an separated issue to fix ofec_committee_history_mv

### DIFF
--- a/fec/data/tests/test_committee.py
+++ b/fec/data/tests/test_committee.py
@@ -551,7 +551,8 @@ class TestCommitteeApiCalls(TestCase):
             'party_full': 'OTHER',
             'designation_full': 'Principal campaign committee',
             'cycles': [],
-            'cycles_has_activity': [2018],
+            # make sure we clean 'None' value in cycles_has_activity correctly
+            'cycles_has_activity': [2018, None],
             'committee_type': 'P',
             'party': 'OTH',
             'designation': 'P',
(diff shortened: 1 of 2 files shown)


fec-jli:
> @fec-jli if we leave the logic as a control on the cms side rather than (or in addition to) sql, then should we also filter for any non-integer value rather than only None? For expediency I am approving now.

Thanks @jason-upchurch review it. for now, we only get 'None' value from sql query return, not sure if we will get other value. I will talk to database team to see any concern. Thanks.

jason-upchurch:
> > @fec-jli if we leave the logic as a control on the cms side rather than (or in addition to) sql, then should we also filter for any non-integer value rather than only None? For expediency I am approving now.
> 
> Thanks @jason-upchurch review it. for now, we only get 'None' value from sql query return, not sure if we will get other value. I will talk to database team to see any concern. Thanks.

Thanks @fec-jli I think the datatype in the db is specified as `integer` so I agree nothing else should come back (because we rely on the root for correct behavior).

fec/data/tests/test_committee.py
```diff
@@ -571,11 +572,18 @@ def test_load_committee_history_uses_last_cycle_has_activity(
         assert all_candidates == []
         assert committee == first_row_data
 
+        # check if cycles_has_activity include 'None' item
+        # in function load_committee_history() in views.py (line 524)
+        # we remove 'None' value in cycles_has_activity
+        # so make sure cycles_has_activity not include 'None' value anymore,
+        # assert should is Flase.
+        assert any(x is None for x in committee['cycles_has_activity']) is False
```
lbeaufort: Thanks for adding test coverage, @fec-jli! Would `assert None not in committee['cycles_has_activity']` be more clear?

PR #3457, opened 2020-01-09
Trigger resize when toggling content

### LINKED ISSUE #3344: Total contributed header not displaying correctly on Raising tab on cmte profile pages
**What we're after:** On candidate and committee profiles pages, the Raising tab, on the Individual contributions datatable (state), the front-end is not properly displaying the Total contributed column header. The column header is off to the left and only once clicked on to change the sort does it fix itself. 

**Completion criteria:**
- [ ] Fix the column header to display in the proper place

**Example:**

rfultz:
## Summary

If it dynamic content (e.g. datatables) wasn't visible when it was created, its width will be 0px, so we'll trigger a resize when we toggle it on (or off) to trigger its layout to refresh.

- Resolves #3344 

- Added a resize event that will fire after any content has been toggled using the `js-panel-toggle-control` class name  

## Screenshots

image

## Related PRs
None

## How to test
- Pull branch, `npm i`, `npm run build`, `./manage.py runserver` like normal.
- Go to a [candidate's raising page](http://127.0.0.1:8000/data/candidate/P60007168/?cycle=2020&election_full=true&tab=raising)
- Under Individual Contributions, switch to the State tab
- Make sure the header for the newly-visible table is the same width as the table content below it (rather than only as wide as the text content requires it to be—see the screenshots from [the original ticket](https://github.com/fecgov/fec-cms/issues/3344))
- Toggle back & forth to make sure nothing else is affected
- Resize the window to make sure that still works fine (can't imagine it wouldn't)
- Check other tabs/toggles that use dynamic content like a datatables table
- Check other tabs/toggles that don't use dynamic content to make sure they're still fine
- Approve
- If you're the last approval, merge, delete

____

### DIFF
--- a/fec/fec/static/js/modules/toggle.js
+++ b/fec/fec/static/js/modules/toggle.js
@@ -16,6 +16,9 @@ module.exports = {
           $('#' + $input.attr('value')).attr('aria-hidden', 'true');
         });
         $('#' + $elm.attr('value')).attr('aria-hidden', 'false');
+        // Since we're toggling content that may not have been visible yet,
+        // let's trigger a resize so the page can adapt to it having a non-zero width:
+        $(document).trigger('resize');
       });
     });
   }


rfultz:
@JonellaCulmer, should we get another ticket put together to address that the columns aren't exactly the same width? Also, I don't like that there's some kind of crumb visible above the California line and the header's icons (underscore/mdash, down arrow)

JonellaCulmer:
> @JonellaCulmer, should we get another ticket put together to address that the header's columns aren't exactly the same width as the table's?
> 
> Also, I don't like that there's some kind of crumb visible above the California line and the header's icons (underscore/mdash, down arrow)

@rfultz If it's a relatively small change I think we can do it in the same PR. And I just noticed, too, what I'm calling a little "nub" underneath the border. But I defer to you on whether it's one PR or two/two tickets.

PR #3475, opened 2020-01-22
Remove reaction box from WCCF feature

### LINKED ISSUE #3460: Remove reaction box from WCCF feature
**What we're after:**
We've collected two months of reactions from users and can now remove it from the site. 

We can either turn ~on~ off the feature flag for it, if one is present, or remove it entirely. 

**Completion criteria:**
- [x] Determine best path forward for removing the reaction box and implement
- [x] Reaction box is removed from the page

johnnyporkchops:
## Summary 
Resolves #3460

- Remove reaction box macro from raisingbythenumbers.jinja
- Remove reaction box script tag from raisingbythenumbers.jinja
- Remove feature flag env var from settings/base.py

- [x] Followup work: remove `FEC_FEATURE_WCCF_REACTION` from `User Provided Service` vars on Cloud Foundry for prod, dev. stage, feature environments

modified:   data/templates/raising-bythenumbers.jinja
modified:   fec/settings/base.py

## Related PRs
https://github.com/fecgov/fec-cms/pull/3302

## How to test
Include any information that may be helpful to the reviewer(s).
- checkout `feature/3460-remove-wccf-reaction-box`
- runserver
- confirm that reaction box is NOT under WCCF map on http://127.0.0.1:8000/data/raising-bythenumbers/
____

### DIFF
--- a/fec/data/templates/raising-bythenumbers.jinja
+++ b/fec/data/templates/raising-bythenumbers.jinja
@@ -1,6 +1,5 @@
 {% extends "layouts/sidebar-page.jinja" %}
 {% import "macros/bythenumbers.jinja" as breakdowns %}
-{% import 'macros/reaction-box.jinja' as reaction %}
 
 {% block css %}
 <link rel="stylesheet" type="text/css" href="{{ asset_for_css('data-landing.css') }}" />
@@ -51,9 +50,6 @@
 {# <script defer id="gov_fec_contribs_by_state_script" src="{{ asset_for_js('widgets/contributions-by-state.js') }}"></script> #}
 {# <iframe src="/widgets/contributions-by-state" id="fec-gov-contribs-by-state"></iframe> #}
 {% include 'partials/widgets/contributions-by-state.jinja' %}
-{% if FEATURES.wccf_reaction %}
-{{ reaction.reaction_box('contributions_by_state', 'raising-by-the-numbers') }}
-{% endif %}
 </section>
 {% endif %}
 {% endblock %}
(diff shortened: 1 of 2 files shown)


JonellaCulmer:
@johnnyporkchops Had to build in the feature flag for the map to appear, so I'm not sure if there's anything else I need to do in order to verify that the reaction box is gone. If so, let me know. Otherwise, I don't see it and we're good to go!

johnnyporkchops:
> Do we want to remove it completely or just turn it off for now?
> (This would be fine to remove it 'forever' but if we just want to turn it off, we'd just need to set FEC_FEATURE_WCCF_REACTION to false)

@rfultz , yes that is the way to turn off the feature flag. The issue says to "either turn on off the feature flag for it, if one is present, or remove it entirely". I am assuming that means we have no further use for it. 
@JonellaCulmer, is there future work, like adding the datatable view, that would require us to reenable the reaction box.
If we have no use for it in the future, I would opt to remove the code completely, rather than leave it  there.

JonellaCulmer:
> @JonellaCulmer, is there future work, like adding the datatable view, that would require us to reenable the reaction box?
If we have no use for it in the future, I would opt to remove the code completely, rather than leave it there.

I defer to front-end. We may or may not need it in the future. I can't say for sure either way now.

johnnyporkchops:
Why don't we just remove the code from the templates and base.py and leave the variable in CF UPS.  We can add back on with a PR, which would probably go along with any iteration substantial enough to to require user feedback.

johnnyporkchops:
Based on conversations in slack and here, I am going to also remove the feature flag env vars from all CF environments as well. 
@rfultz I accidentally re-requested your review on this PR.

PR #3531, opened 2020-02-07
Update F3X to F24 when 24- and 48-Hour Reports selected

### LINKED ISSUE #3463: Provide Form 24 data when Form 3X is selected for 24/48 hour notices
### What we're after:
On the IE datatable, there are no results when selecting the `24- and 48-Hour Reports` toggle and `Form 3X` is checkbox.

### Root cause
The root cause is `24- and 48-Hour Reports` that are `Form 3X` must be filtered using F24 instead of F3X, which it currently does.

### Requirement
When 24-48 hour toggle is selected, the API call should filter filing_form as F24. Example API call: https://api.open.fec.gov/v1/schedules/schedule_e/?api_key=DEMO_KEY&sort_hide_null=false&sort_nulls_last=true&data_type=processed&is_notice=true&filing_form=F24&sort=-expenditure_date&per_page=30
Screen Shot 2020-01-16 at 11 49 02 AM

### Completion criteria:
- [x] Results appear when `24- and 48-Hour Reports` toggle is selected and Form 3X checkbox is checked. 
- [x] The datatable results only include filing_form records that are F24

patphongs:
## Summary (required)

- Resolves #3463 
_Update the regularly scheduled report form from F3X to F24 when 24- and 48-Hour Reports are selected._

-  Independent expenditures datatable: https://www.fec.gov/data/independent-expenditures/

## Screenshots

-No screenshots, just modified fetch call._

## How to test
- [ ] Checkout this branch
- [ ] npm run build
- [ ] Go to the IE datatable: http://localhost:8000/data/independent-expenditures/
- [ ] Check the Form 3X checkbox under Regularly scheduled report form. Make sure to leave Report type toggle with the `24- and 48-Hour Report` selected.
- [ ] Check the AJAX call in the console to make sure that it's querying for `filing_form=F24`.
- [ ] Change the Report type toggle to have `Regularly scheduled reports` selected.
- [ ] Check the AJAX call in the console to make sure that it's querying for `filing_form=F3X`.
____

### DIFF
--- a/fec/fec/tests/js/tables.js
+++ b/fec/fec/tests/js/tables.js
@@ -550,4 +550,34 @@ describe('data table', function() {
       ).to.be.above(0);
     });
   });
+
+  describe('Update to F24', function() {
+    beforeEach(function() {
+      this.table.xhr = null;
+      $('body').append(
+        '<div id="is_notice-toggle_processed"><ul class="dropdown__selected"></ul><input id="filing-form-f3x_processed" /></div>'
+      );
+    });
+
+    it('filing form updates to F24 when is_notice is true', function() {
+      var serialized = {
+        is_notice: 'true',
+        filing_form: ['F3X']
+      };
+      this.table.filterSet = {
+        serialize: function() {
+          return serialized;
+        },
+        fields: ['is_notice', 'filing_form']
+      };
+      this.table.filterSet.isValid = true;
+      var callback = sinon.stub();
+      this.table.fetch({}, callback);
+      expect(this.table.filters).to.deep.equal(serialized);
+      expect(
+        this.table.filters.filing_form,
+        'Expected filing form to be F24'
+      ).to.deep.equal(['F24']);
+    });
+  });
 });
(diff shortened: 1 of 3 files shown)


JonellaCulmer:
@patphongs Looks good. A couple changes that I noticed. We should change the text from "Regularly scheduled report form" to just "Report form" since 24- and 48-reports aren't regularly scheduled. Can we also remove the dividing line between the two filters? Since they work together a bit.

patphongs:
> A couple changes that I noticed. We should change the text from "Regularly scheduled report form" to just "Report form" since 24- and 48-reports aren't regularly scheduled. Can we also remove the dividing line between the two filters? Since they work together a bit.

@JonellaCulmer These changes have been applied, can you please check and test again? Thanks!

fec/fec/tests/js/tables.js
```diff
@@ -550,4 +550,34 @@ describe('data table', function() {
       ).to.be.above(0);
     });
   });
+
+  describe('Update to F24', function() {
```
lbeaufort: Nice test @patphongs!

PR #3547, opened 2020-02-12
Feature/3514 presidential map 1

rfultz:
## Summary
- Resolves #3513 
- Resolves #3514 
- Closes #3507  
There's still quite a bit to do—especially with styles (size, color, flow, visibility, etc) but progress is, well, progressing.
Think of this as a check-in PR to get a feel for how it's going and to start verifying that data is working correctly.

## Impacted areas of the application
Added more components and functionality.
Also calling the official endpoints (hence closing 3507)

## Screenshots
image

## Related PRs
#3514 

## How to test
- pull branch
- Set feature flag: `export FEC_FEATURE_PRESIDENTIAL_MAP=True`
- `npm i`
- `npm run build`
- `./manage.py runserver`
- check http://127.0.0.1:8000/data/candidates/president/presidential-map/
- Play with the year toggle, click candidates

The left list should change when the years are clicked.
The map colors, breadcrumb, candidate details, and eventual right column should all update as candidates are changed.
The various download links should change their href values as candidates are changed (though the state download button doesn't toggle off.
____

PR #3556, opened 2020-02-18
Add feature flag, silence JS tests, address some feedback for presidential map

lbeaufort:
- Add feature flag for presidential map
- Fix formatting, silence tests for map
- Clean up python logic, fix default year per https://github.com/fecgov/fec-cms/pull/3547#discussion_r380817871
- Use `constants.states_excl_territories` for state list per https://github.com/fecgov/fec-cms/pull/3547#discussion_r380777664

**Feature flag info**
To turn on the feature, `export FEC_FEATURE_PRESIDENTIAL_MAP=True`
To turn it off, `unset FEC_FEATURE_PRESIDENTIAL_MAP`

You may want to add it to your bash profile so that you don't have to keep setting it.

### DIFF
--- a/fec/data/urls.py
+++ b/fec/data/urls.py
@@ -2,6 +2,7 @@ from django.conf.urls import url
 
 from data import views
 from data import views_datatables
+from fec import settings
 
 urlpatterns = [
     url(r'^data/$', views.landing),
@@ -15,9 +16,6 @@ urlpatterns = [
     url(r'^data/elections/$', views.elections_lookup),
     url(r'^data/raising-bythenumbers/$', views.raising),
     url(r'^data/spending-bythenumbers/$', views.spending),
-
-    # Presidential Campaign Finance Map
-    url(r'^data/candidates/president/presidential-map/$', views.pres_finance_map),
 
     # Feedback Tool
     url(r'^data/issue/reaction/$', views.reactionFeedback),
@@ -49,3 +47,9 @@ urlpatterns = [
 
     url(r'^widgets/aggregate-totals/$', views.aggregate_totals),
 ]
+
+if settings.FEATURES.get('presidential_map'):
+    # Presidential Campaign Finance Map
+    urlpatterns.append(
+        url(r'^data/candidates/president/presidential-map/$', views.pres_finance_map)
+    )
--- a/fec/fec/static/js/modules/data-map.js
+++ b/fec/fec/static/js/modules/data-map.js
@@ -1,3 +1,5 @@
+/* eslint-disable no-undef */
+
 'use strict';
 
 /**
@@ -537,3 +539,5 @@ function tooltipTemplate(obj) {
 module.exports = {
   DataMap
 };
+
+/* eslint-enable no-undef */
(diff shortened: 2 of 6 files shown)


fec/data/urls.py
```diff
@@ -2,6 +2,7 @@
 
 from data import views
 from data import views_datatables
+from fec import settings
```
rfultz: Nice! This will let us turn on/off other URLs through feature flags, too

fec/fec/static/js/widgets/pres-finance-map-box.js
```diff
@@ -1171,36 +1174,36 @@ PresidentialFundsMap.prototype.handleExportRaisingClick = function(e) {
   eg: function openDownloads() VS. const openDownloads = function()
   */
 
-  const openDownloads = function(){
-    console.log('callback')
-    
+  const openDownloads = function() {
```
rfultz: Will probably make this more specific to the PresFinMap rather than global to avoid other possible conflicts, but we can get to that for or after launch

fec/fec/static/js/modules/data-map.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-undef */
```
rfultz: Just noting that this is only to get things on dev and should _not_ go into production

fec/fec/static/js/widgets/pres-finance-map-box.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable no-undef */
+/* eslint-disable no-unused-vars */
```
rfultz: Just noting that these are only to get things on dev and should _not_ go into production

fec/fec/static/js/widgets/pres-finance-map-box.js
```diff
@@ -1171,36 +1174,36 @@ PresidentialFundsMap.prototype.handleExportRaisingClick = function(e) {
...
     var wS = this.scrollY,
-       hT = instance.downloadsWrapper.getBoundingClientRect().top + wS,
-       hH = instance.downloadsWrapper.offsetHeight,
-       wH = window.innerHeight;
-    if (wS > (hT+hH-wH)){
-     openDownloads()
+      hT = instance.downloadsWrapper.getBoundingClientRect().top + wS,
+      hH = instance.downloadsWrapper.offsetHeight,
+      wH = window.innerHeight;
+    if (wS > hT + hH - wH) {
```
rfultz: Would like to name these vars more human-friendly but no biggie for now

fec/fec/static/js/widgets/pres-finance-map-box.js
```diff
@@ -1309,3 +1312,6 @@ function logUsage(candID, electionYear) {
 }
 
 new PresidentialFundsMap();
+
+/* eslint-enable no-undef */
+/* eslint-enable no-unused-vars */
```
rfultz: Linter may require a blank line at the end. 🤞

fec/fec/static/js/modules/data-map.js
lbeaufort: @rfultz agreed that it seems prone to human error if we forgot to remove it - if you'd rather address the errors individually, feel free to commit here or on another iteration!

PR #3627, opened 2020-03-17
Feature/3600 presidential map 4

### LINKED ISSUE #3600: Make updates to presidential map based on quality assurance testing
**What we're after:** Following on the [quality assurance testing](https://github.com/fecgov/fec-cms/issues/3497) done, we have a list of items that need to be complete as part of the MVP. 

Action items: 
- [ ] Fix typos: 
            Cards: Raising: by the numbers, not Raising by the numberss
	    Spending: by the numbers, not Spending by the numbers
- [ ] Fix coverage dates. Some are appearing a day behind. (Ex. 12/30/2019 instead of 12/31/2019)
- [ ] In all places, (left and right columns and map and breadcrumbs) fix: “All Candidates”
- [ ] Fix map freeze on 2020 data and not consistently changing to 2016 when using the toggle
            Jonella will ask whoever put in the comment to note exactly what they did, or make a gif to show it happening. cc: @kathycarothers 
- [ ] Review candidate totals for both 2020 and 2016, they are not consistent with other pages such as the party aggregates cc: @PaulClark2 @fecjjeng 
Different totals observed for Trump, Warren, Sanders
- [ ] Default to Gradient 2020 all candidates, on initial load and when clicking “back” to the map
- [ ] Map not functioning in internet explorer
- [ ] When viewing state level, the toggle for Gradient/Bubbles should disappear, there is no choice for users at the state level
- [ ] In Raising accordion, change “Federal funds” to “Public funds”

**Completion criteria:**
- [ ] Update map according to the aforementioned list of action items.

rfultz:
## Summary

- Resolves #3600
Many things from the MoSCoW doc. This should be considered the release beta.

- Application-wise, only this map and WCCF may be affected.
- Technically any other data page with accordions could be affected but shouldn't (If the other data pages are impacted, it would be in that other accordions would be limited so only one item is open at a time and they wouldn't close correctly.)

## Screenshots
image

## Related PRs
None active

## How to test
I added basically everybody as a reviewer; feel free to remove or add whomever

Either pull, build, and run as normal, or
*Just check Feature*

- I didn't check off anything in the MoSCow doc so if one of you could, great! (Seemed like grading my own paper)
- The Musts should be taken care of (except the 'loading' visual feedback)
- Pretty sure most of the Shoulds and Coulds are in, too
- Check against different browser widths from as small as you can go (400-500px) up to normal large (1200?)

____


kathycarothers:
The map looks really good. I went through the doc and it seems like all the major items have been fixed. Here is what I noticed so far:

- The totals still seem to be off from what we noted in the testing document (I might have missed a discussion this one point) 
- Firefox and Chrome work well
- I had problems while using IE:

image


image

patphongs:
Is the spending accordion supposed to disappear from the map container when expanding raising? You can scroll to see it, but there’s no scroll bar unless you actively try and scroll (Pat)

accordion-interaction

rfultz:
> Is the spending accordion supposed to disappear from the map container when expanding raising? You can scroll to see it, but there’s no scroll bar unless you actively try and scroll (Pat)
> 
> accordion-interaction

Yeah, that's Chrome's default behavior. I could make it so the content _inside_ each accordion is scrollable instead of the accordion's parent being scrollable. Might not be a bad idea but I doubt it would be today—will require re-working the accordions and not using (here) the aria-accordion package we're using on the rest of the site. ?

dorothyyeager:
The methodology link needs to be changed to point to /presidential-map-data/ (rather than -help) so that people can test it properly. (The help link points to the old page, but that's been redrafted and put into resource page format.) Thanks!

rfultz:
> The methodology link needs to be changed to point to /presidential-map-data/ (rather than -help) so that people can test it properly. (The help link points to the old page, but that's been redrafted and put into resource page format.) Thanks!

I just made a commit to address the url

rfultz:
Changing this to closed, like its PR

PR #3695, opened 2020-04-17
Update sitemap-pdf for changes to files

dorothyyeager:
## Summary (required)

Updates the site map to pick up on changes made to the PDFs in it.

## How to test

- verify YYYY-MM-DD format is the one to use for pub dates. Dates should be in April or March 2020.


dorothyyeager:
I can fix it @jason-upchurch - kind of like to keep you as reviewer for this!

PR #3722, opened 2020-04-28
Set python get request timeout

patphongs:
## Summary

- Resolves https://github.com/fecgov/fec-accounts/issues/248
_Sets python get requests timeout to 90 seconds. This will ensure that requests are closed at a regular interval so that CMS PG connections do not build. If timeout is reached a 500 error page will come up and, this will appear in the logs:

```
raise ReadTimeout(e, request=request)
requests.exceptions.ReadTimeout: HTTPConnectionPool(host=‘localhost’, port=5000): Read timed out. (read timeout=90)
```

-  FEC python requests in api_caller.py

## How to test

- Follow instructions here to create local API load https://github.com/fecgov/fec-accounts/issues/248#issuecomment-618541364
- Connect local CMS to local API
- Run the CMS and try a page that uses a GET request from the api_caller such as a candidate page: http://localhost:8000/data/candidate/P80006117/
- Result should have timeouts after 90 seconds as seen in notes above

____

### DIFF
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -29,7 +29,8 @@ def _call_api(*path_parts, **filters):
 
     path = os.path.join(settings.FEC_API_VERSION, *[x.strip("/") for x in path_parts])
     url = parse.urljoin(settings.FEC_API_URL, path)
-    results = session.get(url, params=filters)
+    timeout = 90
+    results = session.get(url, params=filters, timeout=timeout)
 
     # Log the caller function and API endpoint
     current_frame = inspect.currentframe()
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -29,7 +29,9 @@ def _call_api(*path_parts, **filters):
 
     path = os.path.join(settings.FEC_API_VERSION, *[x.strip("/") for x in path_parts])
     url = parse.urljoin(settings.FEC_API_URL, path)
-    timeout = 90
+    # Timeout is set in seconds 
+    timeout = 90 
+    
     results = session.get(url, params=filters, timeout=timeout)
 
     # Log the caller function and API endpoint


fec/data/api_caller.py
```diff
@@ -29,7 +29,8 @@ def _call_api(*path_parts, **filters):
 
     path = os.path.join(settings.FEC_API_VERSION, *[x.strip("/") for x in path_parts])
     url = parse.urljoin(settings.FEC_API_URL, path)
-    results = session.get(url, params=filters)
+    timeout = 90
```
lbeaufort, line 32: Can you indicate with a comment or variable name that the unit is seconds? 
https://requests.readthedocs.io/en/master/user/quickstart/

patphongs: @lbeaufort Yes, I've added it just now, thanks

lbeaufort:
This is great progress, but I think we should have a follow-up issue or keep fecgov/fec-accounts#248 open to make sure a 90 second timeout will still address the amount of load we saw that prompted the original issue. 

We may need to timeout sooner, and the only way to test that is to generate load comparable to 4/15. @patphongs thoughts?

patphongs:
> This is great progress, but I think we should have a follow-up issue or keep [fecgov/fec-accounts#248](https://github.com/fecgov/fec-accounts/issues/248) open to make sure a 90 second timeout will still address the amount of load we saw that prompted the original issue.
> 
> We may need to timeout sooner, and the only way to test that is to generate conditions comparable to 4/15. @patphongs thoughts?

Thanks @lbeaufort, I agree we should test this more. I think setting a longer interval is fine for now, we can always fine tune this later, I've put in another ticket: https://github.com/fecgov/fec-cms/issues/3729

pkfec:
I paired with @fec-jli and tested the timeouts in local CMS. Great work @patphongs. Thank you so much for all the research and testing different options in CMS. I wish I could pair more with you on the work. Learned a lot.

fec-jli:
Thanks for your great work.
I test locally and try different timeout.
when setup timeout > 10 s, will see the `connection pool is full` message.
when timeout < 10 s, only see `timeout` message.
Thanks

PR #3824, opened 2020-06-08
Feature/3651 flake8 linting

### LINKED ISSUE #3651: Automate PEP8 linting tests
We should automate Python linting in the API and CMS like we do for JavaScript in the CMS.

See openFEC PR https://github.com/fecgov/openFEC/pull/4259 

- [ ] Add linting to pytests

rfultz:
## Summary

- Resolves #3651 
- Adds flake8 linting to `pytest` as mentioned in 

- Added flake8 and updated the config files accordingly
- Copied pytest flake8 rules to the `.flake8` config file that IDEs use
- Exempted the `.vscode/settings.json` from git so my/local settings don't override others' preferences
- Lots of linting fixes that were simple formatting (e.g., too many or too few blank lines, too few/many spaces)
- Several linting changes that are less clear to me—for those I added `# noqa ____` lines, exempting that one rule for that one line

## Screenshots

Should be _zero_ front-end differences

## Related PRs

branch | PR
------ | ------
feature/4258-add-flake8 | [Open FEC #4259](https://github.com/fecgov/openFEC/pull/4259)

## How to test

- Pull the branch
- `pip install -r requirements.txt`
- `npm i` (just in case? Just a habit for me)
- `npm run build` (another habit)
- Review file changes
  - Check **all** files with `# noqa` in them. Ideally every instance should be fixed and the noqa exception removed. There were a couple I removed that broke things so I'd like those more familiar with that code to double-check
  - Check indentation? I think they're all correct but there was one I had to adjust a couple times. Would be nice to have other eyes on that type of thing
- `pytest` should run all of the tests
- `pytest --linting` should run only the Python linting
- All tests should pass

## Next steps
To me, it feels like the python linting should also happen inside `npm run watch` and `npm run build-production` but I'm still working through that. Not a required part of this ticket

____


conftest.py
```diff
@@ -0,0 +1,74 @@
+# import manage
+# from tests import common
+# from webservices import rest
+
+# import pytest
+# import subprocess
+import logging
+
+
+# @pytest.fixture(scope="session")
```
lbeaufort, line 10: @rfultz can you remove the commented out imports and code, since it looks like they pertain to the API?

fec/data/templates/widgets/widgets-data.py
```diff
@@ -7,7 +7,7 @@
 # from data import utils
 
 # register = template.Library()
-def election_years = 2008
+# def election_years = 2008
```
lbeaufort, line 10: Can we remove this whole file?

fec/data/views_datatables.py
```diff
@@ -1,7 +1,7 @@
 from django.shortcuts import render, redirect
 from django.http import Http404
 
-import datetime
+# import datetime
```
lbeaufort, line 4: If we're not using an import, let's delete it

conftest.py
rfultz: Will do. I wasn't familiar enough with the code to know if if would break things

fec/data/templates/widgets/widgets-data.py
rfultz: What, because it's blank? 😆
It's outta here

fec/data/views_datatables.py
rfultz: 👍

fec/fec/wagtail_hooks.py
```diff
@@ -1,9 +1,9 @@
 import json
 import wagtail.admin.rich_text.editors.draftail.features as draftail_features
-from wagtail.admin.rich_text.converters.html_to_contentstate import InlineStyleElementHandler, BlockElementHandler
-from wagtail.admin.rich_text.converters.editor_html import WhitelistRule
+from wagtail.admin.rich_text.converters.html_to_contentstate import BlockElementHandler  # InlineStyleElementHandler
+# from wagtail.admin.rich_text.converters.editor_html import WhitelistRule
```
lbeaufort, line 4: Looks like two remaining commented out imports, `# from`...

rfultz: Updates incoming

PR #3828, opened 2020-06-08
Replaced financial document icon with new example icon

### LINKED ISSUE #3602: Update example icon globally
**What we're after:**
We've created a new icon to work for all example pages of varying types. The previous icon worked only for reporting examples. We need to update the icon used for examples, such as cards, globally. 

Icon is here: https://drive.google.com/drive/folders/18kchMACk2e8QWxSi0jW4N04KNxi8CIii

### Related issues
https://github.com/fecgov/fec-cms/issues/3588 New icon for disclaimer example

### Completion criteria
- [ ] Example icon is replaced with the new icon globally

rfultz:
## Summary

- Resolves #3602 
Removing the financial icon found on the [Advertising and disclaimers page](https://www.fec.gov/help-candidates-and-committees/advertising-and-disclaimers/) with the icon in the screenshot

Should only affect references to the img/financial-document-circle.svg, which was only in one place in the repo

## Screenshots

Icon in Sketch:
image

## Related PRs

None

## How to test

- TODO: duplicate the cms db to localhost to test
- Otherwise, check the Advertising and disclaimers page on dev (linked above)

____

## Next steps
Update the pattern library with this new (replacement) icon


rfultz:
Thanks, @johnnyporkchops—good catch! There's a new version on its way to git now


> I think we also need to replace this with an existing or new class for this template option on collection_page.html:
> https://github.com/fecgov/fec-cms/blob/7d956fccf45477889fa2f176d4e5aead51162c33/fec/home/templates/home/collection_page.html#L58
> 
> Here is an example of where it is used now:

johnnyporkchops:
@rfultz 
- We should keep the `icon-heading--financial-document` class and create another for example because we do use it some places. 
- And it maybe good idea to keep the SVG in `statsc/img` in case we need it and finding the svg or recreating it is kind if a  pain.
Apologies for not pointing that out in my review comment! Icons on our site can be confusing.

I think @JonellaCulmer would want us to  use `icon-heading--financial-document primary` for middle circle because its a financial thing (?) On home page:
Screen Shot 2020-06-09 at 8 36 58 PM

rfultz:
@JonellaCulmer, this is it in the wild:
image

PR #4035, opened 2020-09-10
Add empty string when null action

### LINKED ISSUE #4033: Remove dependence on commission_votes['action'] for MURs
### User story
As a user,  I would like to see at least partial data for MURs, even when all data are not available. Example: for issue https://github.com/fecgov/openFEC/issues/4592 the dev cms shows "Server error" because the action in commission_votes is null.

### Completion criteria
- [ ] partial data shows for MURS that have action=`null`

jason-upchurch:
## Summary (required)

- Resolves #4033 

MURS

- https://www.fec.gov/data/legal/matter-under-review/(mur case number)
 (for example https://www.fec.gov/data/legal/matter-under-review/7594)

### BEFORE:
Screen Shot 2020-09-10 at 2 25 06 PM

### AFTER:
Screen Shot 2020-09-10 at 2 20 07 PM

### COMPARE TO PROD (CASE WHEN DATA UPLOADED PROPERLY):
Screen Shot 2020-09-10 at 2 20 17 PM

## How to test:
- (@patphongs OR @lbeaufort )
- I added a test as well based on Pat's suggestion to add to logger when we come across a mur with no action. One way to run the test is with: `pytest -o log_cli=true data/tests/test_legal_search.py -s` (thanks Pat for having that in slack history)
- verify server error screen: https://dev.fec.gov/data/legal/matter-under-review/7594/
- check out branch
- run local cms and point to dev api (thanks Laura for explaining this vs local api)
- go to localhost:8000/data/legal/matter-under-review/7594/ and see that page loads (with disposition text not available)

### DIFF
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -151,13 +151,11 @@ def load_legal_mur(mur_no):
         for each index in mur["commission_votes"], if there is an empty action then log the mur and the
         particular index
         '''
-        i = 0
-        for d in mur["commission_votes"]:
+        for i, d in enumerate(mur["commission_votes"]):
             if not d["action"]:
-                logger.warning(
+                logger.error(
                     "MUR " + str(mur_no) + ": There were no data for commission_votes action at index " + str(i)
                 )
-            i += 1
         mur["collated_dispositions"] = collate_dispositions(mur["dispositions"])
         mur["complainants"] = complainants
         mur["participants_by_type"] = _get_sorted_participants_by_type(mur)
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -147,6 +147,17 @@ def load_legal_mur(mur_no):
                 complainants.append(participant["name"])
 
         mur["disposition_text"] = [d["action"] if d['action'] else '' for d in mur["commission_votes"]]
+        '''
+        for each index in mur["commission_votes"], if there is an empty action then log the mur and the
+        particular index
+        '''
+        i = 0
+        for d in mur["commission_votes"]:
+            if not d["action"]:
+                logger.warning(
+                    "MUR " + str(mur_no) + ": There were no data for commission_votes action at index " + str(i)
+                )
+            i += 1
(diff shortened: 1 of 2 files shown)


fec/data/api_caller.py
```diff
@@ -146,7 +146,9 @@ def load_legal_mur(mur_no):
             if "complainant" in participant["role"].lower():
                 complainants.append(participant["name"])
 
-        mur["disposition_text"] = [d["action"] for d in mur["commission_votes"]]
+        mur["disposition_text"] = [d["action"] if d['action'] else '' for d in mur["commission_votes"]]
```
patphongs, line 149: If commission_votes action was missing, can we have it also do a logger.error so that we can find the issue later? See line 168 below for an example

lbeaufort (approved):
Looks great, thanks for this fix, and for adding test coverage @jason-upchurch! Just one stylistic comment:

https://github.com/fecgov/fec-cms/blob/ff370bd1ac435a63b8ac3f58c10f51111d866d36/fec/data/api_caller.py#L154-L160

We can avoid iterating "manually" with `enumerate`"
```
        for i, d in enumerate(mur["commission_votes"]):
            if not d["action"]:
                logger.warning(
                    "MUR " + str(mur_no) + ": There were no data for commission_votes action at index " + str(i)
                )
```

fec/data/api_caller.py
```diff
@@ -147,6 +147,17 @@ def load_legal_mur(mur_no):
...
 
         mur["disposition_text"] = [d["action"] if d['action'] else '' for d in mur["commission_votes"]]
+        '''
+        for each index in mur["commission_votes"], if there is an empty action then log the mur and the
+        particular index
+        '''
+        i = 0
+        for d in mur["commission_votes"]:
+            if not d["action"]:
+                logger.warning(
```
lbeaufort, line 157: This should probably be `logger.error` if we want to match the logging level of line 176.

patphongs (approved):
Looks good, thanks @jason-upchurch

PR #4054, opened 2020-09-17
fixed sitemap PDF xml

patphongs:
## Summary

_Sitemap XML needed to be fixed and also the updated candidate guide lastmod date so that the search.gov indexer would reindex it again._

-  Guidance search


dorothyyeager (approved):
Thank you @patphongs for straightening out our code!

PR #4066, opened 2020-09-23
Pin greenlet@0.4.16 to fix build problem

jason-upchurch:
## Summary (required)

Pins greenlet@0.4.16 to fix build problem.


fec-jli (approved):
great job. Thanks

Issue #3160, opened 2019-09-11
Tag Manager implementation


rfultz:
Extending from [#3154](https://github.com/fecgov/fec-cms/pull/3154), a ticket to continue the Tag Manager implementation. This issue may or may not have a PR as the work _could_ happen entirely in the gtm interface and not touch the repo at all.

Progress:
- [x] Add feature flag (done in #3154)
- [x] Wrap current analytics inside the feature flag (done in #3154)
- [x] Create Tag Manager (GTM) container for non-Production (done in #3154)
- [x] Add Production GTM container to pages behind feature flag (done in #3154)
- [x] Add non-Production container to pages as an alternate to the Production container (done in #3154)
- [x] Add `<noscript>` GTM elements to templates (done in #3154)
- [x] Add code to non-Production container to confirm it's loading (done in #3154)
- [x] Confirm proper GA/DAP/GTM elements are loading (done in #3154)
- [x] Comb site to migrate any necessary GA events to GTM data layer
- [x] Confirm successful migration of GA events to GTM
- [x] Confirm GSA DAP is unaffected (all GSA DAP events are still coming through)
- [x] Remove current/then-retired GA implementation and conditionals (#4104 )

#3906 will serve as a kind of PR since this won't be in Git

Next steps in order:
- #3906 Test to make sure that GTM is successfully going into pages, adding GA accounts to pages
- #3972 Duplicate Google Tag Manager to prod 
- #3971 Approve Google Tag Manager on production
- #3973 Deploy Google Tag Manager with env vars
- #3948 Add new events into Google Tag Manager

jason-upchurch:
@rfultz I moved this to PI 13 innovation sprint--please let me know if I can help in any way. Thank you!

rfultz:
Final check is checked, so I'm closing it

PR #4075, opened 2020-09-25
Feature/3160 tag manager updates

rfultz:
## Summary

- Will resolve part of 3160; I named the branch poorly. Merging this shouldn't close 3160 so I intentionally didn't connect that ticket.

PR is to move the GSA/DAP so that it's always in the page (for Production), regardless of GTM environment variable.

Also reformatted middleware.py so it's a little more human-friendly

Technically every page on the site on every environment will be affected but will only affect whether DAP goes into the page with GTM or without. Reformatting the middleware file could change content permissions but I think they're fine.

## Screenshots

Nothing to show

## Related PRs

None

## How to test

Will need to check the source for Feature, Dev, and Stage to make sure DAP isn't going into the page regardless of GTM. Will need to check Production to make sure DAP loads regardless of GTM environment variable

Would like to mention #3160 to make a non-resolving connection between that ticket and this PR

____

### DIFF
--- a/fec/data/templates/layouts/widgets.jinja
+++ b/fec/data/templates/layouts/widgets.jinja
@@ -93,6 +93,9 @@
     ga('create', 'UA-16134356-1', 'auto');
     ga('send', 'pageview');
   </script>
+{% endif %}
+
+{% if (FEC_CMS_ENVIRONMENT == 'PRODUCTION') %}
   <script id="_fed_an_ua_tag" src="https://dap.digitalgov.gov/Universal-Federated-Analytics-Min.js?agency=FEC"></script>
 {% endif %}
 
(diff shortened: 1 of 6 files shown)


fec/data/templates/layouts/widgets.jinja
```diff
@@ -93,6 +93,9 @@
     ga('create', 'UA-16134356-1', 'auto');
     ga('send', 'pageview');
   </script>
+{% endif %}
+
+{% if (FEC_CMS_ENVIRONMENT == 'PRODUCTION') %}
```
patphongs, line 98: Thanks for making DAP independent of tag manager

patphongs (approved):
Looks great!

johnnyporkchops (approved):

PR #4086, opened 2020-09-29
Consolidate GTM code, update to use environments

### LINKED ISSUE #4083: Change GTM to use separate environments rather than separate containers
To better use Tag Manager, let's update our GTM code to use different environments under one container rather than two containers.

Environments will let us publish different versions to different environments.

We've been using two containers (one for production, one for everything else). The big downside is that exporting from one container (non-prod) is easy enough, but importing to a container (production) could mean replacing the importing container.

rfultz:
## Summary

- Resolves #4083 

- Consolidated the GTM code to a few external files rather than update code site-wide.
- Updated GTM code to use the various environments of our main GTM container rather than rely on separate containers.

Source code of every page

## Screenshots

No visible changes

## Related PRs

None

## Reviewers

- Having other eyes is always good, but feel free to add or remove yourself

## How to test

- pull and build branch like normal
- `./manage.py runserver`
- Check each page's source, being sure to include a the homepage, a Wagtail page, a data page, a legal page, and a widget
  - Check with GTM turned on
  - and with GTM turned off
- If `echo $FEC_FEATURE_USE_TAG_MANAGER` tells you `true`, you should see entries for GTM in every page's source code; if not, no page on the site should have 'gtm' in the source code.
  - To turn on GTM, use `export FEC_FEATURE_USE_TAG_MANAGER=true`
  - and `export FEC_FEATURE_USE_TAG_MANAGER=false` will turn it off

____

### DIFF
--- a/fec/data/templates/layouts/main.jinja
+++ b/fec/data/templates/layouts/main.jinja
@@ -23,16 +23,7 @@
     </script>
   {% endblock %}
 
-  {# Google Tag Manager #}
-  {% if FEATURES.use_tag_manager %}
-    {% if FEC_CMS_ENVIRONMENT == 'PRODUCTION' %}
-      {# Google Tag Manager for Production #}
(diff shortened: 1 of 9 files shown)


patphongs (approved):
Changes look good @rfultz, deployed to the dev environment and it picked up the correct container

rfultz:
@patphongs should we go ahead and merge? @johnnyporkchops ?

johnnyporkchops (approved):
Tested locally with  `$FEC_FEATURE_USE_TAG_MANAGER` set and unset and works! Good job, this will be very useful.

PR #4112, opened 2020-10-09
[WIP] Fix election profile page

lbeaufort:
## Summary (required)

- Resolves #[_Issue number_]

-  

## Screenshots

- _Include a screenshot of the new/updated features in context (“in the wild”). If it is an interface or front-end change, include both Before and After screenshots._

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

## How to test

Include any information that may be helpful to the reviewer(s).
_This might include:_

- links to sample pages to test
- Any local environmental setup that is unusual (env vars, API version to point to, etc)

____


jason-upchurch:
@rfultz @johnnyporkchops @patphongs sorry for the short fuse on this, but the issue is targeted for release--can you let me know if I can help (perhaps review as stand in for one of you?)

@lbeaufort please  let me know if you need all 3 reviewers or if I can help. Thank you all!

fec/fec/static/js/modules/helpers.js
```diff
@@ -328,8 +328,8 @@ function buildUrl(path, query) {
   return uri.addQuery(query).toString();
 }
 
-function buildTableQuery(context) {
-  var pageLength = pageLength || 0;
+function buildTableQuery(context, params) {
+  var pageLength = params.per_page || 100;
```
rfultz, line 332: I was thinking about this. We didn't see that buildTableQuery was being used anywhere else but, just in case, could we change this line back to how it was, then add another line after it so the two look like:
`var pageLength = pageLength || 0;`
`if (params.per_page) pageLength = params.per_page;`

That will let us still default to whatever global pageLength value might exist, with 100 as its fallback, but also let us override that if `params.per_page` is provided

fec/fec/static/js/pages/elections.js
```diff
@@ -44,7 +43,7 @@ $(document).ready(function() {
       order: [[3, 'desc']]
     }
   };
-
+  var query = helpers.buildTableQuery(context.election, { per_page: 20 });
```
rfultz, line 46: Just noting for others that we moved this line down to be closer to where it's referenced

rfultz (changes requested):

fec/fec/static/js/modules/helpers.js
lbeaufort: @rfultz thanks for your help on this, this commit https://github.com/fecgov/fec-cms/pull/4112/commits/2c3414218f618f1da7d308d8978893610edc13bd we worked on together should address your comment

rfultz (approved):
Yay!

patphongs (approved):
Nice work and thanks for adding tests for this @lbeaufort!

Issue #4043, opened 2020-09-16
[Snyk: Medium] package.json - Denial of Service  (due 11/16/20)


hcaofec:
**Denial of Service:**

Vulnerable module:	node-fetch
Introduced through: draft-js@0.10.5
Exploit maturity:  No known exploit
Fixed in:	2.6.1, 3.0.0-beta.9

**Detailed paths**
Introduced through: fec-cms@1.0.0 › draft-js@0.10.5 › fbjs@0.8.17 › isomorphic-fetch@2.2.1 › node-fetch@1.7.3
Remediation: Upgrade node-fetch to version 2.6.1, 3.0.0-beta.9 or higher.

**Overview**
node-fetch is an A light-weight module that brings window.fetch to node.js

Affected versions of this package are vulnerable to Denial of Service. Node Fetch did not honor the size option after following a redirect, which means that when a content size was over the limit, a FetchError would never get thrown and the process would end without failure.

https://app.snyk.io/vuln/SNYK-JS-NODEFETCH-674311

**Action items**
- [ ] identify priority/categorizing and document. Notify _concerned_ parties.

**Completion criteria:**
- [ ] Verify that this is a problem and address the vulnerability appropriately
- [ ] see if this fix also solves the vulnerability from https://github.com/fecgov/fec-cms/pull/4087 by removing ua-parser-js from package.json. If so, we can keep it removed from package.json.

jason-upchurch:
There is a live discussion in `draftail` repo about `draft-js` dependency. (Remediating through the current method: `Upgrade draft-js@0.10.5 to draft-js@0.11.7` results in unmet dependencies.)
https://github.com/springload/draftail/issues/213#issuecomment-717851461

rfultz:
We've moved this to Blocked because there's no remediation for the affected package's dependency grandparent (if that's a thing. The package that requires the package that requires the affected package) and that package hasn't seen any development in over a year.

The vulnerability is only Medium and isn't publicly exposed for us until a user logs in to the admin section.

jason-upchurch:
@rfultz @patphongs I moved this to blocked 13.8. Please let me know if it should be moved elsewhere.

lbeaufort:
Because the vulnerability can't be exploited except for authenticated user, this is not a meaningful vulnerability. This is also no longer being flagged by Snyk.

PR #4137, opened 2020-10-20
4043 Snyk Med, node-fetch

rfultz:
## Questions
- Do we still want/need to do this upgrade if it doesn't address the vulnerability?
- Should we change #4043 to can't fix?

## Summary

- Doesn't really resolve # 4043 so I'm not linking it. Yet.

This ticket upgrades draftail to the most recent version, but even it requires the draft-js version in this vulnerability.

Draftail powers our page admin interface so is inside the locked admin area.

**Complications**:
- This ticket is about node-fetch
- node-fetch is being used by a dependency of a dependency of draft-js, which is only being used by Draftail.
- Draftail requires draft-js 0.10.5 and gives me a warning when I use a newer version (.5 is the last of 0.10 and 0.11.0 errs)
- For us, this ticket requires an upgrade of Draftail, which doesn't exist

**More complications**: 
We may need to look into a new editor as there hasn't been a Draftail update since August 2019. There's an [issue](https://github.com/springload/draftail/issues/406) from August 2020 asking whether the Draftail project is dead with no replies

dependency tree: draft-js@0.10.5 › fbjs@0.8.17 › isomorphic-fetch@2.2.1 › node-fetch@1.7.3

Our site admin editor

## Screenshots

Should be no visible changes

## Related PRs

None

## How to test

- Pull the branch
- `npm i`
- `npm run build-production`
- `./manage.py runserver`
- Go into the admin area of localhost:8000 or http://127.0.0.1:8000 and edit a couple pages
- Should be no difference from current functionality
- Unfortunately, `snyk test` still lists the vulnerability is #4043

____


rfultz:
Closing this because it doesn't really solve 4043. About to move #4043 to Blocked, too

PR #4219, opened 2020-12-01
Feature/3313 smallest menu height

### LINKED ISSUE #3313: Allow users to see all mobile menu items
**What we're after:**
Users should be able to view the entire mobile menu. When viewing the About or Legal resources menus, you're unable to see all menu options in a given accordion. 

This has been observed on both Chrome and Safari on iPhone X. 

Screenshot 2019-10-31 at 4 17 59 PM

### Completion criteria

- [ ] Ensure users can view all mobile menu items

rfultz:
## Summary

- Resolves #3313 

Adjusting 

-  

## Screenshots

Default open state
image

With Campaign finance expanded but with About scrolled into view
image

## Related PRs

None

## How to test

- pull the branch like normal
- `npm i`
- `npm run build`
- `./manage.py runserver`
- On [any page](http://127.0.0.1:8000), shrink the browser window width to as small as possible
- Make sure the menu still opens and the bottom of About is always accessible
- If the "LOCAL" banner is at the top, remove it and test that version
  - right click that top blue banner and choose Inspect or whatever your browser calls it
  - in the developer tools that open, a child element of `<div class="banner-dev">` should be selected
  - click the `banner-dev` element
  - delete it (delete, backspace, whatever)
  - close the developer tools if you'd like
  - check the menu positions again (open different menu options, make sure the bottom of About is still accessible)

____

### DIFF
--- a/fec/fec/static/scss/components/_nav.scss
+++ b/fec/fec/static/scss/components/_nav.scss
@@ -18,14 +18,13 @@
 // div that contains the ul and all panels
 .site-nav__container {
   $top: u(6.3rem);
-  $adjustment: 40px;
   @include transform(translateX(-100%));
   @include transition(left, .2s, ease-out);
-  height: calc(100vh - #{$top} - #{$adjustment});
+  height: calc(100vh - #{$top});
   position: absolute;
   left: 0;
   right: 0;
-  top: $top;
+  top: 4rem;
   overflow-x: hidden;
   z-index: $z-navigation;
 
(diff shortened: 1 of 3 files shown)


JonellaCulmer (approved):
Approving this because I believe it will work just fine on production. However, when the local banner is up it pushes the other content down and we have the same problem. I suspect the same will happen on our other environments including dev, stage and feature.

fec/fec/static/scss/components/_nav.scss
```diff
@@ -34,6 +34,13 @@
   }
 }
 
+// If we have the dev banner, we'll need to adjust #site-menu
+.banner-dev + .usa-banner + .site-header nav #site-menu {
+  $devBannerTop: 4rem;
```
johnnyporkchops, line 39: @rfultz , I am getting this error, for some reason the SCSS var is not working. But I do have all the new files pulled.
Screen Shot 2020-12-09 at 12 31 01 PM

rfultz: (There's a new version on its way)

PR #4256, opened 2020-12-15
feature/fix-typo

pkfec:
Replace `dev` with `stage` from the comments as FEC_API_URL is pointing to stage api.


patphongs (approved):
Thanks for fixing that @pkfec

PR #4314, opened 2021-01-08
Jinja autoescape

rfultz:
## Summary
Resolves https://github.com/fecgov/fec-accounts/issues/340

Setting Jinja's autoescape to True and changing two instances to use the `safe` filter instead of bypassing autoescape

autoescape was already True but we want to explicitly set it. The meeting page and jobs templates each had one place where we were turning off autoescape; we've changed those to use the `safe` filter instead

Could affect any part of the site where we're processing content with Jinja—every Jinja template or component, including *.jinja and *.html

## Screenshots

None

## Related PRs

None

## How to test

- pull the branch like normal
-`npm i`, `npm run build` just in case? Should have zero effect
- `pip install -r requirements.txt` just in case? Likely have zero effect
- `./manage.py runserver`
- Make sure the blog/Wagtail parts of the site behave correctly, specifically checking pages with content that comes from Wagtail (about, legal, etc). I imagine any errors would look like html characters being visible in the browser front-end, like `&lt;` instead of `<` or `&amp;` instead of `&`.

Don't merge this yet—could have a couple more changes

____

### DIFF
--- a/fec/data/jinja2.py
+++ b/fec/data/jinja2.py
@@ -5,5 +5,6 @@ from compressor.contrib.jinja2ext import CompressorExtension
 def environment(**options):
     """Create a jinja2 environment with the CompressorExtension added in"""
     options['extensions'] += [CompressorExtension]
+    options['autoescape'] = True  # This was already True but we want to set it explicitly
     env = jinja2.Environment(**options)
     return env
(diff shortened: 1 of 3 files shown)


rfultz:
@johnnyporkchops and I just experimented with these and they're working as expected

johnnyporkchops (approved):
Looks good top me!

lbeaufort (approved):
Thanks for your work on this, @rfultz! FYI, bandit still flags this issue because it's being set in the `options` object, so it's now a false positive. I appreciate the helpful comment!

PR #4500, opened 2021-03-19
Replace sanitize-html

### LINKED ISSUE #4026: [Snyk: High] Arbitrary Code Execution in sanitize-html (due 10/9/20)
Vulnerable module: `sanitize-html`

Introduced through: sanitize-html@1.18.4
No known exploit
Fixed in: 2.0.0-beta

**Detailed paths**
Introduced through: fec-cms@1.0.0 › sanitize-html@1.18.4
Remediation: No remediation path available.

**Overview**
sanitize-html is a library that allows you to clean up user-submitted HTML, preserving whitelisted elements and whitelisted attributes on a per-element basis

Affected versions of this package are vulnerable to Arbitrary Code Execution. Tag transformations which turn an attribute value into a text node using transformTags could be vulnerable to code execution.

**Completion criteria:**
- [ ] Verify that this is an issue and if it is take appropriate to address the vulnerability

lbeaufort:
## Summary

- Resolves #4026 

To silence recurring errors for sanitize-html, how about we switch to dompurify instead?

### QUESTIONS
The tests for this kept failing for something I'm not sure is the best expectation:

While escaping `"` characters, for some reason the text wants (straight double) quotes to be converted to `&quot;`, (okay, fine, not unexpected) but then the regex removes the `&` and `;` so quotes get turned into `quot` rather than a legit escaped value. Seems like they should be stripped out or turned into `&quot;`, `&#34;`, `0x22`, `%22`…—is there a reason we're going through the trouble of escaping it only to rip it to pieces, too?

I've added a patch for this to pass tests and behave as currently expected but I don't like it.

Are there other characters where this is or should be happening but which aren't in tests?

If we change this, how will it affect site search and filters? I can see removing straight quotes from filters but they help with grouping terms for site search. There are special characters in organizations' names, too: lots of companies with `&` in their name.

Anything that uses helpers.sanitizeValue(), which is every typeahead and some tests. Typeaheads including filters, site search, candidate lookups in widgets… _but_ if it's fixed in search + filters, it's likely good everywhere.

I changed the `_.isArray()` calls to the modern `Array.isArray()` since it was so easy to take another step from Underscore

## Screenshots

No visual changes

## Related PRs

None

## How to test

- pull branch
- `npm i`
- `npm run test-single` (it also does its own build)
- through unpredictable searches into various typeaheads, especially for search and for filters that create chiclets. For examples of unusual searches, see lines 88-151 of fec/fec/tests/js/helpers.js

Because we've removed sanitize-html, that vulnerability warning should vanish

____


patphongs (approved):
Looks great, thanks!

johnnyporkchops (commented):
looks good, great job getting domPurify to work!

PR #4564, opened 2021-04-23
Feature/4542 simple document list for custompage

### LINKED ISSUE #4542: Add document list block to custom page template
**What we're after:**
As part of our work to move the matching fund submission pages off of transition, we need to make adjustments to the custom page template to accommodate the information and display it the best way we can. To do this we need to add the document list block to the page template.

**Depending on the desired presentation and styling We could possible use one of the following:**
 - The existing FeedDocumentBlock:
 https://github.com/fecgov/fec-cms/blob/6ebd8317056c960fc61761e7f53ac0387f595112/fec/home/blocks.py#L96

 - A Wagtail DocumentChooserBlock with structure and styling set by the template and CSS.

**Completion criteria:**
- [ ] Document list block added to the CustomPage template

**Mockup:**
mockup_documentlist_custompage

johnnyporkchops:
## Summary (required)

- Resolves #4542

- Add ability for editors to create a list of documents on the CustomPage with  document icon, document title,  and file-type(PDF, EXCEL etc.)
 -  Using the existing FeedDocumentBlock but calling a new template: `simple-document-list.html`
 -  Add new `get_file_type` filter for templates 
 -  Add the new get_file_type filter to the existing `document-list` on ResoucePage based on follow-on comment in issue:
 https://github.com/fecgov/fec-cms/issues/4542#issuecomment-825161821

### Required reviewers

 Reviewers: One designer and one front end developer. Content team might have input on its reusability on other templates. 

   CustomPage templates body-streamfield will have new `Simple Document List` item
   Resource page sections-streamfield `Document List` item now auto detects file-type

	new file:   home/migrations/0114_auto_20210430_1346.py
	new file:   home/templates/blocks/simple-document-list.html
	modified:   home/models.py
	modified:   home/templatetags/filters.py
	modified:   home/templates/blocks/document-list.html

### New "Simple Document List"  block in CustomPage
Screen Shot 2021-04-22 at 9 20 31 PM

<hr>

### Using both "document list" and "simple document list"  on ResourcePage

Screen Shot 2021-04-30 at 1 30 36 PM

<hr>

### Using new "get_file_type"  filter instead of  hardcoded file-type for "Document List" on ResourcePage

Screen Shot 2021-04-22 at 8 32 34 PM

<hr>

## How to test
**On feature space**
https://fec-feature-cms.app.cloud.gov/
- Example page: 
     - View: https://fec-feature-cms.app.cloud.gov/test-simple-doc-list/
     - Edit: https://fec-feature-cms.app.cloud.gov/admin/pages/11633/edit/
- Also test `Simple document list` on a new `CustomPage`
- Also confirm that the existing `document list` on a `ResourcePage` now shows correct file type:
https://fec-feature-cms.app.cloud.gov/introduction-campaign-finance/understanding-ways-support-federal-candidates/presidential-elections/public-funding-presidential-elections/test-of-document-block-for-zip-files/
- Both `document list` and `simple document list` have been added to both `ResourcePage` and `CustomPage` if you want to test any of that functionality

**Developers:**
- Checkout `feature/4542-simple-document-list-for-custompage`
- .`/manage.py migrate`
-  Create a new CustomPage and try out the `Simple document list` option in the body section.
- Here is the mockup of desired look: https://user-images.githubusercontent.com/31663028/114922792-1aaf1a00-9dfa-11eb-93c2-929906a4b7f1.png
- Both `document list` and `simple document list` have been added to both `ResourcePage` and `CustomPage` if you want to test any of that functionality


fec/home/templatetags/filters.py
```diff
@@ -167,3 +167,12 @@ def get_social_image_path(identifier):
     else:
         imageFilename = 'fec-logo'
     return 'https://www.fec.gov/static/img/social/{}.png'.format(imageFilename)
+
+
+@register.filter(name='get_file_type')
+def get_file_type(value):
+    file_extension = value.rsplit('.', 1)[1].upper()
```
rfultz, line 174: Out of curiosity, is it possible to upload a file without an extension?

Is there a built-in function for this kind of thing, in Wagtail or Django? Surely there is. Some kind of extension/mime-type check? If there isn't and we're going to build one, should it be expanded to include other possibilities? Word files, PowerPoint, csv, pdf, txt, rtf, …?

rfultz:
From the comp, how would one get a section head like "Nixon for president"? It's not a Heading object:
image

rfultz (commented):
Left a couple comments but nothing breaking

JonellaCulmer:
@johnnyporkchops Here's the screenshot I took during our conversation regarding bottom padding. 
Screen Shot 2021-04-30 at 11 26 14 AM

Let me know if you'd like me to test it again once "document list" is added to the custom template and "simple document list" is added to the resource template. 

Thanks for your work on this!

fec/home/templatetags/filters.py
johnnyporkchops: So it works for PDF, CSV, anything that we say the file type, the same as the extension (uppercased) xls and xlsx eas the only one I could think of that we call it EXCEL instead of xls.

In terms of an existing extension, I did search and the suggestions were to roll your own. 

I also considered your question about a non-extension file, and I would think we do not want people to be able to upload such a file. Not that this prohibits that,. We can test it though in Wagtail.

johnnyporkchops:
@rfultz It looks like Wagtail recognizes the file type on a no-extension file and appends it when uploaded
0001_1250100861_medium

johnnyporkchops:
> From the comp, how would one get a section head like "Nixon for president"? It's not a Heading object:
> image

@rfultz You can use a paragraph or heading block above. This way its optional

johnnyporkchops:
After discussion with  @JonellaCulmer , @dorothyyeager (design/content), We decided to make both these document list items available to both ResourcePage and CustomPage to avoid having to go back later and add it upon an arising use-case for content team.

JonellaCulmer (approved):
Everything works great, @johnnyporkchops. 

One thing I tried was to do the document list and then a simple document list right below it on the resource template. The simple document list (when made within the same section) bumps right up against the bottom dividing line of the last document. Didn't look great, but I can't picture a use case for that now. Just a thing to document if in the future we "discover" that is an issue. 

Either way, approved. Thanks again for your hard work!

johnnyporkchops:
> Everything works great, @johnnyporkchops.
> 
> One thing I tried was to do the document list and then a simple document list right below it on the resource template. The simple document list (when made within the same section) bumps right up against the bottom dividing line of the last document. Didn't look great, but I can't picture a use case for that now. Just a thing to document if in the future we "discover" that is an issue.
> 
> Either way, approved. Thanks again for your hard work!

@JonellaCulmer I noticed the same thing and was going to warn about it but as you said, I could not think of a use case. Also ,  when I went back and added a title for the second the,  list it did not bump up against it anymore.  In a bind one could add an html block with a <br> to solve this--not ideal but would work in rare use -cases. As in this screenshot:
Screen Shot 2021-04-30 at 3 25 08 PM

fec/home/templatetags/filters.py
rfultz: Are we being consistent with using "EXCEL" rather than "xls" or "xlsx" and "WORD" for "doc" and "docx" and "POWERPOINT" for "ppt" and "pptx"?

It does seem weird that we're using "PDF" (a file extension) and "EXCEL" (a capped name of one program that can open those files).

Trying to think of a different analogy… is this like comparing "vegetable" and "Tropicana orange juice"—one generic category vs one very specific term?

Could be a larger question and certainly no reason to hold up this merge—we could easily come back to it and likely a UX question for @JonellaCulmer

rfultz (approved):

PR #4604, opened 2021-05-06
Feature/4172 eliminate render blocking resources

### LINKED ISSUE #4172: Optimization: Eliminate render-blocking resources
### Background
Looking to [optimize the site's performance](https://github.com/fecgov/fec-cms/issues/3711) and that of the homepage specifically, the Lighthouse tool in Chromium browsers has made some suggestions. Some of these will have repercussions throughout the app so we're going to make them their own tickets.

# Recommendation: Eliminate render-blocking resources
Resources are blocking the first paint of your page. Consider delivering critical JS/CSS inline and deferring all non-critical JS/styles. [details](https://web.dev/render-blocking-resources/?utm_source=lighthouse&utm_medium=devtools)

## Overview
We could embed the content of our CSS and JS files into the page rather than linking to them

## Benefit (the 80 of 80/20)
Minimal

## Effort (the 20 of 80/20)
Likely minimal

## Complications
- Possible security complications? Thinking about the sandbox between in-page `<script>` and `<script src="…">`. Check into CORS rules
- Would want to make sure we aren't affecting templates unexpectedly
- Could affect script load timing and race conditions, like with a `<script async defer>` in the head (which would load last) vs putting all that code into the page right then

#### Related tickets
- [Preload key requests](https://github.com/fecgov/fec-cms/issues/4168)
- [Remove unused JavaScript](https://github.com/fecgov/fec-cms/issues/4169)
- [Remove unused CSS](https://github.com/fecgov/fec-cms/issues/4170)
- [Preconnect to required origins](https://github.com/fecgov/fec-cms/issues/4171)
- [Ensure text remains visible during webfont load](https://github.com/fecgov/fec-cms/issues/4173)
- [Reduce the impact of third-party code](https://github.com/fecgov/fec-cms/issues/4175)
- [Does not use passive listeners to improve scrolling performance](https://github.com/fecgov/fec-cms/issues/4176)
- [Image elements do not have explicit width and height](https://github.com/fecgov/fec-cms/issues/4177)
- [Serve static assets with an efficient cache policy](https://github.com/fecgov/fec-cms/issues/4178)
- [Minimize main-thread work](https://github.com/fecgov/fec-cms/issues/4179)
- [Avoid an excessive DOM size](https://github.com/fecgov/fec-cms/issues/4180)
- [Reduce JavaScript execution time](https://github.com/fecgov/fec-cms/issues/4181)
- [Avoid enormous network payloads](https://github.com/fecgov/fec-cms/issues/4182)
- [[aria-hidden="true"] elements contain focusable descendents](https://github.com/fecgov/fec-cms/issues/4183)
- [Elements with an ARIA [role] that require children to contain a specific [role] are missing some or all of those required children](https://github.com/fecgov/fec-cms/issues/4184)
- [[aria-*] attributes do not have valid values](https://github.com/fecgov/fec-cms/issues/4185)
- [ARIA IDs are not unique](https://github.com/fecgov/fec-cms/issues/4186)
- [Lists do not contain only <li> elements and script supporting elements (<script> and <template>)](https://github.com/fecgov/fec-cms/issues/4187)
- [[user-scalable="no"] is used in the <meta name="viewport"> element or the [maximum-scale] attribute is less than 5](https://github.com/fecgov/fec-cms/issues/4188)
- [Links to cross-origin destinations are unsafe](https://github.com/fecgov/fec-cms/issues/4189)
- [Serves images with low resolution](https://github.com/fecgov/fec-cms/issues/4190)

rfultz:
## Summary

- Resolves #4172 
- deferring script loads to improve our page performance
- preloading the homepage hero image

### Required reviewers

Two front-end and/or UX?

Preloading the homepage hero has no downside.

Now we're delaying script loads for a few templates. The most likely complication is that one of the scripts won't have loaded when it was expected, meaning a component isn't activated or isn't ready to when a user thinks it should have been. This could happen the first time a site visitor requests a script after our last deploy (pages that share scripts will use their cached versions) and when there's a slow connection.

The affected templates are for the homepage, elections, election search, and landing.

The pages 

## Screenshots

No visual changes

## Related PRs

None

## How to test

- pull the branch
- `./manage.py runserver` 
- Make sure the interactive parts work: (menus, search, widgets, glossary, etc)
   - [homepage](http://127.0.0.1:8000)
   - election-lookup (homepage and /data/elections/)
   - [elections](http://127.0.0.1:8000/data/elections/)
   - [landing](http://127.0.0.1:8000/data/)
   - Slow the connection speed and try again
- In Inspector, under the Network tab, change "No throttling" to something else and re-check the pages. They'll load more slowly and may not become useable until the last file is loaded and processed. But they should become interactive just like usual (just a little later than we're expecting).

For the homepage, Lighthouse should no longer mention "eliminate render blocking resources"

### DIFF
--- a/fec/home/templates/home/home_page.html
+++ b/fec/home/templates/home/home_page.html
@@ -109,12 +109,11 @@
     document.head.appendChild(pfScriptElem);
   }
 </script>
-<script src="{% asset_for_js 'home.js' %}"></script>
-<script src="{% asset_for_js 'bythenumbers.js' %}"></script>
-<script src="{% asset_for_js 'dataviz-common.js' %}"></script>
-<script src="{% asset_for_js 'election-lookup.js' %}"></script>
+<script defer src="{% asset_for_js 'home.js' %}"></script>
+<script defer src="{% asset_for_js 'bythenumbers.js' %}"></script>
+<!-- <script defer src="{% asset_for_js 'dataviz-common.js' %}"></script> -->
+<script defer src="{% asset_for_js 'election-lookup.js' %}"></script>
 <script>
-  //remove competing/confusing querystrings on homepage
   $(function(){
     cleanURI()
     function cleanURI(){
--- a/fec/data/templates/election-lookup.jinja
+++ b/fec/data/templates/election-lookup.jinja
@@ -112,6 +112,6 @@ var context = {
   }
 };
 </script>
-<script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-<script src="{{ asset_for_js('election-lookup.js') }}"></script>
+<script defer src="{{ asset_for_js('dataviz-common.js') }}"></script>
+<script defer src="{{ asset_for_js('election-lookup.js') }}"></script>
 {% endblock %}
--- a/fec/data/templates/elections.jinja
+++ b/fec/data/templates/elections.jinja
@@ -62,7 +62,7 @@ var context = {
 };
 
 </script>
-<script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-<script src="{{ asset_for_js('elections.js') }}"></script>
+<script defer src="{{ asset_for_js('dataviz-common.js') }}"></script>
+<script defer src="{{ asset_for_js('elections.js') }}"></script>
 {% endblock %}
 
--- a/fec/data/templates/landing.jinja
+++ b/fec/data/templates/landing.jinja
@@ -301,6 +301,6 @@
 {% endblock %}
 
 {% block scripts %}
-  <script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-  <script src="{{ asset_for_js('data-landing.js') }}"></script>
+  <script defer src="{{ asset_for_js('dataviz-common.js') }}"></script>
+  <script defer src="{{ asset_for_js('data-landing.js') }}"></script>
 {% endblock %}
--- a/fec/fec/templates/home_base.html
+++ b/fec/fec/templates/home_base.html
@@ -4,6 +4,7 @@
 <html lang="en">
   <head>
     {% include 'partials/meta-tags.html' %}
+    <link rel="preload" as="image" href="/static/img/Homepage-hero-image-flag.jpg">
     {% include 'partials/meta-tags-preloads.html' %}
     {% if settings.FEC_CMS_ENVIRONMENT == 'PRODUCTION' %}
     {% include 'partials/meta-tags-preconnects.html' %}


patphongs (commented):
@rfultz Looks like lighthouse is still flagging the elections.css in the "Eliminate render blocking" report. Is there a way to defer that?

`<link rel="stylesheet" type="text/css" href="{% asset_for_css 'elections.css' %}">`

Screen Shot 2021-05-17 at 11 22 56 AM

patphongs:
> @rfultz Looks like lighthouse is still flagging the elections.css in the "Eliminate render blocking" report. Is there a way to defer that?
> 
> `<link rel="stylesheet" type="text/css" href="{% asset_for_css 'elections.css' %}">`

Per @rfultz, this might be addressed in the clean up work in [this PR](https://github.com/fecgov/fec-cms/pull/4606).

patphongs (approved):
Looks good!

fec/home/templates/home/home_page.html
```diff
@@ -109,12 +109,11 @@ <h1>Commissioners</h1>
     document.head.appendChild(pfScriptElem);
   }
 </script>
-<script src="{% asset_for_js 'home.js' %}"></script>
-<script src="{% asset_for_js 'bythenumbers.js' %}"></script>
-<script src="{% asset_for_js 'dataviz-common.js' %}"></script>
-<script src="{% asset_for_js 'election-lookup.js' %}"></script>
+<script defer src="{% asset_for_js 'home.js' %}"></script>
+<script defer src="{% asset_for_js 'bythenumbers.js' %}"></script>
+<!-- <script defer src="{% asset_for_js 'dataviz-common.js' %}"></script> -->
```
johnnyporkchops, line 114: Do you want to leave this commented tag in here?

fec/home/templates/home/home_page.html
```diff
@@ -109,12 +109,11 @@ <h1>Commissioners</h1>
...
+<script defer src="{% asset_for_js 'election-lookup.js' %}"></script>
```
johnnyporkchops, line 115: The `cleanURI()` function that we use on the homepage to keep the querystring out of the URL does not work if we defer `election-lookup.js`

It needs jQuery. Specifically, the .`on` method. I tried to start writing a vanilla JS version, but it gets tricky trying to emulate `$.on`. Should we replace it with vanilla JS or is there a better way around this?

johnnyporkchops (changes requested):
The `cleanURI()` function that we use on the homepage to keep the querystring out of the URL does not work if we defer `election-lookup.js`

It needs jQuery. Specifically, the .`on` method. I tried to start writing a vanilla JS version, but it gets tricky trying to emulate `$.on`. Should we replace it with vanilla JS or is there a better way around this?

fec/home/templates/home/home_page.html
rfultz: What if we change election-lookup to async?

fec/home/templates/home/home_page.html
rfultz: Oof. Nope! Good catch!

fec/home/templates/home/home_page.html
rfultz: I changed this to native on another ticket. Should I merge that one into this one?

johnnyporkchops: I did try async and got same problem

johnnyporkchops: > I changed this to native on another ticket. Should I merge that one into this one?

Whatever fixes it

johnnyporkchops:
Thanks @rfultz .
Now it is removing  the querystring on-load, but then if you interact with map, it keeps only state or only the date. 

If we  ended up deciding that its OK to have querystrinngs once someone interacts with the map, we would probably want to have it register every parameter they choose

johnnyporkchops:
@rfultz this works if we put it back on `home_page.html` inside a `window.onload` function.
```
<script>
window.onload = function () {
  $(function() {
    cleanURI()
      function cleanURI(){
        const uri = window.location.toString();
        if (uri.indexOf('?') > 0) {
          window.history.replaceState({}, document.title, location.href.split('?')[0]);
        }
      }
      $('#main').on('change ', 'input, select', function(){
        cleanURI();
      })
      //handle Chrome insconsistency with History API
      $('.js-office').val('S');
      $('.js-election-year').val('2022')
      $('.js-chart-toggle').filter('[value=receipts]').prop('checked', true)
  })
}
</script>
```
- it  works intermittently inside `home.js`, but loses the race condition intermittently, which I have not seen happening when putting it directly in `home_page.html`. Let me know what you find if you test it. 
- On some browsers I could remove jQuery's shorthand `$(function()` wrapper( which is the same as `$(document).ready(function()` ) and just wrap it in `window.onload`but then Opera does not work. So even though it might feel redundant to have two onloads, I think that is the most ubiquitious coverage. 

Let me know what you think

fec/home/templates/home/home_page.html
```diff
@@ -114,60 +114,23 @@ <h1>Commissioners</h1>
...
-  const uri = window.location.href.toString();
-  if (uri.indexOf('?') > 0) {
-    window.history.replaceState(
-      {},
-      document.title,
-      window.location.href.split('?')[0]
-    );
-  }
+window.onload = function () {
+  $(function() {
```
rfultz, line 118: So we'd be using the the native `onload` event to trigger a jQuery function (`$(function)`) that creates another function (`cleanURI`), calls it, and calls four other jQuery commands (`on`, `val`, `val`, and `filter.prop`)?

johnnyporkchops: Yes. Not ideal but since it was the only working solution yet, it's worth logging it to the conversation should we need to fall back on it.  And only requires adding two lines of code to the existing. Do the latest changes committed since work?  Were you able to confirm that before pushing.?

patphongs:
We're getting some errors here, so we can't merge this yet:

1. Webpack error attributed to the defer on the `vendor.js`.
```
calc-admin-fines-modal-02db7d5fde06fe303948.js:1 Uncaught ReferenceError: webpackJsonp is not defined
    at calc-admin-fines-modal-02db7d5fde06fe303948.js:1
```

2. Updating the district dropdown does not strip the URL. All the other dropdowns do strip the URL as expected.

patphongs (changes requested):
Some things need to be addressed first: https://github.com/fecgov/fec-cms/pull/4604#issuecomment-856247835

PR #4606, opened 2021-05-06
Feature/4170 remove unused css

### LINKED ISSUE #4170: Optimization: Remove unused CSS
### Background
Looking to [optimize the site's performance](https://github.com/fecgov/fec-cms/issues/3711) and that of the homepage specifically, the Lighthouse tool in Chromium browsers has made some suggestions. Some of these will have repercussions throughout the app so we're going to make them their own tickets.

# Recommendation: Remove unused CSS
Remove dead rules from stylesheets and defer the loading of CSS not used for above-the-fold content to reduce unnecessary bytes consumed by network activity. [details](https://web.dev/unused-css-rules/?utm_source=lighthouse&utm_medium=devtools)

## Overview
For the homepage in particular, we're including the elections-hash.css; Is this necessary?
Would like to prune the css so we're only including what is actually being used.

## Benefit (the 80 of 80/20)
Little? Likely less than its effort

## Effort (the 20 of 80/20)
Minimal to extensive
We could simply remove unused script tags and leave everything else alone
and/or
We could re-structure scss/css files and/or build so each page loads exactly the CSS it needs.

## Complications
- With all the sass inheritance, we could quickly get tangled up in untangling variables, mixins, etc.
- While untangling, we could wind up making our lives harder by duplicating code
- With templates being shared and inherited, we should look out for removing css from one page when it's being used by another page that's using the same template—could require additional templates.
- Will need to account for additional modules and feature-flagged features: we should make sure to include necessary css for flag-activated features but not necessarily include it when the feature isn't 'on'.
- Could inadvertently abandon non-desktop styles

#### Related tickets
- [Preload key requests](https://github.com/fecgov/fec-cms/issues/4168)
- [Remove unused JavaScript](https://github.com/fecgov/fec-cms/issues/4169)
- [Preconnect to required origins](https://github.com/fecgov/fec-cms/issues/4171)
- [Eliminate render-blocking resources](https://github.com/fecgov/fec-cms/issues/4172)
- [Ensure text remains visible during webfont load](https://github.com/fecgov/fec-cms/issues/4173)
- [Reduce the impact of third-party code](https://github.com/fecgov/fec-cms/issues/4175)
- [Does not use passive listeners to improve scrolling performance](https://github.com/fecgov/fec-cms/issues/4176)
- [Image elements do not have explicit width and height](https://github.com/fecgov/fec-cms/issues/4177)
- [Serve static assets with an efficient cache policy](https://github.com/fecgov/fec-cms/issues/4178)
- [Minimize main-thread work](https://github.com/fecgov/fec-cms/issues/4179)
- [Avoid an excessive DOM size](https://github.com/fecgov/fec-cms/issues/4180)
- [Reduce JavaScript execution time](https://github.com/fecgov/fec-cms/issues/4181)
- [Avoid enormous network payloads](https://github.com/fecgov/fec-cms/issues/4182)
- [[aria-hidden="true"] elements contain focusable descendents](https://github.com/fecgov/fec-cms/issues/4183)
- [Elements with an ARIA [role] that require children to contain a specific [role] are missing some or all of those required children](https://github.com/fecgov/fec-cms/issues/4184)
- [[aria-*] attributes do not have valid values](https://github.com/fecgov/fec-cms/issues/4185)
- [ARIA IDs are not unique](https://github.com/fecgov/fec-cms/issues/4186)
- [Lists do not contain only <li> elements and script supporting elements (<script> and <template>)](https://github.com/fecgov/fec-cms/issues/4187)
- [[user-scalable="no"] is used in the <meta name="viewport"> element or the [maximum-scale] attribute is less than 5](https://github.com/fecgov/fec-cms/issues/4188)
- [Links to cross-origin destinations are unsafe](https://github.com/fecgov/fec-cms/issues/4189)
- [Serves images with low resolution](https://github.com/fecgov/fec-cms/issues/4190)

rfultz:
## Summary

- Resolves #4170 

Our homepage was loading elections.css, which is a huge file.
Inspector's coverage report says we're using ~13% of it.
So let's get rid of 85% of it!

Here's a possible approach—whatcha think?

### Required reviewers

Two

The homepage's styles / appearance and our build process.

- New file `fec/static/scss/home.scss` is a duplicate of elections.css.
- New file `fec/home/templates/sample-homepage-for-purgecss.html` is a copy and paste of a compiled homepage with notes / instructions at the top for how this all works.
- New Gulp task ('purgecss')
- New npm script `build-homepage-css` builds it.

## Screenshots

Inspector's coverage should no longer list elections-*css. It should include home-*.css, but the Unused Bytes should be way less than 87%
image

Here's the 'before':
image

No visual changes:
image

## Related PRs

None

## How to test

- pull the branch
- `npm i` to install the new package (gulp-purgecss)
- `npm run build` to build the css files, including the new home.css
- `npm run build-homepage-css` to take home.css and throw away everything not needed, according to the new sample homepage template
- `./manage.py runserver`
- Check every detail of the homepage. Don't forget:
  - Multiple widths
  - The skip navigation link (Should be highlighted on first tab, but you can tab backward from the top nav, too)
  - Blue dev banner at the top
  - US Gov banner
  - Top nav, open
  - Top nav, closed
  - Red announcement banners under the hero
  - Search box
  - Typeaheads
  - Rollovers
  - Interactive components
  - Feedback
  - Glossary
  - Bottom nav

Lighthouse should no longer tell us to remove unused css (because of elections-*.css _or_ home-*.css)


patphongs (approved):
Thanks for walking through this with me @rfultz, this looks good to go

johnnyporkchops (approved):
Looks good, great job reducing loaded resource size!

PR #4649, opened 2021-05-24
Update IE support messaging

### LINKED ISSUE #4648: Update website messaging that we no longer support IE
### Summary

**What we're after:**
_As a team, we have decided that we will no longer be supporting Internet Explorer for our website._

### Completion criteria

- [x] Update IE messaging to reflect that we are no longer supporting Internet Explorer

patphongs:
## Summary

- Resolves #4648 

Updated messaging to indicate that we are no longer supporting Internet Explorer.

### Required reviewers

1 front end and 1 content

-  Anyone using Internet Explorer

## Screenshots

Screen Shot 2021-05-26 at 12 53 20 PM

## How to test

- For content staff: take a look at the screenshot to see if the language in the black banner for no longer support IE looks ok.
- For front end: check out this branch and try it in Internet Explorer. Make sure this banner appears for all versions of Internet Explorer.

- [WIKI on testing local dev server on Windows](https://github.com/fecgov/fec-cms/wiki/Testing-a-dev-branch-on-iPhone,-Internet-Explorer-Windows,-Android)


rfultz:
❤️

JonellaCulmer:
Browser support pages/banners on public websites that no longer support IE:
- https://www.hennepin.us/your-government/open-government/browser-support
- https://www.usajobs.gov/Help/faq/troubleshoot/
- 
image

patphongs:
@kathycarothers I've updated the text based on our conversations in Slack. It's ready for you to take a look at the updated screenshot above. Thanks!

kathycarothers (approved):

johnnyporkchops:
Internet explorer conditional statements are not supported in `IE10` and above so these tags are ignored. See screenshots showing `IE9` recognizing the statements and showing the warning message, but not `IE10`.

There is a JS alternative, or a CSS 'hack'. I tested both and they seem to work.
JS Option:
```
   function isIE () {
     var myNav = navigator.userAgent.toLowerCase();
     return (myNav.indexOf('msie') != -1 || myNav.indexOf('trident') != -1) ? true : false;
   }
   
   if(isIE()){
     document.getElementById('ie_warning').style.display='block'
   }
 ```

  CSS solution, some may call a 'hack' . 
```

<style>
   #ie_warning {
      display:none;
       }
   @media screen and (-ms-high-contrast: active), (-ms-high-contrast: none), (-ms-high-contrast: black-on-white), (-ms-high-contrast: white-on-black) {  
       #ie_warning {
       display:block;
       }
   }
</style>
  
  ```
  There are a few variations on using CSS to conditionally apply styles in IE only which can be found by googling.
  
  @patphongs , let me know if you want me to push up changes with one of these solution.
 
 
 <hr> 
 
 ##  Screenshots confirming support for conditional comments:
  
  ### IE9 shows warning 
  
ie9

### IE10 Does not show warning:

ie10

fec/fec/templates/base.html
```diff
@@ -19,10 +19,10 @@
   </head>
 
   <body class="{% block body_class %}{% endblock %}">
-    <!--[if lte IE 10]>
+    <!--[if lte IE 11]>
```
johnnyporkchops, line 22: Internet explorer conditional statements are not supported in IE10 and above so these tags are ignored. See comment inn conversation for screenshots and alternatives

johnnyporkchops (changes requested):

fec/fec/templates/base.html
patphongs: Thank you @johnnyporkchops for catching this! I've pushed up a new commit to use the JS conditional you provided instead of the IE HTML conditional statements. Could you please take another look at the homepage, data pages, and wagtail pages to make sure the IE warning isn't showing up unless it's IE?

johnnyporkchops:
@patphongs The JS solution works to show the warning in IE only on all pages and I can approve it but...

Do you think there it might be  better to use the  CSS solution for something like this?  I did not completely  think through the implementation. Now that I see that the JS snippet has to go in four places, I wonder  if it would more manageable if we could find one css files that gets included everywhere and put it there instead. (It works everywhere if we put it in  `_nav_scss`)

The CSS solution is referred to as a hack by definition I suppose, but using a small hack for dealing with IE seems fitting at this point.

I would strip it down to just detecting `(-ms-high-contrast: none)`, to make sure we don't accidentally target an obscure MS Edge user. That is the one feature not supported above 11. Confirmed by testing this on Edge and IE 10/11 locally with Windows high contrast settings. https://docs.microsoft.com/en-us/previous-versions/hh771830(v=vs.85)

patphongs:
> The CSS solution is referred to as a hack by definition I suppose, but using a small hack for dealing with IE seems fitting at this point.

@johnnyporkchops I can see your point, but when I look up the CSS version of using `-ms-high-contrast`, it says that it's non-standard and not to use it on production sites: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast. So I feel more inclined to use the JS version even though it's more bulky. Thoughts?

johnnyporkchops:
> > The CSS solution is referred to as a hack by definition I suppose, but using a small hack for dealing with IE seems fitting at this point.
> 
> @johnnyporkchops I can see your point, but when I look up the CSS version of using `-ms-high-contrast`, it says that it's non-standard and not to use it on production sites: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast. So I feel more inclined to use the JS version even though it's more bulky. Thoughts?

That info ^^ is why it works, because it is only supported in <= IE11. So we are not actually trying to use the feature, just detecting for non-modern browsers that support it.

johnnyporkchops:
https://stackoverflow.com/questions/18907131/detecting-ie11-using-css-capability-feature-detection

johnnyporkchops (approved):
Tested in IE 11/10 and Edge and modern browsers works as expected. 
Even tested in Edge with Windows ms-high-contrast settings enabled and does not show in Edge. 
Thanks Pat!

PR #4652, opened 2021-05-24
Removes hyphens in cash on hand

dorothyyeager:
Takes out hyphens in cash on hand in two different entries (cash on hand and ending cash on hand).

And also in Net outstanding campaign obligations

And also in the presidential map.

image


patphongs (approved):
Looks good, thanks @dorothyyeager !

PR #4655, opened 2021-05-25
Activate state election cards on election profile pages

### LINKED ISSUE #4654: Link state cards on election profile pages
### Summary

**What we're after:**
_Link state cards on election profile pages to pre-filtered election search by state._

Screen Shot 2021-05-25 at 1 59 27 PM

### Completion criteria

- "All - [state] elections" card is linked on the house and senate election profile pages

patphongs:
## Summary

- Resolves #4654 

Activate state election cards and links them on the election profile pages

### Required reviewers

1 front end and 1 UX

-  Senate and house election profile pages

### Presidential election profile page
Screen Shot 2021-05-27 at 9 57 12 AM

### Senate election profile page
Screen Shot 2021-05-27 at 9 56 16 AM

### House election profile page
Screen Shot 2021-05-27 at 9 56 43 AM

## How to test

- checkout this branch
- `./manage.py runserver`
- Go to each type of office (President, Senate, and House) election profile page. Ex: http://localhost:8000/data/elections/president/2020/, http://localhost:8000/data/elections/senate/CA/2022/, http://localhost:8000/data/elections/house/CA/01/2022/, 
- Check that the:
    - [ ] Has the correct election cards displayed up at the top
    - [ ] Election card at the top is active and 
    - [ ] Links back to the correct state and year search on the election search page

### DIFF
--- a/fec/data/templates/partials/elections/election-profile-cards.jinja
+++ b/fec/data/templates/partials/elections/election-profile-cards.jinja
@@ -24,15 +24,17 @@
         </aside>
     </div>
   {% else %}
-    <div class="grid__item is-disabled">
-      <aside class="card card--horizontal card--neutral-base">
-        <div class="card__image__container">
-          <span class="card__icon i-election"><span class="u-visually-hidden">Icon of elections</span></span>
-        </div>
-        <div class="card__content">
-          All {% if state %} - {{ state|fmt_state_full }} {% endif %}state elections
-        </div>
-      </aside>
+    <div class="grid__item">
+      <a href="/data/elections/?cycle={{ cycle }}&state={{ state }}">
+        <aside class="card card--horizontal card--neutral-base">
+          <div class="card__image__container">
+            <span class="card__icon i-election"><span class="u-visually-hidden">Icon of elections</span></span>
+          </div>
+          <div class="card__content">
+            All {% if state %} - {{ state|fmt_state_full }} {% endif %}state elections
+          </div>
+        </aside>
+      </a>
     </div>
   {% endif %}
   <div class="grid__item is-disabled">
--- a/fec/data/templates/partials/elections/election-profile-cards.jinja
+++ b/fec/data/templates/partials/elections/election-profile-cards.jinja
@@ -22,7 +22,7 @@
             <span class="card__icon i-election"><span class="u-visually-hidden">Icon of elections</span></span>
           </div>
           <div class="card__content">
-            All {% if state %} {{ state|fmt_state_full }} {% endif %}state elections
+            All federal {% if state %} {{ state|fmt_state_full }} {% endif %} elections
           </div>
         </aside>
       </a>
@@ -34,7 +34,7 @@
         <span class="card__icon i-election"><span class="u-visually-hidden">Icon of elections</span></span>
       </div>
       <div class="card__content">
-        All {{ office|title }} elections
+        All federal {{ office|title }} elections
       </div>
     </aside>
   </div>
--- a/fec/data/templates/partials/elections/election-profile-cards.jinja
+++ b/fec/data/templates/partials/elections/election-profile-cards.jinja
@@ -13,25 +13,16 @@
     </a>
   </div>
   {% if office == 'president' %}
-    <div class="grid__item is-disabled" style="display: none;">
-        <aside class="card card--horizontal card--neutral-base">
-          <div class="card__image__container">
-            <span class="card__icon i-election"><span class="u-visually-hidden">Icon of elections</span></span>
-          </div>
-          <div class="card__content">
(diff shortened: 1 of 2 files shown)


rfultz (approved):
Works well!
Functionality is approved, but I have design/content questions I'll add to the thread

rfultz:
Works well!

Is it weird that we have three buttons across the top and they're three different colors? I thought maybe it was a current-active-inactive thing but no.

Also, why is it "All House elections" but "All - [state name] state elections"? I'd like to remove the… hyphen? dash? (It's not an emdash and doesn't look like an endash)

rfultz (changes requested):
Works well!

It doesn't like medium-width browser windows

patphongs:
Thanks @rfultz, I've logged an issue with the medium-width browsers. This seems to be a thing for pages like this with a left nav, it doesn't stack well and some content is cut off. https://github.com/fecgov/fec-cms/issues/4656

JonellaCulmer:
Great catches!

> Is it weird that we have three buttons across the top and they're three different colors? I thought maybe it was a current-active-inactive thing but no.

There should only be two colors. As you said, an active and inactive state. @patphongs Let me know if you need anything from me. We should probably resolve this while we're messing with buttons, right?

> Also, why is it "All House elections" but "All - [state name] state elections"? I'd like to remove the… hyphen? dash? (It's not an emdash and doesn't look like an endash)

I agree. We should drop the hyphen. @dorothyyeager just in case you have thoughts.

dorothyyeager:
I concur, drop the hyphen.

patphongs:
Thanks for your reviews @JonellaCulmer @rfultz. I've updated by removing the hyphen and using the primary colors for the state card. Could you have another look please?

Screen Shot 2021-05-26 at 10 03 35 AM

JonellaCulmer:
@patphongs It looks great with the colors fixed. A couple things I noticed:
- The election icons, for the state card and inactive all house card are two different sizes. Can you make the inactive election icon the same size as the state card?
- We use the default language "....state elections" for the state cards, but not all are states. Like Guam or Northern Mariana Islands. Can we customize the language for certain places if we know their not states? And depending on the year, most of those places only have the one election going on.  cc: @dorothyyeager @kathycarothers Let me know if you have thoughts on this.
Example:
Screen Shot 2021-05-26 at 10 37 39 AM

dorothyyeager:
For the territories, you are right @JonellaCulmer - usually they only have one (federal) election going on and sometimes it's at weird times. For example, I think Puerto Rico does their elections in odd-numbered years. (Unless the plural refers to primaries and generals.) 

Maybe we just take out the word "state" to resolve the territory part of this? I kind of find it confusing anyway considering we are talking federal elections.

patphongs:
> Maybe we just take out the word "state" to resolve the territory part of this? I kind of find it confusing anyway considering we are talking federal elections.

@dorothyyeager @JonellaCulmer I was thinking to remove the word "state" as well, that's certainly a simpler solution.

As for the "All [state_or_territory_name] elections" card, the wording on that is difficult because there's multiple possibilities of what we'll show depending on year. The fact that we will sometimes show the presidential election when it's a presidential year is problematic because that's a federal election, not a state specific one, so we can't really call it, for example, a California election.

- Possibility 1 - Display president, house, and senate elections. Ex: https://www.fec.gov/data/elections/?cycle=2020&state=AL
- Possibility 2 - Display senate and house elections. Ex: https://www.fec.gov/data/elections/?cycle=2018&state=AL
- Possibility 3 - Display only house elections. Ex: https://www.fec.gov/data/elections/?cycle=2006&state=AL

If we cross that hurdle, this is another thing that I'm not sure about. The plural of the word "elections", doesn't sound grammatically correct in the singular if we use the word "all". Any suggestions?

patphongs:
> The election icons, for the state card and inactive all house card are two different sizes. Can you make the inactive election icon the same size as the state card?

Done! Pushed up a commit to update this.

dorothyyeager:
@patphongs The only thing I can think of is to ditch "all" and just say "Northern Marianas elections" but as you pointed out, President is a nationwide election (although each state/territory administers it in their locality). Maybe "Northern Marianas federal ballot"? But that's clunkier.

patphongs:
After discussing with @JonellaCulmer @dorothyyeager @kathycarothers, we've decided to update the card names to include the word "federal". This is what they'll look like: "All federal [state_or_territory_name] elections", "All federal House elections", and "All federal Senate elections". 

@JonellaCulmer @rfultz It's all ready for your review again, thanks!

JonellaCulmer (approved):
Everything looks great. Thanks, @patphongs

rfultz (approved):
Looks good!

PR #4685, opened 2021-06-07
Sync string values of environment names for GTM

rfultz:
## Summary

- Resolves 

Not sure if these got out of sync or never were in sync, but GTM is using the Prod (default) settings because the string values for Dev and Stage weren't the same as are being set in base.py

### Required reviewers

1-2

Only which version of GTM is being used for Dev and for Stage.

## Screenshots

No visual changes

## Related PRs

None

## How to test

- Will have to test on Dev and Stage.
- Load any page and look at its source.
- Dev should mention `_vM9qqeervv47SONwk4KRg`
- Stage should mention `ZYJODko0wNc0p1iWNPgvFw`
- Neither Dev nor Stage should mention `EDR0yhH3jo_lEfiev6nbSQ`


patphongs (approved):
These new values look correct to me

PR #4743, opened 2021-07-02
4630 Only link to document.file_url if it exists

### LINKED ISSUE #4630: Server error when no PDF url exists
### Summary

_When the PDF URL field is empty, a server error occurs on the page._

```
   ERR     return self.file_url.rsplit('.', 1)[1].upper()
   ERR IndexError: list index out of range
```

image

### How to Reproduce

1. Go to any report and remove the PDF URL completely.
1. Publish the report
2. Go to the feed that the report is in and you will get a server error.

Screen Shot 2021-05-17 at 3 19 54 PM

rfultz:
## Summary

- Resolves #4630 

Since the template crashes if fec_url isn't specified and there's no file at that url, let's not link to the non-file url

### Required reviewers

Just a couple—feel free to edit the list

Only the document.html template

## Screenshots

The testing document has no url so isn't a link and doesn't have an icon or file size
image

## Related PRs

None

## How to test

- pull the branch
- go to the [list of reports](http://127.0.0.1:8000/about/reports-about-fec/agency-operations/)
- Make a mental note of a report with a link
- Find that report in Wagtail
- Remove its url
- Publish it
- Go back to the list of reports
- The report should still be listed but shouldn't be a link and shouldn't have an icon or a file size listed
- All other reports in the list should be unchanged

Alternatively, make a new report with no url, publish it, and make sure it doesn't get a linked title

### DIFF
--- a/fec/home/templates/partials/document.html
+++ b/fec/home/templates/partials/document.html
@@ -16,15 +16,17 @@
         <a class="js-read-more post__read-more" href="{{ document.url }}">Read more</a>
     </div>
   {% else %}
-    {% if document.fec_url and document.extension %}
     <h3>
+      {% if document.file_url %}
       <a href="{{ document.file_url }}">{{ document.title }}</a>
+      {% else %}
+      {{ document.title }}
+      {% endif %}
+      {% if document.extension is not '' %}
       <i class="icon icon--inline--right icon--inline--left i-document"></i>
       <span class="t-sans t-normal">({{ document.extension }}){% if document.size %} | ({{ document.size }}){% endif %}</span>
+      {% endif %}
     </h3>
-    {% else %}
-    <h3>{{ document.title }}</h3>
-    {% endif %}
   {% endif %}
     <div class="post__meta">
       <a class="tag tag--secondary" href="">{{ document.category }}</a>
--- a/fec/home/templates/partials/document.html
+++ b/fec/home/templates/partials/document.html
@@ -16,11 +16,15 @@
         <a class="js-read-more post__read-more" href="{{ document.url }}">Read more</a>
     </div>
   {% else %}
+    {% if document.fec_url and document.extension %}
     <h3>
       <a href="{{ document.file_url }}">{{ document.title }}</a>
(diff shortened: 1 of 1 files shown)


johnnyporkchops:
@rfultz, Should we make the field required so editors have to link the report to a document, otherwise it seems like the entry for that report would be useless without something to click through to?
Since it's a "DocumentPage" I would assume we need a document to link to. I don't know if there are any use cases for a report entry with no link?

Heads up, I noticed that the API PR referenced in the issue https://github.com/fecgov/openFEC/pull/4865 is not actually related to document links in Wagtail, which is a bit confusing.

dorothyyeager:
@johnnyporkchops There are a few "reports" that are web pages and not PDFs (if that's what you're asking):
https://www.fec.gov/about/reports-about-fec/agency-operations/impact-national-voter-registration-act-1993-administration-federal-elections-html/
https://www.fec.gov/about/reports-about-fec/agency-operations/e-filing-study-2016/

johnnyporkchops:
> @johnnyporkchops There are a few "reports" that are web pages and not PDFs (if that's what you're asking):
> https://www.fec.gov/about/reports-about-fec/agency-operations/impact-national-voter-registration-act-1993-administration-federal-elections-html/
> https://www.fec.gov/about/reports-about-fec/agency-operations/e-filing-study-2016/

@dorothyyeager Those are ResourcePages that are made as children off ReportsAboutFEC. I was asking if you have a use case for having a DocumentPage, like in the screenshot example above, where it does not link to a report (or does not link to a report yet, because you are awaiting the final doc or something)?

dorothyyeager:
@johnnyporkchops The only time I can think of that happening would be if the Information Division started work on a page for uploading documents to be used in a training, but didn't yet have the documents ready. I'm thinking of these pages:
https://www.fec.gov/help-candidates-and-committees/trainings/webinar-materials/
https://www.fec.gov/help-candidates-and-committees/trainings/conference-materials/

johnnyporkchops (approved):
Looks good. I was concerned that this field should be made  required as well, but the fix with this PR solves the  immediate problem of the whole Reports section getting server error when the field is blank.

PR #5045, opened 2022-01-28
Landing template for house-senate overview pages

### LINKED ISSUE #5019: All House and Senate election overview pages: Build new landing page template
**What we're after:**
All House and Senate election overview pages: Build new landing page template. 

This work requires the creation of a new page template based on our others. (front-end lead, with design)

Refer to the mockup in the epic.
https://github.com/fecgov/fec-epics/issues/213

**Completion criteria:**
- [ ] Develop new template for overview pages for House and Senate

johnnyporkchops:
## Summary
- Landing template for house-senate overview pages based on current profile pages (main template snd partials) 
- Starting point for the view in `views.py`
- entries in `urls.py`
- Two new select  macros that  we may or may not actually use for these, but, at a minimum, helps as placeholders
- TODO comments for code and patters yet to be determinned
- **Please note:** This initial template departs slightly from mockup as it was done using existing patterns but we can certainly make changes wherever necessary

Mockup: https://github.com/fecgov/fec-epics/issues/213

- Resolves #5019

### Required reviewers
One frontend, one UX

	new file:   data/templates/house-senate-overview.jinja
	modified:   data/templates/macros/cycle-select.jinja
	new file:   data/templates/partials/house-senate-overview/election-data.jinja
	new file:   data/templates/partials/house-senate-overview/sidebar-nav.jinja
	modified:  data/urls.py
	modified:  data/views.py

## Screenshots

Screen Shot 2022-01-28 at 11 06 59 AM
<hr>

### Or narrower content areas

Screen Shot 2022-01-28 at 10 58 14 AM

## How to test
- Checkout and run branch
- Test page with /house, /senate added to end of url (defaults to senate)
    -  http://127.0.0.1:8000/data/house-senate-overview/
- Please comment on things we need to change and/or flesh out more
Mockup: https://github.com/fecgov/fec-epics/issues/213

### DIFF
--- a/fec/data/templates/house-senate-overview.jinja
+++ b/fec/data/templates/house-senate-overview.jinja
@@ -1,0 +1,56 @@
+{% extends 'layouts/main.jinja' %}
+{% import 'macros/page-header.jinja' as header %}
+{% if office == 'house' %}
+  {% set crumb = office|title  %}
+{% elif office == 'senate' %}
+  {% set crumb = office|title %}
+{% else %}
+  {% set crumb = office|title %}
+{% endif %}
+
+
+
+{% block title %}{{ title }}{% endblock %}
+
+{% block css %}
+  <link rel="stylesheet" type="text/css" href="{{ asset_for_css('elections.css') }}" />
+{% endblock %}
+{% block body %}
+  {{ header.header(title, crumb) }}
+  <div class="u-padding--left u-padding--right tab-interface">
+    <header class="main">
+      <div class="heading--section">
+        <h1 class="entity__name">
+          {{ office|title }} elections
+        </h1>
+      </div>
+    </header>
+    <div class="data-container__wrapper">
+      {% include 'partials/house-senate-overview/sidebar-nav.jinja' %}
+      {% include 'partials/house-senate-overview/election-data.jinja' %}
+      <div class="main__content--right-full">
+
+      </div>
+    </div>
+  </div>
+{% endblock %}
+{% block scripts %}
+<script>
+var context = {
+  election: {
+    cycle: '{{ cycle }}',
+    election_full: true,
+    duration: '{{ election_duration }}',
+    office: '{{ office }}',
+    state: '{{ state or '' }}',
+    stateFull: '{{ state_full or '' }}',
+    district: '{{ district or '' }}'
+  }
+};
+
+</script>
+<script src="{{ asset_for_js('dataviz-common.js') }}"></script>
+<script src="{{ asset_for_js('elections.js') }}"></script>
+<script src="{{ asset_for_js('committee-single.js') }}"></script>
+{% endblock %}
+
--- a/fec/data/urls.py
+++ b/fec/data/urls.py
@@ -18,6 +18,8 @@ urlpatterns = [
     url(r'^data/elections/$', views.elections_lookup),
     url(r'^data/raising-bythenumbers/$', views.raising),
     url(r'^data/spending-bythenumbers/$', views.spending),
+    url(r'^data/house-senate-overview/$', views.house_senate_overview),
+    url(r'^data/house-senate-overview/(?P<office>\w+)/$', views.house_senate_overview, name=''),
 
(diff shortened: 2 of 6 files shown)


fec/data/templates/house-senate-overview.jinja
```diff
@@ -0,0 +1,56 @@
...
+  {% set crumb = office|title  %}
+{% elif office == 'senate' %}
+  {% set crumb = office|title %}
+{% else %}
+  {% set crumb = office|title %}
+{% endif %}
+
+
+
+{% block title %}{{ title }}{% endblock %}
```
patphongs, line 13: This looks like it's looking for `Election United States` within the `page-header.jinja` for the breadcrumbs to come out correctly. In addition, can you pass the office for the title: `{{ office | title }}`

fec/data/urls.py
```diff
@@ -18,6 +18,8 @@
     url(r'^data/elections/$', views.elections_lookup),
     url(r'^data/raising-bythenumbers/$', views.raising),
     url(r'^data/spending-bythenumbers/$', views.spending),
+    url(r'^data/house-senate-overview/$', views.house_senate_overview),
+    url(r'^data/house-senate-overview/(?P<office>\w+)/$', views.house_senate_overview, name=''),
```
patphongs, line 22: Eventually @JonellaCulmer @johnnyporkchops, won't the path for these be `/data/elections/house/` or `/data/elections/senate/`? Those paths don't currently exist. Something else to think about is if we'll want to be able to pass election year parameters into the URL, allowing for URLs to be persistent and shared around more easily. For example, `/data/elections/senate/2022/`. This may get a bit confusing though because of the contributions over time feature, it'll have a min election year and max election year. So we have to figure out how to default that too. Also, in concept, they'll be able to change any of the features to different years, as there's no one year controller.

patphongs (commented):
A few observations I had @johnnyporkchops

fec/data/urls.py
JonellaCulmer: @patphongs I'm fine with those paths for the new pages. Simple and clean. Beyond that I don't have a preference. Something easily shared and bookmarked would be ideal and what you've suggested seems to fit the bill.

johnnyporkchops: > Eventually @JonellaCulmer @johnnyporkchops, won't the path for these be `/data/elections/house/` or `/data/elections/senate/`? Those paths don't currently exist. Something else to think about is if we'll want to be able to pass election year parameters into the URL, allowing for URLs to be persistent and shared around more easily. For example, `/data/elections/senate/2022/`. This may get a bit confusing though because of the contributions over time feature, it'll have a min election year and max election year. So we have to figure out how to default that too. Also, in concept, they'll be able to change any of the features to different years, as there's no one year controller.

@pat, We would probably have to utilize the JS history API to be able to have a sharable URL that always represent the state of all the selects (and vice versa  at any given time).  One way would  be to prefix the parameters we pass to the query string for each  feature interaction (ie: summary-election-year, raising-election-year etc.).

patphongs: `two-year-period` I think is clearer. For now @johnnyporkchops, could you update the path and put it under a feature flag? I think the bones of the template are good, we just have to figure out more of how the logic will work once the design is more finalized following UX research.

JonellaCulmer:
@johnnyporkchops How easy will it be to modify that first header "House election data" later, in case we want to change the language?

johnnyporkchops:
> @johnnyporkchops How easy will it be to modify that first header "House election data" later, in case we want to change the language?
@JonellaCulmer  We can change the headers any time, no problem.

JonellaCulmer (approved):

fec/data/templates/house-senate-overview.jinja
patphongs: @johnnyporkchops This specific text need to be added to the HTML title tags `Election United States` for the breadcrumbs to function correctly. But we can save that to do later since this is the initial concept code under a feature flag.

PR #5141, opened 2022-03-29
Stop using JavaScript to hide columns

rfultz:
## Summary

- Resolves #4913 

### Required reviewers

- front-end
- UX
- someone else to help check possible weirdness

Potentially all datatables

## Screenshots

image

left: initial load;
right: after opening a details panel on the smallest browser width

## Related PRs

None

#### How to replicate

- go to a [datatables page](http://127.0.0.1:8000/data/receipts/?data_type=processed&two_year_transaction_period=2022&min_date=01%2F01%2F2021&max_date=12%2F31%2F2022)
- shrink the browser window to the smallest width
- reload the page
- right-click the first Source name listed and Inspect
- see how there are seven `<td>`? That quantity is what we're watching
- back to the main window
- close the filters (they're just in the way)
- click one of the ⊕ in the right column of the table
- see how now there are fewer `<td>`?
- now expanding the browser window won't restore those `<td>` because they no longer exist
- that's what this PR is trying to fix

#### Testing this PR

- pull the branch
- maybe `npm i` but likely not
- `npm run build` (or just `npm run build-js`)
- `./manage.py runserver`
- go to the datatables page linked in 'How to replicate'
- follow the replication steps
- now, expanding the browser window after having opened and closed the ⊕ panel should show the same columns as a wide-layout page load

#### Side effects

- **Does this break anything on other datatables pages?**
- This won't be a fun test.
- We'll need to test opening and close panels on different width on every template.
- Why was `api.columns…visible()` being used at all? Was it just left-over from before the `hide-panel-*` classes were implemented? There's a comment in tables.js: `// TODO figure way to share these values with CSS.`

## Next steps

If approved, remove the commented code from tables.js

### DIFF
--- a/fec/fec/static/js/modules/tables.js
+++ b/fec/fec/static/js/modules/tables.js
@@ -174,8 +174,8 @@ function modalRenderFactory(template, fetch) {
             $row.toggleClass('row-active', true);
             $('body').toggleClass('panel-active', true);
             accessibility.restoreTabindex($modal);
-            var hideColumns = api.columns('.hide-panel');
-            hideColumns.visible(false);
+            // var hideColumns = api.columns('.hide-panel');
+            // hideColumns.visible(false);
 
             // Populate the pdf button if there is one
             if (fetched.pdf_url) {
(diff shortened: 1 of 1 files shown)


johnnyporkchops (commented):
@rfultz I think this code was to hide some columns to keep the details panel from getting pushed under the table at certain desktop breakpoints. I think the issue was that the columns were not reappearing  after the panel is closed. Is there a better way to solve that display  problem ?

Here is what I am seeing with this code removed on the PR branch:

details_panel_stacking_below_table

JonellaCulmer (approved):
Everything is working as it should again - no disappearing columns at the smallest widths! Thank you, @rfultz

Issue #5090, opened 2022-02-28
Draft methodology for contributions across time feature for all House and all Senate


JonellaCulmer:
**What we're after:**
We need to draft a methodology for the all House and all Senate contributions across time feature. 
Based on usability testing, users are interested in the following information in this methodology:
- Explanation of where the money came from
- How the data was compiled, formulas
- Include definitions of some of the terms
- What kinds of committees are included
- Plain language
- Coverage dates

**Completion criteria:**
- [x] Methodology is drafted 
- [ ] Methodology is reviewed and approved by content and dev teams

**Designs:**
<img width="709" alt="Screen Shot 2022-02-28 at 5 00 34 PM" src="https://user-images.githubusercontent.com/31663028/156065893-e658566c-78d6-4be8-82e9-810937cc0345.png">

JonellaCulmer:
@johnnyporkchops Here's the information to include in the methodology from our discussion with content. Do you mind updating your draft? Then we should tag some folks to look at it. 

**Individual contributions:**
Include Form 3, Line 11(a)(i)

**PACs and other committee contributions:**
Include Form 3, Line 11(c). All itemized contributions from PACs and other political committees, including state PACs and other federal candidate committees. Party committee and Super PAC contributions are not included.

**Transfers:**
Include Form 3, Line 12

**General methodology language additions:**
Include link to archived contribution limits page.

johnnyporkchops:
@JonellaCulmer I  incorporated  this language into the draft in the [comment above ](https://github.com/fecgov/fec-cms/issues/5090#issuecomment-1071011390)

dorothyyeager:
Just my thought on this line; I don't feel strongly about this if others disagree:

```
PACs and other committee contributions:
Include Form 3, Line 11(c). All itemized contributions from PACs and other political committees, including state PACs and other federal candidate committees. Party committee and Super PAC contributions are not included.
```

Suggest changing to:
Other committee contributions
Include Form 3, Line 11(c). All itemized contributions from other political committees, including federally registered PACs, committees registered at a state or local level and other federal candidate committees. Party committee and Super PAC contributions are not included.

Also for this line: This data is generated from [FEC Form 3- REPORT OF RECEIPTS AND DISBURSEMENTS](https://www.fec.gov/resources/cms-content/documents/fecfrm3.pdf) 
Take out "FEC" (content guide thing).

PaulClark2:
> PACs and other committee contributions:
[...]

Issue #5131, opened 2022-03-23
All elections totals table: Front-end to begin implementation of table


johnnyporkchops:
### Summary

**What we're after:**
All elections totals table: Front-end to begin implementation of table
<img width="653" alt="Screen Shot 2022-03-23 at 1 46 04 PM" src="https://user-images.githubusercontent.com/5572856/159763809-b4e20f94-8b55-4331-8ffd-748a48427e73.png">

### Related issues
Epic: https://github.com/fecgov/fec-epics/issues/213

### Completion criteria

The All elections totals table is built

### Tech steps or considerations (optional)

- [x] What endpoint do we use for this?
  - [x] Use  new aggregate endpoints being created in this issue https://github.com/fecgov/openFEC/issues/5026
- [x] Fix wrapping disclaimer text on [contributions across time](https://github.com/fecgov/fec-cms/pull/5166#issuecomment-1115160985)

PR #5166, opened 2022-04-15
Contributions across time chart for House/Senate Overview page

### LINKED ISSUE #5034: Contributions across time: Build out chart
**What we're after:**
Build out the contributions across time chart. Pair with @JonellaCulmer as much as necessary or to answer any questions. 

**Completion criteria:**
- [x] UX and frontend team meet to establish user interaction for two-year selects and resulting data presentation.
- [ ] Chart is built out and ready for QA testing by front and back-end teams (17.5)

**Design:**
Screen Shot 2022-02-28 at 5 00 34 PM

johnnyporkchops:
## Summary 

Chart for comparing  contributions over time by  two-year periods starting at  current election year

Tested in Chrome, Firefox, Safari, Opera, MS Edge Chromium/Legacy on WIndows

- Resolves #5034

Related issues:
Methodology: https://github.com/fecgov/fec-cms/issues/5090

### Required reviewers

Two frontend (ideally) , one UX

	modified:   data/templates/house-senate-overview.jinja
	modified:   data/templates/partials/house-senate-overview/election-data.jinja
	modified:   data/views.py
	new file:   fec/static/js/pages/contributions_across_time.js
	new file:   fec/static/scss/widgets/house-senate-overview.scss

<hr>

### Desktop
   
Screen Shot 2022-04-15 at 12 31 58 AM

### Responsive:

cat_reponsive

<hr>

## How to test
This is deployed  to feature: https://fec-feature-cms.app.cloud.gov/data/elections/senate/#contributions-over-time but please run locally too. 
- [ ] Remove feature deploy task before merging
- Checkout branch and `npm run build`
- Go to http://127.0.0.1:8000/data/elections/house/ and http://127.0.0.1:8000/data/elections/senate/
- Test it 
- Any  responsive issues?: The meter bars are set to a percentage width and so everything stays on one line until you resize to `<400px` or `Iphone SE` in the Chrome responsive inspector,  at which point the items stack.
-  You could also test the logic that sets current election year (and offers nine periods back) in the selects by changing the setting in `main.jinja` to something like `DEFAULT_ELECTION_YEAR = '2024';`

### DIFF
--- a/fec/data/templates/house-senate-overview.jinja
+++ b/fec/data/templates/house-senate-overview.jinja
@@ -36,21 +36,22 @@
 {% endblock %}
 {% block scripts %}
 <script>
-var context = {
-  election: {
-    cycle: '{{ cycle }}',
-    election_full: true,
-    duration: '{{ election_duration }}',
-    office: '{{ office }}',
-    state: '{{ state or '' }}',
-    stateFull: '{{ state_full or '' }}',
-    district: '{{ district or '' }}'
-  }
-};
+var context = {{ context_vars | to_json | safe }};
+
+// var context = {
+//   election: {
+//     cycle: '{{ cycle }}',
+//     election_full: true,
+//     duration: '{{ election_duration }}',
+//     office: '{{ office }}',
+//     state: '{{ state or '' }}',
+//     stateFull: '{{ state_full or '' }}',
+//     district: '{{ district or '' }}'
+//   }
+//};
 
 </script>
 <script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-<script src="{{ asset_for_js('elections.js') }}"></script>
-<script src="{{ asset_for_js('committee-single.js') }}"></script>
+<script src="{{ asset_for_js('contributions_across_time.js') }}"></script>
 {% endblock %}
 
--- a/fec/data/templates/house-senate-overview.jinja
+++ b/fec/data/templates/house-senate-overview.jinja
@@ -34,5 +34,5 @@
 var context = {{ context_vars | to_json | safe }};
 </script>
 <script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-<script src="{{ asset_for_js('contributions_across_time.js') }}"></script>
+<script src="{{ asset_for_js('elections-overview.js') }}"></script>
 {% endblock %}
--- a/fec/data/templates/house-senate-overview.jinja
+++ b/fec/data/templates/house-senate-overview.jinja
@@ -1,12 +1,6 @@
 {% extends 'layouts/main.jinja' %}
 {% import 'macros/page-header.jinja' as header %}
-{% if office == 'house' %}
-  {% set crumb = office|title  %}
-{% elif office == 'senate' %}
-  {% set crumb = office|title %}
-{% else %}
-  {% set crumb = office|title %}
-{% endif %}
+
 
 {{ cycle }} Election - US {{ office|title }} {% if state %} - {{ state|fmt_state_full }} {% if district %} - District {{ district }} {% endif %} {% endif %}
 
@@ -14,6 +8,7 @@
 
 {% block css %}
   <link rel="stylesheet" type="text/css" href="{{ asset_for_css('elections.css') }}" />
+  <link rel="stylesheet" type="text/css" href="/static/css/widgets/house-senate-overview.css" />
 {% endblock %}
 {% block body %}
   {{ header.header(title, crumb) }}
--- a/fec/data/templates/house-senate-overview.jinja
+++ b/fec/data/templates/house-senate-overview.jinja
@@ -32,21 +32,7 @@
 {% block scripts %}
 <script>
 var context = {{ context_vars | to_json | safe }};
-
-// var context = {
-//   election: {
-//     cycle: '{{ cycle }}',
-//     election_full: true,
-//     duration: '{{ election_duration }}',
-//     office: '{{ office }}',
-//     state: '{{ state or '' }}',
-//     stateFull: '{{ state_full or '' }}',
-//     district: '{{ district or '' }}'
-//   }
-//};
-
 </script>
 <script src="{{ asset_for_js('dataviz-common.js') }}"></script>
 <script src="{{ asset_for_js('contributions_across_time.js') }}"></script>
 {% endblock %}
-
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -108,7 +108,8 @@
       <div class="range__select range__select--min" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Beginning</label>
-          <select id="{{id}}-cycle" class="js-min-period-select" name="cycle" data-cycle-location="query">
+          <select id="{{id}}-cycle" class="js-min-period-select" name="cycle" data-cycle-location="query" data-period=
+          "min">
             {% for each in cycles | sort(reverse=True) %}
               <option
                   value="{{ each }}"
@@ -122,7 +123,8 @@
       <div class="range__select range__select--max" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Ending</label>
-          <select id="{{id}}-cycle" class="js-max-period-select" name="cycle" data-cycle-location="query">
+          <select id="{{id}}-cycle" class="js-max-period-select" name="cycle" data-cycle-location="query" data-period=
+          "max">
             {% for each in cycles | sort(reverse=True) %}
               <option
                   value="{{ each }}"
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -98,47 +98,6 @@
 {% endmacro %}
 
 
-{# range of two year periods.Select's class is "js-period-select". Unlike class="js-cycle-select", these do not auto-reload page upon change to a new window.locatio with query params. #}
- {% macro cycle_range_select(cycles, cycle=none, id='') %}
- {% set cycle = cycle | int %}
-  <div class="row content__section">
-  <label class="label t-inline-block">Two-year period date range</label>
-  <fieldset id="two-year-period-date-range" name="two-year-period-date-range">
-    <div class="range range--date js-date-range">
-      <div class="range__select range__select--min" data-filter="range">
-        <div class="cycle-select">
-          <label for="{{id}}-cycle">Beginning</label>
-          <select id="{{id}}-cycle" class="js-min-period-select" name="cycle" data-cycle-location="query" data-period=
-          "min">
-            {% for each in cycles | sort(reverse=True) %}
-              <option
-                  value="{{ each }}"
-                  {% if cycle and cycle <= each and cycle > (each - 2) %}selected{% endif %}
-                >{{ each|fmt_year_range }}</option>
-            {% endfor %}
-          </select>
-        </div>
-      </div>
-      <div class="range__hyphen">-</div>
-      <div class="range__select range__select--max" data-filter="range">
-        <div class="cycle-select">
-          <label for="{{id}}-cycle">Ending</label>
-          <select id="{{id}}-cycle" class="js-max-period-select" name="cycle" data-cycle-location="query" data-period=
-          "max">
-            {% for each in cycles | sort(reverse=True) %}
-              <option
-                  value="{{ each }}"
-                  {% if cycle and cycle <= each and cycle > (each - 2) %}selected{% endif %}
-                >{{ each|fmt_year_range }}</option>
-            {% endfor %}
-          </select>
-        </div>
-      </div>
-    </div>
-  </fieldset>
-  </div>
-{% endmacro %}
-
 {% macro two_year_select(cycles, cycle=none, id='') %}
 {% set cycle = cycle | int %}
   <div class="content__section">
--- a/fec/data/templates/macros/cycle-select.jinja
+++ b/fec/data/templates/macros/cycle-select.jinja
@@ -108,7 +108,7 @@
       <div class="range__select range__select--min" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Beginning</label>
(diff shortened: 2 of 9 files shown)


fec/data/templates/house-senate-overview.jinja
```diff
@@ -36,21 +31,8 @@
...
-    election_full: true,
-    duration: '{{ election_duration }}',
-    office: '{{ office }}',
-    state: '{{ state or '' }}',
-    stateFull: '{{ state_full or '' }}',
-    district: '{{ district or '' }}'
-  }
-};
-
+var context = {{ context_vars | to_json | safe }};
```
rfultz, line 34: Ooooh I like it

fec/data/templates/house-senate-overview.jinja
```diff
@@ -36,21 +31,8 @@
...
 </script>
 <script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-<script src="{{ asset_for_js('elections.js') }}"></script>
-<script src="{{ asset_for_js('committee-single.js') }}"></script>
+<script src="{{ asset_for_js('contributions_across_time.js') }}"></script>
```
rfultz, line 37: the underscores go against the convention of using dashes for js and css

fec/data/templates/macros/cycle-select.jinja
```diff
@@ -108,7 +108,8 @@
       <div class="range__select range__select--min" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Beginning</label>
```
rfultz, line 110: Is the label being removed from the page? I don't see it on https://fec-feature-cms.app.cloud.gov/data/elections/senate/#contributions-over-time

fec/data/templates/macros/cycle-select.jinja
```diff
@@ -122,7 +123,8 @@
       <div class="range__select range__select--max" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Ending</label>
```
rfultz, line 125: Is the label being removed from the page? I don't see it on https://fec-feature-cms.app.cloud.gov/data/elections/senate/#contributions-over-time

fec/data/templates/macros/cycle-select.jinja
```diff
@@ -108,7 +108,8 @@
       <div class="range__select range__select--min" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Beginning</label>
-          <select id="{{id}}-cycle" class="js-period-select" name="cycle" data-cycle-location="query">
+          <select id="{{id}}-cycle" class="js-min-period-select" name="cycle" data-cycle-location="query" data-period=
```
rfultz, line 111: On https://fec-feature-cms.app.cloud.gov/data/elections/senate/#contributions-over-time, the id is evaluating to "-cycle" and there are two elements (`<select>`) with the same id

fec/data/templates/macros/cycle-select.jinja
```diff
@@ -122,7 +123,8 @@
       <div class="range__select range__select--max" data-filter="range">
         <div class="cycle-select">
           <label for="{{id}}-cycle">Ending</label>
-          <select id="{{id}}-cycle" class="js-period-select" name="cycle" data-cycle-location="query">
+          <select id="{{id}}-cycle" class="js-max-period-select" name="cycle" data-cycle-location="query" data-period=
```
rfultz, line 126: On https://fec-feature-cms.app.cloud.gov/data/elections/senate/#contributions-over-time, the id is evaluating to "-cycle" and there are two elements (`<select>`) with the same id

fec/data/templates/partials/house-senate-overview/sidebar-nav.jinja
```diff
@@ -12,10 +12,13 @@
           {{ office|title }} election data
         </a>
         <ul>
+            {# 
             <li><a href="#election-summary">{{ office | title }} election summary</a></li>
+            #} 
             <li><a href="#contributions-over-time">{{ office | title }} candidate contributions over time</a></li>
-            <li><a href="#money-raised-accross-elections">Money raised accross {{ office | title }} elections</a></li>
+            {#  <li><a href="#money-raised-accross-elections">Money raised accross {{ office | title }} elections</a></li>
```
rfultz, line 19: Any reason to keep these—are we bringing them back soon? (if so, can we change to 'across')

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+
+      // Create start/end years for display on page
+      let electionYear = theResults[i].election_year;
+      let starYear = electionYear - 1;
+      let end_year = electionYear;
+      let twoYearPeriod = `${starYear} - ${end_year}`;
+
+      // Trick gitleaks for false-positive for "----"orized
+      let sub = 'au';
+      let subConcat = `${sub}thorized_committee`;
```
rfultz, line 108: 😆 I was wondering what was going on, then saw the comment

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+
+      // Create object for creating querystring for the link for each value
+      let searchFilters = {
+        two_year_transaction_period: electionYear,
+        recipient_committee_type: context.office_code,
+        line_number: lineNumbers[line]
+      };
+
+      // Generate the querystring for link from the searcFilters object
+      // Could be done more efficiently with URLSearchParams(), but since MS Edge-legacy does not support, we'll do this for now
```
rfultz, line 122: Excellent. Well, not excellent that we have to, but excellent that the comment explains why

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+      let totalUrl = buildAppUrl(['receipts']) + `?${queryString}`;
+
+      // HTML for each result row
+      let theInnerHTML =
+          `<div class="simple-table__row" role="row">
+            <div role="cell" class="simple-table__cell js-total-period">${twoYearPeriod}</div>
+            <div role="cell" class="simple-table__cell">
+              <meter min="0" max="${maxValue}" value="0" title="US Dollars"></meter>
+            </div>
+            <div role="cell" class="simple-table__cell js-total-value t-mono-stacked-currency"><a href=${totalUrl}>${textValue}</a></div>
```
rfultz, line 137: Does `<a href=${totalUrl}>` output with the href quoted? (If not, we should in case a space sneaks into the path and breaks the link)

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+    periodWrapper.innerHTML = theInnerHTML;
+
+    // Append each row to the section on the page
+    dataSection.appendChild(periodWrapper);
+
+    // Push  each adjustedTotal to  adjustedTotals array for animating later
+    adjustedTotalArray.push(adjustedTotal);
+
+    // Now find each meter that was created above in theInnerHTML
+    let meterElement = periodWrapper.getElementsByTagName('meter');
```
rfultz, line 153: Any worries about picking up other <meter> elsewhere, like if we add this to another page or add other features to this page?

The variable is meterElement (singular) but `getElementsByTagName` will return all of them, right?

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+
+AcrossTime.prototype.handleYearChange = function(e) {
+
+     e.preventDefault();
+
+     // Set action (min or max) based on which select was changed.
+     let action = e.target.dataset.period;
+
+     // Determins which select was changed
+     let beginning = action == 'min' ? e.target.value : this.minYearControl.value;
```
rfultz, line 213: I'm curious what these two lines are doing—checking for an action of 'min' and using that element's value or otherwise grabbing the value from minYearControl? So, grab min's value if it was changed or if it wasn't changed?

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+
+  window
+    .fetch(buildUrl(this.basePath_officeTotal, query), {
+      cache: 'no-cache',
+      mode: 'cors'
+    })
+    .then(function(response) {
+      if (response.status !== 200)
+        throw new Error('The network rejected the grand total request.');
+      // else if (response.type == 'cors') throw new Error('CORS error');
```
rfultz, line 272: I think this is my fault. We should probably remove it

fec/fec/static/js/pages/contributions_across_time.js
```diff
@@ -0,0 +1,283 @@
...
+      response.json().then(data => {
+        instance.displayUpdatedData(data);
+      });
+    })
+    .catch(function() {
+      // TODO - handle catch
```
rfultz, line 278: I think this is my fault. We should probably remove it but that feels wrong. Thoughts?

fec/data/views.py
```diff
@@ -878,17 +886,31 @@ def house_senate_overview(request, office, cycle=None):
             reverse("elections-president", args=(cycle, ))
         )
 
+    office_codes = {
```
rfultz, line 889: Oh, I didn't think about us needing the reverse…
Can we make the quotes consistent?
Should we add P?

rfultz (changes requested):
Solid! A few tiny things, questions, and 👍🏼

fec/data/templates/house-senate-overview.jinja
patphongs: @johnnyporkchops, if you're going to rename this file, could you rename it to something more generic, maybe "elections-overview.js"? I'm hoping to also include the code for the all elections totals table within the same JS file, so we don't need to have separate files per feature.

JonellaCulmer:
Cannot view locally, so any tips are appreciated. The following review is based on your screenshot/gif.

For the current two-year period, the bar chart needs to have a different appearance to clarify and reinforce that the cycle is currently in progress. This detail is very important and helps to put the data into context when comparing to previously completed cycles. I consider this detail a blocker to completion. 

The speed with which the columns populate are a bit slow - can we match the speed of the bar charts in the [raising/spending summaries](https://www.fec.gov/data/raising-bythenumbers/). 

The kerning is tight on the section headers. 

I don't want to introduce a third bar chart style: the slightly rounded corners. We need to mimic the style of the bar charts in the raising and spending summaries (but larger), or in the Who's raising the most charts. I'd prefer the latter. But both can be found on the Raising and Spending by the numbers pages: https://www.fec.gov/data/raising-bythenumbers/

fec/data/templates/macros/cycle-select.jinja
johnnyporkchops: @rfultz I ended up not using this macros for the feature, so we can remove it.

fec/data/templates/macros/cycle-select.jinja
johnnyporkchops: @rfultz I ended up not using this macros for the feature, so we can remove it.

fec/fec/static/js/pages/contributions_across_time.js
johnnyporkchops: @rfultz Good question, I thought it was OK since its in the context of the periodWrapper element.

fec/fec/static/js/pages/contributions_across_time.js
johnnyporkchops: @rfultz , It determines which select was chosen for the math.min/max check so it can transpose min/max in call to data if year  of min select is  > year  of max select

fec/data/views.py
johnnyporkchops: @patphongs , @rfultz I  left it this way in case the answer to Pat's question is yes. 
The reverse is also still in the `def elections` view but I don't think we need it in both places now that we added the new `def elections_president`  view when we created `house_senate overview` templates.

fec/data/templates/house-senate-overview.jinja
johnnyporkchops: @pat, I will rename it as you suggested. I am wondering though if the other complicated  features on the page, (like Treemap) will be better off in their own  JS file so maybe eventually we can rename this one less generic, and more specific to the features it applies to (at this point it would be the across-time and the feature you are adding)

fec/fec/static/js/pages/contributions_across_time.js
johnnyporkchops: Can we leave this in here for now as a TODO?

fec/fec/static/js/pages/contributions_across_time.js
johnnyporkchops: @rfultz Can we leave this in here for now as a TODO?

fec/data/templates/house-senate-overview.jinja
```diff
@@ -36,21 +31,8 @@
...
-    district: '{{ district or '' }}'
-  }
-};
-
+var context = {{ context_vars | to_json | safe }};
 </script>
 <script src="{{ asset_for_js('dataviz-common.js') }}"></script>
-<script src="{{ asset_for_js('elections.js') }}"></script>
-<script src="{{ asset_for_js('committee-single.js') }}"></script>
+<script src="{{ asset_for_js('elections-overview.js') }}"></script>
```
rfultz, line 37: Excellent!

fec/data/views.py
rfultz: Does Python handle single and double quotes differently—any reason we're using both right next to each other?

rfultz (approved):
Inconsistent quotes (single and double in same file and on the same lines) but looks good for a merge not intended to be public (methodology work, etc)

JonellaCulmer:
@johnnyporkchops This is looking really great! The only thing I found, that I didn't see because I couldn't view it locally for my first review, is the wrapping of the "footnote"-type language. The button stacks underneath the text at smaller widths and I don't think we want that. For other charts with the footer language, there's a line break about halfway across the feature, and the buttons stack to the right. Here's an example: https://www.fec.gov/data/raising-bythenumbers/

patphongs:
Merging this in since the disclaimer text and methodology button will be taken care in #5131

PR #5241, opened 2022-05-23
Remove debts from Loans and debts page

JonellaCulmer:
## Summary (required)

- Resolves #5168 

Removes debts language from the loans and debts page, the main navigation and the browse data left navigation

### Required reviewers

One front-end

-  Browse data
-  Main navigation

## Screenshots
Screen Shot 2022-04-19 at 11 56 58 AM

## How to test

- The word "debt" does not appear on the browse data section, main navigation

### DIFF
--- a/fec/data/templates/browse-data.jinja
+++ b/fec/data/templates/browse-data.jinja
@@ -48,7 +48,7 @@
                 tabindex="0"
                 aria-controls="loans-debts"
                 href="#loans-debts"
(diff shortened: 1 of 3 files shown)


JonellaCulmer:
@patphongs Made the changes outlined in the ticket but we found a few other references to debts that we weren't sure we should change:
- jinja file name
- ID tag
- URL

PR #5415, opened 2022-09-19
[WIP] Django cron job logging

cnlucas:
I've changed the logs from stdout going to terminal mail to a log file. I illustrated that we can use the stdout OR logging and send to a file via two different jobs. Transition uses logging and web_app_pages uses the stdout. One might be better than the other depending on the tool we use to set up a slack bot once the backup is also automated. 

Setup and run cron job:
- checkout feature branch: `git checkout setup-django-cron-job`
- activate python virtual env: `pyenv virtualenv 3..9.13 venv-cron`
- install requirements: `pip install -r requirements.txt`
- install and build node dependencies: `npm i` & `npm run build`
- migrate database: `cd fec` and `./manage.py migrate`
- add scheduler job: `python manage.py crontab add` (cron job is set to run every one minute for testing purpose)
- verify cron job is up: `python manage.py crontab show`
- remove cron job: `python manage.py crontab remove`


fec/search/management/commands/scrape_transition_pages.py
```diff
@@ -21,7 +29,8 @@ def add_arguments(self, parser):
         )
 
     def handle(self, *args, **options):
-        self.stdout.write(self.style.WARNING('Getting pages...'))
+        #self.stdout.write(self.style.WARNING('Getting pages...'))
+        logger.info('Getting pages...')
```
patphongs, line 33: I would suggest adding some timestamps to the logs so that we can understand when each job ran

PR #5574, opened 2023-01-23
Add Type to footer button

### LINKED ISSUE #5573: Footer Button flagged by eslint
### What we’re after

The footer button is breaking eslint's rule about a button needing an explicit type: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.md

This is blocking us upgrading babel/es packages. When we upgrade eslint it will break builds due to the above rule. 
Screen Shot 2023-01-20 at 2 58 55 PM

Screen Shot 2023-01-20 at 3 01 16 PM 

### Action item(s)

(These are the smaller tasks that should happen in order to complete this work)
- [x] Add a type to the button
- [x]  Create ticket for babel/es upgrade

tmpayton:
## Summary (required)

- Resolves #5573 

This PR adds an explicit type to the footer button so we can upgrade babel and es packages. 

-  footer

## How to test

1. Navigate to the home page - http://127.0.0.1:8000/ 
2. Scroll down to the footer
3. Click the button that states "Sign up for FECMail"
4. Ensure that the button works still

Other links to try: 
- http://127.0.0.1:8000/calendar/?calendar_category_id=25&calendar_category_id=26&calendar_category_id=27 (base.html)
- http://127.0.0.1:8000/data/raising-bythenumbers/

### DIFF
--- a/fec/data/templates/layouts/main.jinja
+++ b/fec/data/templates/layouts/main.jinja
@@ -137,7 +137,7 @@
         Washington, DC 20463</p>
 
         <a href="https://public.govdelivery.com/accounts/USFEC/subscriber/topics?qsp=CODE_RED" target="_blank" rel="noopener">
-          <button class="button--standard button--envelope">Sign up for FECMail</button>
+          <button class="button--standard button--envelope" type="button">Sign up for FECMail</button>
         </a>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


johnnyporkchops (approved):
Buttons work as expected. Great catch @tmpayton !

PR #5601, opened 2023-02-06
Sort MUR by case_no

### LINKED ISSUE #5261: Case sorting by MUR number, with descending as default, adding ascending
**What we're after:**
MUR's are sortable in ascending and descending order

Look into javascript sorting as this is not datatables

**Related issues**
- https://github.com/fecgov/openFEC/issues/5152

**Completion criteria**
- [ ] Able to sort by MUR number in ascending and descending order

johnnyporkchops:
## Summary
Make MURS sortable by case number at:
https://www.fec.gov/data/legal/search/murs/ --OR--
https://www.fec.gov/data/legal/search/enforcement/  (defaults to MUR search)

- Resolves #5261

### Future work: 
This should work for AOs and possibly ADR/AFS. If we want to repurpose it for these, I recommend we remove the JS from the MURs template and revise the script to work for all of them and put it under /`js/pages` ,  `/js/modules` or `legal.js`. If that is possible, given the differences in those results-tables. The templates/views also would need to be edited like the MURs'  were for this PR.

### Required reviewers

One frontend, one backend, one UX

### Impacted areas of the application

modified:   fec/static/scss/components/_table-styles.scss
modified:   legal/templates/legal-search-results-murs.jinja
modified:   legal/templates/partials/legal-search-results-mur.jinja
modified:   legal/templates/partials/legal-pagination.jinja
modified:   legal/views.py

## Screenshots

mur_sort

## Related PRs

[#5246](https://github.com/fecgov/openFEC/pull/5246) Add sorting by case number and by advisory opinions number in legal/search

## How to test
- [ ] **Remove develop deploy task before merge**

- checkout `feature/5261-sort-murs-by-case-no` 
- `npm run build`
- `npm run test-single`
- `pytest` -AND- `cd ../ && pytest`
-  Go to http://127.0.0.1:8000/data/legal/search/murs/
- Test sort toggle on first column
- Apply some filters. Does it maintain the current sort order after applying filters?
- Test pagination with/without filters applied.
- Make sure the `aria` attributes on the sort toggle are correct upon changing sort direction
- Do we want the extra JS that maintains the position of the page when the page reloads (`see lines 115 and 120 of murs template` )? It has pros and cons -- Its great when using the toggle button, but maybe not as great when you are scrolled down to use the `Apply filters` button. 
     -  Also,  clicking pagination btns does not maintain page position because it loads the refreshed page to the `#results-murs` anchor. 
    - FYI... admittedly, the gif screenshot above implies that there is no visual page-jump on reload. While I was able to capture this ideal experience in a GIF for the PR, that's not always the experience since the page does have to reload to sort the data. :- {

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -74,7 +74,7 @@ def _detect_space(repo, branch=None, yes=False):
 DEPLOY_RULES = (
     ('prod', _detect_prod),
     ('stage', lambda _, branch: branch.startswith('release')),
-    ('dev', lambda _, branch: branch == 'feature/5261-sort-murs-by-case-no'),
+    ('dev', lambda _, branch: branch == 'develop'),
     # Uncomment below and adjust branch name to deploy desired feature branch to the feature space
     # ('feature', lambda _, branch: branch == '[BRANCH NAME]'),
 )
--- a/tasks.py
+++ b/tasks.py
@@ -74,7 +74,7 @@ def _detect_space(repo, branch=None, yes=False):
 DEPLOY_RULES = (
     ('prod', _detect_prod),
     ('stage', lambda _, branch: branch.startswith('release')),
-    ('dev', lambda _, branch: branch == 'develop'),
+    ('dev', lambda _, branch: branch == 'feature/5261-sort-murs-by-case-no'),
     # Uncomment below and adjust branch name to deploy desired feature branch to the feature space
     # ('feature', lambda _, branch: branch == '[BRANCH NAME]'),
 )
--- a/fec/fec/static/scss/components/_legal-search.scss
+++ b/fec/fec/static/scss/components/_legal-search.scss
@@ -119,3 +119,32 @@
   margin: u(2rem 0 0 0);
   padding: u(1rem);
 }
+
+.simple-table__header-cell { 
+
+    &.sorting,
+    &.sorting_asc,
+    &.sorting_desc {
+      background-position: calc(100% - 1rem) 50%;
+      background-repeat: no-repeat;
+      cursor: pointer;
+      padding-right: u(3rem);
+    }
+
+    &.sorting {
+      @include u-icon-bg($dash, $gray);
+      background-size: u(1rem);
+    }
+
+    &.sorting_asc {
+      @include u-icon-bg($arrow-up-border, $primary);
+      background-size: u(1rem);
+      background-color: $gray-lightest;
+    }
+
+    &.sorting_desc {
+      @include u-icon-bg($arrow-down-border, $primary);
+      background-size: u(1rem);
+      background-color: $gray-lightest;
+    }
+}
--- a/fec/fec/static/scss/components/_legal-search.scss
+++ b/fec/fec/static/scss/components/_legal-search.scss
@@ -119,32 +119,3 @@
   margin: u(2rem 0 0 0);
   padding: u(1rem);
 }
-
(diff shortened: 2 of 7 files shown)


patphongs:
> If we want to repurpose it for these, I recommend we remove the JS from the MURs template and revise the script to work for all of them and put it under /`js/pages` or `/js/modules`

@johnnyporkchops You read my mind here. I'm pretty sure that the ADRs and AFs are developed in the same way as MURs. So it would be nice to repurpose the sort capability in those scenarios. Both use `case_no` as an identifier too. Can you try to put the script where we can easily grab it if we need it for those other legal docs?

johnnyporkchops:
> @ patphongs wrote: 
Can you try to put the script where we can easily grab it if we need it for those other legal docs

@patphongs, Good question. The script is inside the murs jinja template, so it would be easy to move to a separate script. Since the AF/ADR templates and views would also need to be edited, I was inclined to keep it in the murs template until then because those changes may inform how the JS gets revised to work for all the targeted legal docs. It could be its own file, or maybe  added to `legal.js` which is already included.

pkfec:
@johnnyporkchops deployed your changes to `Dev` space and observed the following issue with pagination

1. Sort by case_no ascending. Archived murs are shown, which is the expected behaviour. Click on next page. The next page results are showing current murs in descending order. **This is incorrect.** 

2. Sort by case_no descending. Current murs are shown, which is the expected behaviour. Click on next page. The next page results are showing currents murs in descending order. **This is correct.**

johnnyporkchops:
> @johnnyporkchops deployed your changes to `Dev` space and observed the following issue with pagination

@pkfec Good catch! This should be fixed now. Can you pull and redeploy to dev?

pkfec:
> data/legal/search/enforcement/

@johnnyporkchops [Redeployed](https://app.circleci.com/pipelines/github/fecgov/fec-cms/2440/workflows/57b01e62-280e-4367-a2d9-aa4967a9b882/jobs/5793) your latest changes to Dev space this afternoon. The sort works as expected. Great job fixing the pagination bug. I am now able to paginate the results in both asc and desc order.

Sorting by case_no works as expected!

pkfec (approved):

JonellaCulmer (approved):
Looks/works great from my perspective! Thanks @johnnyporkchops

patphongs:
> Since the AF/ADR templates and views would also need to be edited, I was inclined to keep it in the murs template until then because those changes may inform how the JS gets revised to work for all the targeted legal docs (especially if we include AOs which use `ao_no`)

Thanks for the additional clarification @johnnyporkchops. Keeping it in just the MUR template for now makes sense to me then.

tasks.py
```diff
@@ -74,7 +74,7 @@ def _detect_space(repo, branch=None, yes=False):
 DEPLOY_RULES = (
     ('prod', _detect_prod),
     ('stage', lambda _, branch: branch.startswith('release')),
-    ('dev', lambda _, branch: branch == 'develop'),
+    ('dev', lambda _, branch: branch == 'feature/5261-sort-murs-by-case-no'),
```
patphongs, line 77: @johnnyporkchops Everything looks good, last change is can you please put this back to `develop`? After this change, I can merge.

patphongs (changes requested):

patphongs (approved):
👍

PR #5637, opened 2023-03-06
Order snippets descending by most recently created

### LINKED ISSUE #5608: Wagtail wish list: Ability to put snippets in order so the most used are on page 1.
### Summary

**What we're after:**
As a Wagtail user who needs a snippet, it would be good to not have to go to the last page for the one I need.

Right now the snippets are in the order they were created. That means some of the most used ones (contribution limits for 2023-2024) and COVID snippet) are on the last page.

Background information: https://stackoverflow.com/questions/63472227/wagtail-how-to-order-snippets

### Completion criteria

- [x] Snippets for 2023-2024 contribution limits, the COVID snippet and the coordinated party expenditures archive appear on page 1 of the snippet list.

johnnyporkchops:
## Summary 
Content team wants to see the most recently created snippets on page 1

- Resolves #5608

**Possible future work:**
Wagtail 4 will have a ton of new ways to handle Snippets, including adding sortability in the admin panel, although the [implementation ](https://docs.wagtail.org/en/stable/topics/snippets.html#customising-snippets-admin-views)looks a bit complex at first glance of the docs.

### Required reviewers

one person (frontend or content)

	new file:   home/migrations/0125_alter_embedsnippet_options.py
	modified:   home/models.py

### Page one of Snippets
Screen Shot 2023-03-05 at 7 42 31 PM

## How to test
**Content team:** This is pushed to dev space for testing: https://dev.fec.gov/
**Developers:**
- checkout and run branch
- Go to Wagtail admin > Snippets
- Confirm that snippets are ordered newest to oldest
- You can  view the snippet ID on the table-row in the inspector,  `<tr id="snippet-row-29">`


dorothyyeager (commented):
Thanks for this @johnnyporkchops - the one thing though is we wanted them in the order of **most used**. In most cases this is the most recently created, but there's one snippet at the very end of the last page - "Archive of past years' coordinated party expenditure limits" - that we were hoping could slide onto page 1 too.

Basically what we need up front;
1. COVID snippet
2. Contribution limit table for the most recent cycle
3. Contribution limit footnotes for the most recent cycle
4. Archive of coordinated party expenditure limits

After those four, the order doesn't really matter and the way you have them is fine.

Is it possible to do this? (Perhaps by putting them in order of most recently updated?)
One more thought - after May, it's likely we won't need the COVID snippet much anymore. So if it can be done in the order of "most recently updated" then that might work the best for content.

johnnyporkchops:
@dorothyyeager There is not option to ordeer by `updated-date` like you could for  a Page . The only fields in Snippets to order by  are:
 `ID` (number)(which is what I used for ordering by last-created)
 `Title` (alpha)
 `Description` (alpha)
 `Text` (alpha)
 
 - So you could come up wtih a convenntion for prepending a number or alpha letter  to the `Title` field and  then I could order by that field (that field does not show on the page when its  published)
 - There is also ability to add Tagging to Snippets like images and documents and you could come up with a tagging connvention that I might(might) be able to order by.
 
But I think you'll agree that alll  of that is too complicated to get into just for ordering them in the admin.

 I can either push this PR   as is if you  think it helps or we can wait  for Wagtail 4 where there is sorting capability built-in and  
 other configurable options.

PR #5669, opened 2023-03-21
fix dependency issues in migrations

johnnyporkchops:
## Summary (required)

- Resolve migration dependency issue with two migrations depending on the same migration

### Required reviewers
One:


cnlucas (approved):
Deployed to dev to test and succeeded.

PR #5798, opened 2023-07-13
Fix for reporting dates with no header/foot notes

johnnyporkchops:
## Summary (required)

Certain Reporting dates pages don't use header notes, like this one: https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/2022-reporting-dates/24-and-48-hour-reports-independent-expenditures-periods-congressional-primaries-2022/
This broke the new JS logic, making zebra-striping and state dropdown filter not work. So a fix is put in to handle cases where there are no Header notes or Footnotes.

### Required reviewers
on frontend or content

## How to test
Previous to this change, lack of a Header note section on the template would break JS, this fixes that.
- Checkout branch and run `npm run build`
- Create a new ReportingDatesTable page 
- Create a Dates table, remove the last two columns fo there are four columns total
- Copy/paste cells A1 through D321 from this google sheet into the table:
https://docs.google.com/spreadsheets/d/1gvJ1_Dzdk7dbrIgwGZDgVPVABlCOOElT/edit#gid=1971065264
- Don't create any Footnote sections, preview page (state dropdown filter and zebra striping should still work)
- Create only a Footnote section, but no Header notes section, preview the page  (state dropdown filter and zebra striping should still work)
Here is an example IE page using old JS and template with only Footnotes:
https://www.fec.gov/help-candidates-and-committees/dates-and-deadlines/2022-reporting-dates/24-and-48-hour-reports-independent-expenditures-periods-congressional-primaries-2022/


djgarr:
The updates are working properly. Thanks

PR #5931, opened 2023-09-28
[MERGE BEFORE 09/28/23 RELEASE] Add link to document page

johnnyporkchops:
## Summary (required)

- Resolves #5918

`DocumentPages`, otherwise not navigable on the site, are showing in search results. This is a temporary fix so users will be able to get to the document if they access a DocumentPage through a search result link.  A final solution for handling this issue is being researchd. See this [comment/mockup in issue](https://github.com/fecgov/fec-cms/issues/5918#issuecomment-1738106863)

### Required reviewers

one frontend

## Impacted areas of the application

Document pages in Reports about the FEC

## Screenshots

Screen Shot 2023-09-27 at 9 47 40 PM

## Related PRs

Related PRs against other branches:

There is a WIP/DRAFT PR that shows more options for the final solution https://github.com/fecgov/fec-cms/pull/5922

## How to test

- checkout and run branch
- You can login to Wagtail and navigate from the sidebar `Pages > Home > About > Reports about the FEC`, then go into one of the sections like `Agency Operations or FOiA` 
- Click the LIVE button  in the Status column to view these pages live
- Example page from screenshot above: http://127.0.0.1:8000/about/reports-about-fec/agency-operations/federal-employee-viewpoint-survey-small-agency-management-report-2020/
- You can't navigate to them on the site, but they are showing up in search results since they are Wagtail pages, so this is a stopgap measure to ensure users can get to the document if they access the page through a search result link.

PR #5957, opened 2023-10-14
Fix server error on some reports

johnnyporkchops:
## Summary
The new logic that shows file extension on `DocumentPages(Reports about the fec)` was  causing some report pages to get a 500 server error with `index out of range` error. Since `self.file_extension` is a property on the model and not a field value it always returns "true" [Example 500 error](https://www.fec.gov/about/reports-about-fec/oig-reports/audit-agency-controls-governing-process-procurement-vendor-training-services-executive-summary/).

Screen Shot 2023-10-13 at 9 25 19 PM

This logic was added in https://github.com/fecgov/fec-cms/pull/5931

- Resolves #issue_number

### Required reviewers
One frontend

Reports about the FEC pages that do not have a document associated with them like :
https://www.fec.gov/about/reports-about-fec/oig-reports/audit-agency-controls-governing-process-procurement-vendor-training-services-executive-summary/

modified:   home/templates/home/document_page.html

## How to test

- checkout and run branch
- Make sure  reports in the list that that have `Read more...` links on this page do not get 500 Server error
    - http://127.0.0.1:8000/about/reports-about-fec/oig-reports/?category=audit+report
    - These are reports using DocumentPage that do not reference a file, the page itself is the report.
- @rfultz , I also went ahead and conditionally removed  the paragraph tag with file link and info on pages that do not have a document per [your earlier suggestion](https://github.com/fecgov/fec-cms/pull/5931/files#r1340239967)

### DIFF
--- a/fec/home/templates/home/document_page.html
+++ b/fec/home/templates/home/document_page.html
@@ -27,7 +27,7 @@
     </header>
     {% if self.file_url %}
     <p class="t-sans t-normal">
-      <i class="icon icon--inline--right icon--inline--left i-document"></i><a href="{{ self.file_url }}">{{ self.title}}</a>
+      <i class="icon icon--inline--right icon--inline--left i-document"></i><a href="{{ self.file_url }}">{{ self.title }}</a>
       {% if self.extension is not '' %} | ({{ self.extension }}){% endif %}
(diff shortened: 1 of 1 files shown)


rfultz (approved):
Can't seem to break it! :)

PR #6189, opened 2024-04-05
Single select for line-number filter on Receipts/Disbursements

### LINKED ISSUE #6073: Implement new dropdown filter on receipts and disbursements datatables
**What we're after:**
Change out the filter been using for the line number filter on the receipts and disbursements datatables. 

**Datatable links:**
All receipts: https://www.fec.gov/data/receipts/
Disbursements: https://www.fec.gov/data/disbursements/ 

**Single-select pattern in pattern library:** 
https://fec-pattern-library.app.cloud.gov/components/detail/single-select.html

**Completion criteria:**
- [ ] Line filter now functions as a single-select option

**Designs:**
**Filter tag before:**
Screen Shot 2023-12-27 at 2 54 33 PM

**Filter tag after:**
Screen Shot 2024-01-05 at 10 07 11 AM

**Dropdown without and with selection:**
Screen Shot 2023-12-27 at 3 17 20 PM

**Dropdown (open):**
Screen Shot 2023-12-27 at 3 07 49 PM

**Full dropdown text for receipts (moved line number and Form # added to section rows):**
Screen Shot 2023-12-27 at 3 15 07 PM

**Full dropdown text for disbursements (moved line number and Form # added to section rows):**
Screen Shot 2023-12-27 at 3 16 15 PM

johnnyporkchops:
## Summary (required)

- Resolves #6073

- Created a new macro for single select
- Changed constants for new syntaz for labels and option-group headings
- Edited `select-filter.js` to allow a removeable tag
- Edited `filter-tags js`, `removeable_tag` template to allow aditional text
- Edited `recipts/disbursements-filter.jinja` to remove replace checkbox dropdowns with. a single select for `line number`

### Required reviewers

One frontend, One UX

Impacts receipts and disbursements datatables' line-number filter

modified:   data/constants.py
modified:   data/templates/macros/filters/dropdown-json.jinja
modified:   data/templates/partials/disbursements-filter.jinja
modified:   data/templates/partials/receipts-filter.jinja
modified:   fec/static/js/modules/filters/filter-set.js
modified:   fec/static/js/modules/filters/filter-tags.js
modified:   fec/static/js/modules/filters/select-filter.js

 

## Screenshots

Coming soon...

## How to test

Coming soon...


JonellaCulmer:
@johnnyporkchops Thanks for your work on this. A few thing I noticed. 
1. Headers saying "Received by... (Form #)" or "Made by...(Form #)" should be in bold, not grayed out. Refer to the PACs dropdown filter on either table for reference and look to the formatting for "Separate segregated funds".
2. Under each header for "Receipt type" and "Disbursement type" it should say "_Limit 1 receipt type_" or "_Limit 1 disbursement type_"
3. There's a white border around the dropdown field when you click on it that should not be there. 
4. The hover state does not match other dropdown fields. The fill/background changes instead of the border.

johnnyporkchops:
@JonellaCulmer 
- I can add the "Limit 1 receipt type" or "Limit 1 disbursement type"
- Since this is not one of our checkbox (multiple-select) dropdowns, it uses a single-select form element. 
     - The hover state matches that of our other single selects like here: https://www.fec.gov/data/committee/C00010603/ or: https://www.fec.gov/legal-resources/enforcement/audit-search/ 
- I do not see the white border, I do see a blue border which matches the global styles for the select form element.
- There is honestly no way to change the browsers default for form elements, specifically the [optgroup](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup)  label for a select. 
    - Except for completely eplacing the select with javascript which I do not recommend and seems like overkill just to override the browsers native default for the section headings. 
    - One other option would be to leave a blank line there, to signify sections, whithout having a grayed-out label
      
Screenshot 2024-04-08 at 11 44 54 AM

JonellaCulmer:
@johnnyporkchops Thanks for the explanations and adding the missing text. How are we able to use the bold headers on the PAC dropdown? Is it because that's a multiselect? 

Thanks for flagging the different hover states, I'm going to put in a ticket to evaluate why we have that discrepancy. Also, I appreciate the other single-select examples you provided because those examples also show the white border I mentioned as well. Seems that might be a single-select issue? Here's a screenshot and I can put in a followup ticket to fix that as well. 
Screenshot 2024-04-08 at 1 18 06 PM

johnnyporkchops:
> How are we able to use the bold headers on the PAC dropdown? Is it because that's a multiselect?

@JonellaCulmer Those checkbox dropdowns are not `html select elements` (they are actually not even `html multi-select` ielements. They are basically html divs and list items that wrap `html checkboxes` to achieve the whole checkbox multi-select interaction including where it shows what is selected above and you can then uncheck/check  it there as well. So the real form element there is checkboxes. 

It looks like the  white border is a user-agent style that sets an "outline' property ( not the more common border property). It could be overridden, but I think we should consider accessibility concerns as part of  that ticket as well.

Screenshot 2024-04-08 at 1 38 25 PM

JonellaCulmer:
Here are the followup tickets I mentioned: 
- https://github.com/fecgov/fec-cms/issues/6191
- https://github.com/fecgov/fec-cms/issues/6192

JonellaCulmer (approved):
Approved, thanks for your work on this @johnnyporkchops.

patphongs (approved):
Looks good, thanks @johnnyporkchops!

PR #6214, opened 2024-04-18
Expand keyword search on /updates to include html content

johnnyporkchops:
## Summary (required)

- Resolves #6118
- Added `search_updates()` function to convert the body (on pages with html fields) to a string and do a Python string search on them
- Preserve superior existing backend Postgres search for pages with searchable fields types (non-html)
- Exclude duplicate search results  on rare occurrences where both search options return results for the same query because  a page has both html and other database-searchable fields

### Required reviewers
one frontend AND one UX or content

## Impacted areas of the application
The `/updates` keyword search
modified:   home/views.py

## Screenshots

Coming soon...

## How to test
This has been pushed to dev for testing, check with me to confirm its still live if you are testing there.

- checkout and run branch
- do a keyword search for a term that appears in an html block of a Digest, PressRelease, Record, or TipsForTreasurers page
    - Example search from issue: https://www.fec.gov/updates/?update_type=weekly-digest&search=MUR+6535
- do a keyword search for a term not in an html block of a Digest, PressRelease, Record, or TipsForTreasurers page

**Note:** The html search is a little slower when  searching across all updates (`Publication type`  set to `All` ) , but I am working in innovation sprint to optimize performance for a quicker return time.

Issue #6118, opened 2024-02-12
MUR cannot be found through Latest updates search


JonellaCulmer:
**What we're after:**
A keyword search for "MUR 6535" yields no results through the Latest updates search. Other keywords return results as expected. A result for a weekly digest should come up: **Weekly Digest for the week of December 14-18, 2015**

Link: https://www.fec.gov/updates/?update_type=weekly-digest&search=MUR+6535

**Completion criteria:**
- [ ] Fix has been identified and implemented, MUR 6535 is successfully found through Latest updates keyword search

patphongs:
After researching the issue, it appears that the items the news and meeting items that were just imported HTML into Wagtail are not being searched. These include imported HTML pages for record articles, press releases, weekly digests, and commission meetings. We need to figure out a way to have the Wagtail search feature also search those HTML content types. These are older news and meeting items from the classic site.

johnnyporkchops:
According to this older post, `index.SearchField('body')` will not search the body in a queryset for a subclassed page, only in a queryset for the parent class. `DigestPage` is a subclass of `ContentPage`. Our queryset for /Updates  is  `digests = DigestPage.objects.live()`.  I could not find any update or change to this in more recent version docs.
https://stackoverflow.com/a/63917472/5616606

johnnyporkchops:
This ticket is for the /Updates search, but for followup work we can resolve this for meetings too. Unlike the updates, we could probably target the alternate search by meeting-date. The last open MeetingPage that has `imported_html` is April 27, 2017.

PR #6230, opened 2024-04-29
Faster keyword search for updates and html search for legacy meetings

johnnyporkchops:
## Summary (required)

- Resolves #
- Related issue:  #6118 
- Related PR: https://github.com/fecgov/fec-cms/pull/6214

- The [original PR](https://github.com/fecgov/fec-cms/pull/6214) for resolving un-searchable html in `/updates` searches worked, but the search was noticeably slow when searching across all update-types. This PR speeds that search up.
- This PR also combines  a more inclusive text-search with the existing  `wagtail.search.backends.database` search for  `Open meetiings` and `Hearings` which will find more results for all years in addition to resolving the issue of previously unsearchable legacy pages with `imported_html` content.

### Required reviewers

One dev
One UX or One Content

-  `/updates` searches (PressRelease, WeeklyDigest, TipsForTreasurers and Record) 
-  Meetings page searches on `/meetings`

	modified:   home/views.py

## How to test
**Content/UX**:
Ask me to push to  featiure or dev and see bullets below for testing 
**Developers:**
- checkout and run branch

- Go to `/updates` and search for "MUR 6535" with "All" chosen in `Publication type`. This should be much faster than the same search on Production.
- Go to `/meetings` and search "Lorenzo Holloway". You should get more results than on Production, with results going back past 2016.
- Try other keyword searches for pages older that 2016 for both `/updates` and `/meetings`.

**Note:**:  Prior to this PR, a common pattern for listing Presenters separated by a slash in meeting pages, would not return results when searching individual names if there was not a space before and after the slash. For example, The built in db search will not find Lorenzo Holloway in `Joshua Blume/Lorenzo Holloway` but it will find it in `Joshua Blume / Lorenzo Holloway`.  With the changes in this PR,  searches will return all pages with the presenter's name, regardless of the syntax.

### DIFF
--- a/fec/home/views.py
+++ b/fec/home/views.py
@@ -348,7 +348,7 @@ def index_meetings(request):
     executive_years = list(
         map(lambda x: x.year, executive_sessions.dates("date", "year", order="DESC"))
     )
-    
+
     # These clear the search field upon navigating tabs as requested by content team
     meetings_query = ""
     hearings_query = ""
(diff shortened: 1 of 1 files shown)


kathycarothers (approved):
I tested keywords for the meetings and updates and my search proved results.

rfultz (approved):
Looks great—I can't break it!

PR #6276, opened 2024-05-20
Error handling for about/careers (USAJobs API)

### LINKED ISSUE #6260: 500 Server Error for Careers page  - USA Jobs API
**Thank you for your interest in helping improve the FEC CMS.**

Below, you'll find questions to help us more quickly identify, prioritize, and then address the issue.

### Summary

_What happened?_
`/about/careers`  page shows a `500 Internal Server Error` when USAJobs is down for scheduled maintenance. This is because the error handling in the `open_jobs.py -- get_jobs()` templatetag only works when the status code for the response is NOT 200, which is not the case for the scheduled maintenance message.

_What should have happened?_
The `about/careers ` page should show the `USA_JOBS_ERROR` in the open positions section instead of failing to load with a server error.

### Frequency

This will happen whenever USA Jobs returns a `Down-for-maintenance error`  from calls to the API that our `open_jobs.py -- get_jobs()` makes.

### How to Reproduce
- This is hard to reproduce because the USAJobs API maintenance window is over and the API is back up and running.
- Attempt to return a json response mocking the  maintenance message that  USAJobs returns from calls to the API during this maintenance window. 
- Contact USAJobs API support to clarify that the response from a call during a maintenance window is a 200 json response and not a 404 or something else

### Device, browser, etc

_Tell us everything you can about:_

All OS's , all browsers

### Screenshots

_Screenshots would help us correctly identify the issue._

### Techical considerations

 - [ ] Research to subscribe to usajobs RSS feed into slack (Like we did for Git, Circleci)

johnnyporkchops:
## Summary (required)

- Resolves #6260
- Expand the error handling for ConnectionErrors not caught by the current check for 200 http response,
- Expand the error handling to catch errors that have a 200 http response, but can still break the page.
    - When the request returns html or json data that is not expected jobs dict which which previously caused a `500 Internal server error`
- Catch connection errors if USAJobs API is not responding or down for maintenance 
- Write the error class name to the page to give devs an indication of what type of error is happening

### Required reviewers

One python developer

`/about/careers/`
`open_jobs.py - get_jobs()` templatetag

## How to test

- Checkout and run branch
- Export the following  env vars from CF in your shell :
    - USAJOBS_API_KEY
    - USAJOBS_WHOMAYAPPLY
    - USAJOBS_AGENCY_CODE (since we currently have no open jobs, you can use `PU00` for `Peace Corps` if you want to confirm job data renders to the page)
 - view the page at `/about/careers/`
 - run `pytest` from /fec-cms and /fec-cms/fec directories
 
 Below are a few ways to trigger various error scenarios if you are interested in testing them:
 - Try changing the JOB_URL in `open_jobs.py` to a non existent website. Should get (ConnectionError) at the end of the error message in the page.
 - Try adding `querystring["page"] = 2` to the querystring parameters, should see: "There are currently no open positions..." message (since it shows a max of 25 jobs per page, it will return no jobs)
  - Comment out `"host": "data.usajobs.gov"` (@ line 36) and then change the JOB_URL to `"https://dummyjson.com/products"`, should get a  (KeyError)


johnnyporkchops:
@JonellaCulmer . Do you think we could come up with different wording for the first sentence in the usajobs error message `"USAJobs is unavailable"`. After digging into this more for this ticket, I realized that there could be a scenario where it is indeed unavailable or "down", but it could also be that just the API is down or not returning the data we need to post jobs. So rather than saying `"USAJobs is unavailable"`, could we say something more generic like `"Something went wrong."`? That way we are not essentially saying `"USAJobs is unavailable....Please visit usajobs.gov..."` Here is the current error message:
"USAJOBS is unavailable. We are temporarily unable to post open FEC positions. Please visit [usajobs.gov](https://www.usajobs.gov/) for more information."

cnlucas (approved):
Works for me, thanks @johnnyporkchops

PR #6472, opened 2024-09-11
Outdated browsers banner - fix the homepage

rfultz:
## Summary

- Resolves #issue_number

Removed the temp test

### Required reviewers

- one?

Only the homepage


patphongs (approved):
Thanks @rfultz for fixing it so quickly!

PR #6542, opened 2024-11-05
Handle URL for archived MUR 3620

### LINKED ISSUE #6508: Handle archive MUR 3620 exception
### Summary

**What we're after:**
MUR 3620 has the same number assigned to both its current and archived versions, which is causing an issue on the individual MUR page. In this case, only the first MUR from the objects list is loaded, regardless of whether it's the current or archived version. This is a rare occurrence and only affects this specific MUR, so we will address it by passing a `mur_type=archived` query parameter in the URL string whenever the archived case is accessed. This will ensure that the correct version is displayed, with the archived MUR being loaded when the parameter is present and the current version when it is not.

Image

### Related issues

_List any relevant related issue(s)_

- https://github.com/fecgov/openFEC/issues/5982

### Completion criteria

- [ ] Individual Archived MUR 3620 renders the archived MUR data
- [ ] Individual MUR 3620 renders the current MUR data

patphongs:
## Summary (required)

- Resolves #6508 

Archived MUR 3620 has a rare case where it was reopened without creating a different MUR number. Therefore, it will have 2 separate MUR cases attached to it, one is the current case and the other is the archived case. This PR will handle the particular URL depending on which case you are interested in seeing.

### Required reviewers

1 engineer

-  MUR 3620

## Screenshots

Screenshot 2024-11-04 at 9 12 20 PM

## How to test

- Go to http://localhost:8000/data/legal/search/murs/?search_type=murs&case_no=3620 and see the results for MUR number 3620. There should be 2, one archived case, and one current case.
   - Archived case should have this URL: http://localhost:8000/data/legal/matter-under-review/3620?mur_type=archived
   - Current case should have this URL: http://localhost:8000/data/legal/matter-under-review/3620
- Make sure that this does not effect other archived MUR or current MUR cases. Go to a random archived MUR case or random current MUR case to test.

### DIFF
--- a/fec/data/api_caller.py
+++ b/fec/data/api_caller.py
@@ -112,15 +112,19 @@ def load_legal_advisory_opinion(ao_no):
     return ao
 
 
-def load_legal_mur(mur_no):
+def load_legal_mur(mur_no, requested_mur_type='current'):
 
     url = "/legal/docs/murs/"
-    mur = _call_api(url, parse.quote(mur_no))
+    murs = _call_api(url, parse.quote(mur_no))
 
-    if not mur:
+    if not murs:
         raise Http404
 
-    mur = mur["docs"][0]
+    # get matching doc by mur_type, else use the first
+    if constants.ARCHIVED_MUR_EXCEPTION == mur_no:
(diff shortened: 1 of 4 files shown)


fec-jli (approved):
Test locally, it works as expected.
Good job. Thanks

PR #6565, opened 2024-11-20
6564 - Fix canonical rel links

### LINKED ISSUE #6564: Fix canonical meta data
### Summary

**What we're after:**

Fix our canonical url links to improve accuracy and search engine crawling, and clean up site reports.

`<link rel="canonical">` tags tell search engines the preferred/single URLs for pages. For our use, it would be removing query params so all the various `…?var=val&var=val&var=val` urls would be seen (by search engines) as the URL before the `?`, i.e. the page path.

`/data/*` should not show its canonical page as `/`
Image

### Related issues

None

### Completion criteria

Our `<link rel="canonical">` look more like individual URLs instead of simply `<link rel="canonical" href="https://www.fec.gov/">` site-wide.

- [Google: What is canonicalization?](https://developers.google.com/search/docs/crawling-indexing/canonicalization)
- Are there pages where we'd want a canonical href to continue past the page path?
- Are there pages that we'd like to assign special, one-off canonical href values?
- Should we remove `CANONICAL_BASE` from our various global/window front-end values? i.e. if it's not being used, is there any reason to keep it?

### Future work

?

rfultz:
## Summary

- Resolves #6564 

(Include a summary of proposed changes and connect issue below)

### Required reviewers

- code review

- Anywhere `CANONICAL_BASE` was being used on the front-end but I couldn't find it in the repo
- The `<link rel="canonical">` href values for every jinja template, so everything under `/data/`—everything not in Wagtail

## Screenshots

image

Removing `CANONICAL_BASE` from front-end:
image

## Related PRs

None

## How to test

- pull the branch
- `npm run build`
- `./manage.py runserver`
- Review the `<link rel="canonical">` `href` values for
  - homepage
  - some Wagtail pages
  - several Jinja pages
  - several Jinja pages with query params (that shouldn't be included in the canonical tag)
- Are there places where `CANONICAL_BASE` was being used by the front-end?

### DIFF
--- a/fec/data/templates/layouts/main.jinja
+++ b/fec/data/templates/layouts/main.jinja
@@ -35,7 +35,6 @@
     END_YEAR = {{ constants.END_YEAR }};
     DISTRICT_MAP_CUTOFF = '{{ constants.DISTRICT_MAP_CUTOFF }}';
     WEBMANAGER_EMAIL = '{{ WEBMANAGER_EMAIL }}';
-    CANONICAL_BASE = '{{ CANONICAL_BASE }}';
     TRANSITION_URL = '{{ TRANSITION_URL }}';
     {% if election_year %}ELECTION_YEAR = '{{ election_year }}';{% endif %}
 
--- a/fec/data/templates/partials/meta-tags.jinja
+++ b/fec/data/templates/partials/meta-tags.jinja
@@ -30,7 +30,7 @@
 
   {# Google
   ================================================== -#}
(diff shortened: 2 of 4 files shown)


johnnyporkchops (commented):
@rfultz We are using `settings.CANONICAL_BASE` in the search views, if we are not goingt o worry about this anymore than we should remove it here:
https://github.com/fecgov/fec-cms/blob/74ad4391497a85fefaecd912d4e3b3661d270352/fec/search/views.py#L63

johnnyporkchops (approved):
Looks good

PR #6569, opened 2024-11-25
5225 - Show HREF on Print-outs

### LINKED ISSUE #5225: Webpage print includes full URL next to hyperlink
### Summary

_As a user, I want to be able to print web pages and still be able to retain the URLs for hyperlinked text so that I am still able to know where the link is going._

### Completion criteria

- [ ] TBD

rfultz:
## Summary

- Resolves #5225 

Adding style rules to print href values when printing pages. Any affected elements will get their border-bottom removed, will get an underline (printing borders can be tricky) and will get its `href` value appended after it in the body of the page, inside brackets: `\00a0[href]` (`\00a0` is a space)

Then the rules are to include a link's href value if it's one of
- `#main a.button` (main content, any `<a>` with the `button` class)
- `#main h2 a` (main content, any `<a>` inside of `<h2>`)
- `#main h3 a` (main content, any `<a>` inside of `<h3>`)
- `#main li > a:first-child` (main content, `<li><a>` (first child directly inside the `<li>`))
- `#main p a` (main content, any `<a>` inside any `<p>`)
- `#main .message a` (main content, any `<a>` inside any `.message`)
- `.footer-links a` (the site footer links)
- any element with the new `print-url` class (in case we want to force showing it someplace

These will cancel the above selectors:
- an `href` starting with `#`
- if the new `no-print-url` class has been added to the element

### Required reviewers

- Front-end
- UX

Any changes will only take effect for print media so nothing should appear on the screen

## Screenshots

About page, actual print preview
image

H4CC
image

Footer links
image

## Related PRs

None

## How to test

- pull the branch
- `npm run build-sass`
- `./manage.py runserver`
- Load a page ([About](http://127.0.0.1:8000/about/) is a decent sample)
- Look at the print preview
- Check several other types of pages
- To make things quicker and check specifically this work:
   1. edit fec/fec/static/scss/elements/_elements.scss
   1. near the bottom, close the `@media print {` before the `// Rules` line
   1. at the very bottom, comment out or remove the very last `}`
   1. `npm run build-sass`
   1. reload pages and browse that way—the various `href` should appear on the screen normally in the same places they would if they were printed (though the rest of the layout will not look like printed versions might)

### DIFF
--- a/fec/fec/static/scss/components/_datatables.scss
+++ b/fec/fec/static/scss/components/_datatables.scss
@@ -187,6 +187,12 @@
   .value-bar {
     margin-top: u(0.5rem);
     height: u(0.5rem);
+
+    @media print {
+      color-adjust: exact;
+      print-color-adjust: exact;
+      -webkit-print-color-adjust: exact; 
+    }
   }
 
   @include media($lg) {
--- a/fec/fec/static/scss/components/_site-header.scss
+++ b/fec/fec/static/scss/components/_site-header.scss
@@ -555,6 +555,7 @@
 
   .masthead {
     background: none;
+    border-bottom: none;
   }
 
   .site-title {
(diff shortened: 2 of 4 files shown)


patphongs (changes requested):
I'd like to get this for the legal tables too

patphongs (approved):
Changes look good @rfultz, longer text for the legal tables are cut off when printing, but that's a separate bug that I will log. This bug isn't due to this PR.

PR #6702, opened 2025-03-05
2808 - Don't remove side padding for smallest $lg widths

### LINKED ISSUE #2808: Add padding/margins site-wide for medium widths
From widths 860px to 953px the site has no padding on the left or right—all content is right up against the edges of the window/viewport. We start to get single pixels of padding as we increase from there, based on how wide the window is over 953 (954 gets 1/2 pixel on each side, 955 has one on each, 957 has two...)

rfultz:
## Summary

- Resolves #2808 

Visual tweak. We were removing the side padding for main content columns as soon as we got into the $lg range but that meant content was against the edge of the browser window. This change adds a media query for that column so the content in full-width .container is never closer to the edges than the padding shared among the other widths

### Required reviewers

- 1 dev for code & UX?
- 1 UX?

The smallest widths of the $lg range for the `.container` selector

## Screenshots

Localhost/this PR above; www below
image

Localhost/this PR above; www below
image

Localhost/this PR above; www below
image

## Related PRs

None

## How to test

- Pull the branch
- `npm run build`
- `./manage.py runserver`
- Load the site in localhost
- Open inspector to more easily see the page widths. We're looking at widths 850px to 1000px (10px before the new media query starts and 10px after it ends)
- Check pages that use the .container class
  - The content in the top red or blue section for pages like [About](http://127.0.0.1:8000/about/) or [Legal Resources](http://127.0.0.1:8000/legal-resources/)
  - The page title and text directly under it for pages like [Mission and History](http://127.0.0.1:8000/about/mission-and-history/)
  - The main page content as a whole for pages like [H4CC Guides](http://127.0.0.1:8000/help-candidates-and-committees/guides/) or [Browse Data](http://127.0.0.1:8000/data/browse-data/?tab=raising)

### DIFF
--- a/fec/fec/static/scss/layout/_layout.scss
+++ b/fec/fec/static/scss/layout/_layout.scss
@@ -11,10 +11,14 @@
   padding: u(0 2rem);
   @include outer-container;
 
-  // There's a point early in $xl where we still want the side padding
-  // 20 px on each side
-  @media (min-width: calc(1088px + 2rem) ) {
+  @include media($lg) {
     padding: 0;
+  }
+}
+// Keep content off the side of the browser in these in-between widths as soon as we get into $lg
+@media screen and (min-width: 860px) and (max-width: 990px) {
+  .container {
+    max-width: calc(100vw - 4rem); // 2rem on each side
   }
 }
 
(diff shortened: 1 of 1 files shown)


patphongs (approved):
It works, I'm seeing the additional padding at the end as expected

PR #6715, opened 2025-03-19
4258 - Auto-size block-table table columns

### LINKED ISSUE #4258: Wagtail wishlist: Control column widths on table blocks
**What we're after:** We should be able to adjust column widths to better fit the content within them. 

**Considerations:**
- Easy for wagtail users to use
- Can column width adjust automatically?
- Pre-set widths based on the content? Fixed options that someone can choose from?
- Consider the new [Wagtail typed-table block](https://docs.wagtail.org/en/v2.15.4/reference/contrib/typed_table_block.html) with customization to allow users to choose  from 10%, 20$, 30% 0 etc width columns. @johnnyporkchops has some prototyped code saved in a local branch.
- See second part of[ this comment below](https://github.com/fecgov/fec-cms/issues/4258#issuecomment-998420749) for possible short term quick solution

**Completion criteria:**
- [ ] Intuitive formatting options implemented in table block for all applicable wagtail templates

Example table : https://www.fec.gov/legal-resources/enforcement/audit-reports/authorized-committee-audit-reports/kelly-for-congress-2016/
**Before:** 

**After:**
Screen Shot 2021-02-10 at 2 38 42 PM

rfultz:
## Summary

- Resolves #4258 

Letting block-table table columns autosize

### Required reviewers

- 1+ devs
- 1+ SMEs?
(I tagged way more than I needed)

Added `table-layout: auto;` to `.block-table table`.

It won't affect any table that's not in a `block-table` class element, but it will affect _every_ table inside one.

**SMEs**: Are there places we do _not_ want this to happen—anyplace where this will break things?

## Screenshots

Production above, localhost below
image

## Related PRs

None

## How to test

- Pull the branch
- `npm run build`
- `./manage.py runserver`
- Check [the example page used for this ticket](http://127.0.0.1:8000/legal-resources/enforcement/audit-reports/authorized-committee-audit-reports/kelly-for-congress-2016/)
- Check other locations `.block-table table` is being used
   - Are they generally okay?
   - Any of them broken?

### How to test in Inspector
- In the browser
  - right-click on something in the table ("Description" is good one)
  - choose Inspector.
- In the Elements tab,
  - click the `<table>` element
- Over (or down) in the Styles panel/tab,
  - click the plus sign
  - paste in `.block-table table`
  - tab
  - type/choose `table-layout`
  - tab
  - type/choose `auto`
  - tab (or enter or click outside)
- It should update in real time

### DIFF
--- a/fec/fec/static/scss/components/_articles.scss
+++ b/fec/fec/static/scss/components/_articles.scss
@@ -26,6 +26,7 @@
 }
 .block-table {
   table {
+    table-layout: auto;
     @extend .simple-table-cms;
   }
 }


JonellaCulmer (approved):
@rfultz This is amazing! I've had a deep dislike for this 50/50 split for some time. Approving based on the screenshots you provided. Thanks!

patphongs (approved):
Works well. Just a note we can apply this at some point to html tables in _richtext.scss. However that would require more testing. If they're using HTML, they probably want to do their own custom thing with the table. So maybe not necessary?

PR #6742, opened 2025-04-02
  [PR TO RELEASE] Sort AO by issue_date,  fix toggle on ao_no sort btn icon

johnnyporkchops:
## Summary (required)

- Resolves https://github.com/fecgov/openFEC/issues/6167
- Add sort by issue_date to AO data table
- Change sort column header styles to match other datatables

### Required reviewers

one frontend, one backend

## Impacted areas of the application

AO datatable `Case(ao_no)` and `Date(issue_date)`  columns

### Default when neither sort column has been clicked

Screenshot 2025-04-12 at 10 11 51 PM

### Sort-order  and  sort-type and are preserved when paginating

Screenshot 2025-04-12 at 10 19 57 PM

## Related PRs

https://github.com/fecgov/openFEC/pull/6173

## How to test

- `npm run build`
- `export FEC_API_URL=https://api-stage.open.fec.gov`
- Test AO datatable sort columns - Case(ao_no) and Date(issue_date)
- Confirm the sort-type and sort-order in the url (i.e: `sort=-ao_no` or `ao_no` and  `sort=-issue_date` or `issue_date`)
- Confirm that the sort icon is always toggles to the correct direction for the current sort order
- Test pagination. The sort-type and sort-order should persist when paginating. 
   - Note: Changing the sort always takes you back to page one is expected behavior

PR #6756, opened 2025-04-16
AO datatable consistent pagination

johnnyporkchops:
## Summary (required)
- After deploying [recent sort changes](https://github.com/fecgov/fec-cms/pull/6742) for AO datatables, we noticed that the top and bottom paginations  do not match when you change sort-order or sort-type
- Also, to match other datatables, fix the initial view on load  to  show the first column (ao-no) as being sorted descending with the icon pointing down. 

- Resolves #issue_number

### Required reviewers

one frontend, one backend

AO datatables sort and pagination functionality

## Screenshots

## Related PRs
https://github.com/fecgov/fec-cms/pull/6742

## How to test
- `npm run build`
- Confirm that the default, on first page load, shows the first column sorted descending 
- Confirm that switching sort-type (from one column to the other), it defaults to descending first.
- Test that the top and bottom pagination and meta data ("Showing 1-20 of etc") always match and that they both go back to page one when you change the sort-order or sort-type when you are paginating.

### DIFF
--- a/fec/fec/static/js/legal-search-ao.js
+++ b/fec/fec/static/js/legal-search-ao.js
@@ -30,7 +30,7 @@ export default function LegalSearchAo() {
   this.noResultsMessage;
   this.paginationElement;
   this.resultsTable;
-  this.sortOrder = 'desc';
+  this.sortOrder;
   this.sortType;
   this.tagList;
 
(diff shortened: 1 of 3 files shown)


fec-jli (approved):
It works as expected. Good Job, thanks

pkfec (approved):
Sort by ao_no and issue_date works well. Thanks for fixing broken pagination on bottom page. Now the top and bottom pagination links are consistent in showing the results. Thanks @johnnyporkchops

Issue #6747, opened 2025-04-08
AO datatable layout updates


JonellaCulmer:
### What we're after
Need to make design enhancements that improve access to final opinions and aligns the design with other legal datatables with similar functionality. 

### Completion criteria
- [x] Designs are shared and approved by relevant stakeholders and PMs
- [ ] Follow-up development tickets have been updated with designs.

### Considerations
- [x] Display the final opinion and other docs when applying the document type filter. Need to consider the appearance for the default view

JonellaCulmer:
@patphongs @AmyKort Here are some updated designs for AOs. It incorporates documents appearing in the front-end along with document highlights. In order to keep it as clean as possible, I re-arranged some of the columns, specifically I swapped the Summary and "Opinion cited by later opinions" column. I also retained the combined the AO number and name column rather than separating them. I don't think we have the real estate to separate them and try to accommodate the documents and highlights. Happy to discuss any of this. 


### No filters applied
![Image](https://github.com/user-attachments/assets/c9d7f607-ee29-4a52-b18f-4e49ead14756)

### Default view - final opinions document type selected
![Image](https://github.com/user-attachments/assets/d251cf2b-aafd-43ee-a062-200a8a0a6990)

### Final opinions and keyword search
![Image](https://github.com/user-attachments/assets/33aea97e-81d8-44e5-aca9-c25f33ab7b43)

### Multiple documents selected
![Image](https://github.com/user-attachments/assets/5770464d-bf85-485b-814c-4987c9424413)

### Multiple documents selected and keyword search
![Image](https://github.com/user-attachments/assets/babc5337-aa87-4b4b-92a1-25c6e4b7fb31)

patphongs:
@JonellaCulmer These new designs should work if we update the column widths so that the `Summary and documents` column is wider. Playing around with the widths, we could probably make the `case` column 15%, `date issued` column 10%, and `AO cited` column 10%. That leaves about 65% for the summary & documents. Could you try to approximate the new look with these adjusted column widths?

patphongs:
@JonellaCulmer @rfultz We talked about updating the design to have the AO results display as a [category feed](https://fec-pattern-library.app.cloud.gov/components/detail/multiple-category.html) instead of a table layout. This allows several advantages: aligning us with other text base driven searches, flexibility without the limitations of columns, responsive design with cards stack better than tables, and better scanning of results visually especially when highlighting search text. This should be responsive and implemented in all layouts.

[...]

PR #6822, opened 2025-07-19
AO Proximity search and new doc display

### LINKED ISSUE #6726: Proximity search in AOs front end implementation
### Summary

**What we're after:**
_We want to add the proximity search capability to the AO filter panel._

### Requirements
- Don't make any changes to the boolean search right now. Retain the way that the search keywords field in the filter panel and the specific boolean fields in the more keywords options modal is working now in production. @JonellaCulmer will make a separate ticket to update the boolean filters with the new design later.
- Determine appropriate validation for the proximity filters. We don't want the query event to fire with every change immediately. We discussed having a delay and/or validation on if both proximity keyword fields are filled out before possibly adding a message.
- Proximity filters will no longer have the ability to add additional primary fields. There was a feeling that most users would only want to define 2 fields anyways, so we're trying this approach first.
- Remove the secondary keyword proximity filter as we got the sense that this was also not needed.
- "More keyword options" has now changed to "Boolean filters"
- The filter tags will be a removal on the entirety of the proximity search fields, it will be all or nothing when clicking the `x` on the tags.

Image

Image

### Related issues

- #6670 

### Completion criteria

- [ ] Proximity search is added to the AO search

### LINKED ISSUE #6766: Enhance AO search results display
### Summary

**What we're after:**
Enhance the Advisory Opinion (AO) search results display by implementing document keyword highlighting. The highlighted terms should reflect the user’s search query and match the visuals shown in the approved design comp.

### Related issues

- Designs: https://github.com/fecgov/fec-cms/issues/6747#issuecomment-2790555126

### Action items:
- [ ] Display highlighted text snippet with accordion for additional matches
- [ ] Group documents by document type category
- [ ] See what column widths may work for the content. For example, try setting the case column 15%, date issued column 10%, and AO cited column 10%. That leaves about 65% for the summary & documents column.

### Completion criteria

- [ ] AO search results display is updated per design spec

### Future work

- Add AO year filter when the [API supports filtering of `ao_year`](https://github.com/fecgov/fec-cms/issues/6763)

johnnyporkchops:
## Summary (required)

- Resolves #6766
(also resolves #6726)

Proximity search for AOs
New document display design for AOs

New designs: https://github.com/fecgov/fec-cms/issues/6747#issuecomment-2790555126

### Required reviewers

One UX, One Content, One Developer

AO Datatable

## Screenshots

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

## How to test
- `npm run build`
- Go to AO datatable and test proximity-only search and proximity search combined with keyword search
- Also test document type filters
- Initial page load and pagination buttons use the Jinja template,
- Changing filters on page 1 uses JS to update the table without page load
- verify the new designs here https://github.com/fecgov/fec-cms/issues/6747#issuecomment-2790555126


fec/fec/static/js/modules/filters/keyword-proximity-filter.js
```diff
@@ -0,0 +1,384 @@
...
+  // TODO ?
+};
+
+/**
+ * Appends the warning message to the DOM (after keyword1) if it doesn't currently exist
+ */
+KeywordProximityFilter.prototype.showWarning = function() {
+  if (!this.showingWarning) {
+    const warning = '<div class="filter__message filter__message--error">' +
+      '<strong>(FPO)</strong> Keyword proximity search requires two terms and a distance</div>';
```
patphongs, line 372: Remove the `(FPO)` message here. I think the message looks fine.

fec/legal/views.py
```diff
@@ -208,6 +208,8 @@ def transform_ecfr_query_string(query_string):
 
 def legal_search(request):
     original_query = request.GET.get('search', '')
+    q_proximities = request.GET.get('q_proximity', '')
+    q_proximitys = request.GET.get('q_proximity', '')
```
patphongs, line 212: `q_proximitys` and `q_proximities` do the same thing: They both fetch the same q_proximity parameter and use `getlist()` to handle multiple values. Let's make it consistent and just use `q_proximities`.

Also, ensure that you update the jinja files too, to ensure that this `q_proximitys` variable is no longer referenced and uses `q_proximities` instead.

fec/legal/views.py
```diff
@@ -245,6 +247,8 @@ def legal_search(request):
     return render(request, 'legal-search-results.jinja', {
         'parent': 'legal',
         'query': original_query,
+        'q_proximities': q_proximities,
+        'q_proximitys': q_proximitys,
```
patphongs, line 251: Make it consistent and just use `q_proximities`.

fec/legal/views.py
johnnyporkchops: Thanks @pat, I planned to hold off on making q_proximitys/ies consistent until we do the PR to change the other legal types, documents display, so I don't have to incluse those templates in this PR at all. I have to call both for the legal landing for now.

JonellaCulmer:
@johnnyporkchops Here are my comments in pink. cc: @patphongs 

- The dividing line for the documents section should sit below other content across all columns. 
- The updated column header should be **Summary and documents**
- The document type column text should be in bold.
Screenshot 2025-07-25 at 11 24 51 AM


- Also pagination is not appearing below the content. So for search results with only one result there is no "bottom" to the table. 
Screenshot 2025-07-25 at 11 25 21 AM

JonellaCulmer (approved):
Thanks for those last few changes, @johnnyporkchops. Everything looks good to me.

isaacjamesbaker (approved):

PR #6840, opened 2025-07-31
6836-add connection pooling via Django/psycopg

### LINKED ISSUE #6836: Add Connection Pooling
### Summary

**What we're after:**
We need to add connection pooling to the CMS db connections-- probably via PGBouncer

### Related issues

_List any relevant related issue(s)_

- #[issue_number]
- #[issue_number]

### Completion criteria

_What does the end state look like? Include any behavior, appearance, etc required for this ticket to be 'complete.'_

_List any considerations the tech team should know. Additionally, any specific tech steps can be included here._

- [ ] First
- [ ] Next
- [ ] Then

### Future work

_Other future work that may be needed following this issue's completion._

cnlucas:
## Summary (required)

- Resolves #6836 

Adds connection pooling, [Django 5.1 added connection pooling using psycopg3's pool](https://docs.djangoproject.com/en/5.2/releases/5.1/#postgresql-connection-pools)
I set max_size at 50 which is 600 possible db connections
I lowered max_idle from 600 to 400 s

We should lower both eventually, but these are conservative estimates. Hoping we can ask cloud.gov for a month of PI db connections like we can do with the API. 

It seems that we regularly need <50 db connections, so for right now django's in built connection pooling should suffice. If this is not enough connections we can consider external pooling with pgbouncer later. 

Max CMS Application DB connections = the lowest of these two options :

     (num of workers) x (instances) x (limit of greenlets spawned) 

OR 

    max_connections in the DB (SHOW max_connections;)

LIMITS OF GREENLETS SPAWNED=

the lowest of these two options: 

    —worker-connections  (gunicorn gevent, default is 1000)

OR

    pool size (if enabled)

POOL SIZE:
```
DJANGO ORM
max_size
```

### Required reviewers

2-4 devs

-  DB connections

## How to test

- Push to dev/stage
- I tested by SSHing into stage ([wiki](https://github.com/fecgov/fec-cms/wiki/Switch-Wagtail-page%E2%80%90type-while-keeping-existing-ID-and-Slug)) lowering the max_size to 5 and increasing the instances to 2. I was able to get to 20 connections and could see the ten extra connections when I increased the instances. 
- You can do this with your local db if you connect via dbeaver
NOTE: when testing, you will also see your own connections and background/system processes.

### DIFF
--- a/fec/fec/settings/base.py
+++ b/fec/fec/settings/base.py
@@ -228,14 +228,18 @@ WSGI_APPLICATION = 'fec.wsgi.application'
 SECRET_KEY = os.environ.get('DJANGO_SECRET_KEY', get_random_string(50))
 
 
-# Database
-# https://docs.djangoproject.com/en/1.8/ref/settings/#databases
-
 DATABASES = {
-    # Be sure to set the DATABASE_URL environment variable on your local
-    # development machine so that the local database can be connected to.
     'default': dj_database_url.config()
 }
+
+# Add or update OPTIONS after the config() call
+DATABASES['default'].setdefault('OPTIONS', {})
+DATABASES['default']['OPTIONS'].update({
+    'pool': {
+        'max_size': 50,  # total possible db conn = Workers x Instances x max_size (600, probably too high)
+        'max_idle': 400,  # default is 600 sec
+    }
+})
 
 
 # Internationalization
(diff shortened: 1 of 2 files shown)


patphongs (approved):
@cnlucas tested and works well! I was able to make sure psycopg version 3.1.20 was installed, checked the connection pool size was max_size=50 and max_idle=400, and db connections working!

PR #7055, opened 2026-03-24
[HOTFIX - MERGE WITH GITFLOW] Add 5 second timeout to search.gov calls

patphongs:
## Summary

Add a 5 second timeout for search.gov API calls so the page can still render other search results.

### Required reviewers

1 engineer

- Global website search
- Guidance search

## How to test

- Use the global legal search and enter any search string. After 5 seconds, the results page should render and if there's an issue with the "Other pages" that we use search.gov API for, it will render the "Something went wrong" error block. Ex: http://localhost:8000/search/?type=candidates&type=committees&type=site&query=contributions
- Use the guidance search and enter any search string. If the results do not come back in 5 seconds, it will render the "Something went wrong" error block. Ex: https://www.fec.gov/legal-resources/policy-and-other-guidance/guidance-documents/?query=rule
- The server error logs should also have this error message every time there's a timeout: "ERROR:search.views: search_site: Search.gov request timed out after 5 seconds for query: contributions" or  "ERROR:search.views: policy_guidance_search_site: Search.gov request timed out after 5 seconds for query: rule"

### DIFF
--- a/fec/search/views.py
+++ b/fec/search/views.py
@@ -127,11 +127,13 @@ def search_site(query, limit=0, offset=0):
         'offset': offset
     }
     try:
-        r = requests.get('https://api.gsa.gov/technology/searchgov/v2/results/i14y', params=params)
+        r = requests.get('https://api.gsa.gov/technology/searchgov/v2/results/i14y', params=params, timeout=5)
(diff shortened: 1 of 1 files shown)


rfultz (approved):
Works for me!

PR #7057, opened 2026-03-25
[PR TO RELEASE 04/07/26 ] Reduce redundancy in rulemaking single page document listings

johnnyporkchops:
## Summary (required)

- Resolves #7041 

-  **compare library filter** : Compares two strings to determine if they are nearly identical or semantically similar using Python's built in  [difflib.SequenceMatcher](https://docs.python.org/3/library/difflib.html):   
    - This seems to work well for our document descriptions/labels with a `similarity_ratio of threshold of .5`.  Meaning if the description and the label are  >= 50% similar, we don't show it.
- **compare_commenter filter** :  compare the `commenter name` to `doc_description`  when one is  "lastname, firstname" and the other is  "firstname lastname."
- Avoid repeating `commenter.name | commenter.role` in the entity listing for single commenter comments where it is already shown  as the pdf link text above.

### Required reviewers

one or two devs

Rulemaking single pages
new filters in filters.py

### BEFORE:
2022-06-BEFORE
<hr>
### AFTER:
2022-06-AFTER

## Related PRs

https://github.com/fecgov/fec-cms/pull/7054

## How to test
 - Test pages confirm document description, labels and commenter names are not repeated verbatim or redundantly 
 - On dev.fec.gov, you can see the item removed by filters in red and items retained by filter in green
 - Examples 2022-06, 2024-01, 2023-02

### DIFF
--- a/fec/data/templatetags/filters.py
+++ b/fec/data/templatetags/filters.py
@@ -185,6 +185,21 @@ def compare(string, string_1):
         return True
 
 
+@library.filter
+def compare_commenter(string, string_1):
+    """Convert `string_1` from "LastName, FirstName" format to "FirstName LastName" 
+    Returns False if `converted_string_1` is in `string`. Else returns True.
+    """
+
+    # Convert name from "LastName, FirstName" format to "FirstName LastName"
+    converted_string_1 = ' '.join(reversed(string_1.split(', ')))
+
+    if converted_string_1 in string:
+        return False
+    else:
+        return True
+
+
 @library.global_function
 def path_for_css(key):
     """Looks up the hashed asset key in rev-manifest-css.json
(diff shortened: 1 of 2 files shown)


patphongs (approved):
@johnnyporkchops This looks good, I couldn't find any instances that this did not cover or was incorrect

patphongs:
@johnnyporkchops One thing I noticed is that the entire `doc_entities` list is not displayed under the comment for your branch. For this Rulemaking, I should see all 3 in the entities list listed under the comment document, but I only see 2 listed locally. https://stage.fec.gov/legal/rulemakings/2022-06/

Screenshot 2026-04-06 at 11 31 19 AM

johnnyporkchops:
@patphongs, 
For single commenters only, I removed the commenter-role from the entity list because it is redundant with the commenter name already shown above. 
`{% for entity in doc.doc_entities if entity.role != 'Commenter'%}`

patphongs (approved):
Looks good @johnnyporkchops ! I'm documenting here that we will show redundancy for single commenters as we want to show the entire entities list.

PR #7081, opened 2026-05-05
[PR TO 05/05/26 RELEASE] Restore legal pagination for datatables

johnnyporkchops:
## Summary 

- Resolves **"Type error"** when loading rulemakings datatable page

Screenshot 2026-05-05 at 2 01 37 PM

Restores legal pagination for Rulemakings in jq Datatables

### Required reviewers

one or two devs

-  Rulemakings datatable page

## Screenshots

Rulemakings datatable page loads OK.  "Type error" no longer shows on browser console.

After: 

Screenshot 2026-05-05 at 2 04 04 PM

## How to test

- `npm run build`
- check http://127.0.0.1:8000/legal/search/rulemakings/ to make sure it works and you do not get a console error
- check new datatables details panels for any issues

### DIFF
--- a/fec/fec/static/js/modules/tables.js
+++ b/fec/fec/static/js/modules/tables.js
@@ -152,12 +152,15 @@ let pagination_legal;
  */
 let legal_type;
 export function getCount(response) {
-  let pagination_count = response.pagination.count; // eslint-disable-line camelcase
+  let pagination_count;
   if (!response.pagination) {
      pagination_legal = true;
      legal_type = Object.keys(response)[0];
      const legal_type_count = `total_${legal_type}`;
      pagination_count = response[legal_type_count];
+  }
+  else {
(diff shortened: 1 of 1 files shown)


fec/fec/static/js/modules/tables.js
```diff
@@ -152,13 +152,16 @@ let pagination_legal;
  */
 let legal_type;
 export function getCount(response) {
-  let pagination_count = response.pagination.count; // eslint-disable-line camelcase
+  let pagination_count;
```
patphongs: @johnnyporkchops I think this is only a partial fix because `response.pagination.count` is still being used in the rounding logic further down. I just made another PR to try to fix it to where pagination count is being used consistently instead of touching response.pagination count directly. What do you think? https://github.com/fecgov/fec-cms/pull/7084

PR #6535, opened 2024-10-31
Upgrade django v4.2.16

tmpayton:
## Summary (required)

- Resolves #6506 and https://github.com/fecgov/fec-cms/issues/6489

This PR upgrades django to remediate a security vulnerability. 

## Impacted areas of the application

General components of the application that this PR will affect:

-  django

## How to test

- `git checkout feature/6506-upgrade-django`
- activate your virtual env
-  `pip install -r requirements.txt`
- `snyk test --file=requirements.txt --package-manager=pip` - make sure that django does not appear in list
- `./manage.py runserver`

PR #7144, opened 2026-06-18
Handle special characters in eCFR legal search queries

### LINKED ISSUE #6339: Certain characters (%, #) in legal search breaks the eCFR api call
### Summary

_What happened?_

Searching keywords with certain symbols in legal search for Regulations breaks the eCFR api call. 

-  Searching "95%" gets`Internal Server Error`: https://stage.fec.gov/data/legal/search/?search_type=all&search=95%25
-  Searching "#6535" or "MUR #6535", gets weird results for regulations: https://stage.fec.gov/data/legal/search/?search_type=regulations&search=%236535.
- Do we need to sanitize the characters we are passing to the eCFR?

### Frequency

_an ongoing issue,_

### How to Reproduce

Go to the global legal search and search for "95%", "#6535" or "MUR #6535" in `All` or `Regulations`. 

### Searched 95%:
Screenshot 2024-06-16 at 6 35 56 PM

<hr>

### Searched #6535:
Screenshot 2024-06-16 at 6 36 46 PM

### Misc

_If you'd like to include anything else that may be of interest, do that here._

arpitjain099:
Fixes #6339

Queries like 95% and #6535 were being passed into the eCFR URL as raw text, so special characters could break parsing and trigger a server error in legal search.

This change switches fetch_ecfr_data to requests.get with params, so query terms are encoded safely. It also adds a fallback empty payload when eCFR returns a non-200 response, so legal search pages render safely instead of crashing.

I also added focused tests in fec/data/tests/test_ecfr_caller.py for:
- special-character query params
- non-200 fallback payload

Verification:
- python3 -m pytest fec/data/tests/test_ecfr_caller.py
- python3 -m pytest fec/data/tests/test_legal_search.py -k transform_ecfr_query_string (fails in this environment: ModuleNotFoundError: No module named django)

### DIFF
--- a/fec/data/ecfr_caller.py
+++ b/fec/data/ecfr_caller.py
@@ -6,25 +6,50 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
+def _empty_response(page):
+    try:
+        current_page = int(page)
+    except (TypeError, ValueError):
+        current_page = 1
+
+    return {
+        "results": [],
+        "meta": {
+            "current_page": current_page,
+            "total_pages": 0,
+            "total_count": 0,
+        },
+    }
+
+
 def fetch_ecfr_data(query, date="current", limit=20, page=1):
     order = "hierarchy"
     if query:
         order = "relevance"
-    url = (f"https://www.ecfr.gov/api/search/v1/results?query={query}&"
-           f"agency_slugs%5B%5D=federal-election-commission&"
-           f"date={date}&"
-           f"per_page={limit}&"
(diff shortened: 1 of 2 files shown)


patphongs:
@arpitjain099 I reviewed the PR, and the core approach looks good. Passing the eCFR search parameters through `requests.get(..., params=...)` is the correct way to handle queries with characters like `%` and `#`. We want to make a few broader changes around other special cases and error handling and will make a separate PR for this that incorporates your ideas. Thank you for your interests in our opensource repo and suggestions to improve our CMS.

PR #7154, opened 2026-06-25
Rulemakings - show first doc if no key doc and restore active-doc indicator  

### LINKED ISSUE #7112: Show most recent document when no key documents are available, chosen-document indicator on single page missing
**What we're after:**
- Show most recent rulemaking document on rulemaking search page if no documents are marked as "key documents". Documents should appear dynamically based on their type, substantive documents to be specific, and documents that should not be shown include submitted comments from the public. 

Example rulemakings and the documents that should appear: (this is not a complete list)
2024-06: https://www.fec.gov/legal/rulemakings/2024-06/?doc_id=425657
2024-03: https://www.fec.gov/legal/rulemakings/2024-03/?doc_id=425082
2024-02: https://www.fec.gov/legal/rulemakings/2024-02/?doc_id=425158

- On rulemaking single page, the chosen-document blue bar seems to be missing

### Completion criteria
- [ ] All rulemakings by default on rulemaking search page have one document listed under the rulemaking name
- [ ]  On rulemaking single page, the chosen-document blue bar is restored

Example of rulemaking missing a document:
Image

johnnyporkchops:
## Summary (required)

- Resolves #7112

- Default to show the first level-1 document if there is no key document on Rulemakings search page
- Restore active-document indicator on Rulemaking single page
- Typo introduced in last changes to `rulemaking.jinja` causing server errors on dev (2024-08, 2024-01, 2023-02 and others)

### Required reviewers
one or two devs

Rulemakings search and Rulemakings single pages

### Restored active-doc  indicator:
Screenshot 2026-06-24 at 8 46 54 PM

### Server error caused by typo in last develop changes to `rulemaking.jinja`:
Screenshot 2026-06-24 at 8 43 05 PM

## How to test
- `npm run build`
- Confirm that RMs without key docs show the first level-1 doc instead (2024-06 is an example)
- Test the active-document indicator on rulemaking single page (see screenshot above)
- Confirm server error does not happen on single page for 2024-08, 2024-01, 2023-02 and others.

### DIFF
--- a/fec/fec/static/js/modules/columns.js
+++ b/fec/fec/static/js/modules/columns.js
@@ -1116,31 +1116,37 @@ export const rulemakings = [
     render: function (data, type, row) {
       // Array to keep track of documents already shown
       let doc_ids = [];
-
       let html = `<p><strong>${row.rm_name}</strong>`;
+      let first_doc;
       if (row.key_documents && row.key_documents.length ) {
+        first_doc = row.key_documents[0];
+      }
+      else {
+        first_doc = row.documents[0];
+      }
+      const firstDocUrl = first_doc.url ? first_doc.url.replace(/#/g, '%23') : '';
+
         html += `<br><span class="icon icon--inline--left i-document"></span>`;
-        const keyDocUrl = row.key_documents[0].url ? row.key_documents[0].url.replace(/#/g, '%23') : '';
         html +=
           buildEntityLink(
-            row.key_documents[0].doc_type_label, keyDocUrl, row.key_documents[0].doc_type_label);
-            if (row.key_documents[0].doc_date !== null) {
-                const doc_date = moment(row.key_documents[0].doc_date).format('MM/DD/YYYY');
+            first_doc.doc_type_label, firstDocUrl, first_doc.doc_type_label);
+            if (first_doc.doc_date !== null) {
+                const doc_date = moment(first_doc.doc_date).format('MM/DD/YYYY');
                 html += ` | ${doc_date}`;
             }
             else {
               html += ' | Undated';
             }
-      }
+
         html += `</p>`;
 
-    const filters = new URLSearchParams(window.location.search);
-    const filters_category_type = filters.has('doc_category_id');
-    const filters_keyword = filters.has('q'); //getAll('q').length;
-    const filters_proximity = filters.has('q_proximity') && filters.getAll('q_proximity').length == 2;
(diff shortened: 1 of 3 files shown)


fec/fec/static/js/modules/columns.js
```diff
@@ -1116,31 +1116,37 @@ export const rulemakings = [
...
       let html = `<p><strong>${row.rm_name}</strong>`;
+      let first_doc;
       if (row.key_documents && row.key_documents.length ) {
+        first_doc = row.key_documents[0];
+      }
+      else {
+        first_doc = row.documents[0];
+      }
+      const firstDocUrl = first_doc.url ? first_doc.url.replace(/#/g, '%23') : '';
+
```
patphongs: `first_doc = row.documents[0]` is being used here without checking that `row.documents` exists or has entries first. This should be checked first. Then line 1127 refers to `first_doc.url`. If the API returns a rulemaking with no key docs and/or no top-level documents, this will throw an error and break the results table.

fec/fec/static/scss/components/_table-styles.scss
```diff
@@ -26,6 +26,10 @@
   border-width: 2px 0;
   font-family: $sans-serif;
   margin: u(2rem 0);
+
+  tr.row-active {
+    border-left: 0.5rem solid $primary;
+  }
```
patphongs, line 32: I'm noticing that the new `border-left` applies only to `tr.row-active`. This is causing the selected row’s content to shift slightly when clicking to other documents. To fix this, maybe consider giving all `tr` in the document table a transparent left border of the same size so it doesn't shift?

patphongs (changes requested):
@johnnyporkchops A few suggested changes to make this a bit smoother

fec/fec/static/scss/components/_table-styles.scss
patphongs: Can you make this `0.5rem`? The 3px is thinner than what it was before

patphongs (changes requested):
@johnnyporkchops Last minor change:

fec/fec/static/scss/components/_table-styles.scss
johnnyporkchops: Oh, yeah, my mistake

patphongs (approved):
Good to go, thanks @johnnyporkchops

PR #7217, opened 2026-08-19
add legal key

cnlucas:
## Summary (required)

Adds seperate legal key 

### Required reviewers

1-2 devs

-  

## Screenshots

## Related PRs

branch | PR
------ | ------
fix/other_pr | [link]()
feature/other_pr | [link]()

## How to test

- deploy to stage https://app.circleci.com/pipelines/github/fecgov/fec-cms/5045/workflows/83c0372f-144c-43bf-9343-72f558bb461b/jobs/8663
- Currently it's using one of my test keys, but you can see in the API umbrella legal is using my key instead of the regular private key.


fec/data/api_caller.py
```diff
@@ -41,9 +41,14 @@
 session.mount("https://", http_adapter)
 
 
+def _call_legal_api(*path_parts, **filters):
+    filters["api_key"] = settings.FEC_LEGAL_API_KEY_PRIVATE
+    return _call_api(*path_parts, **filters)
```
patphongs, line 47: 2) If `FEC_WEB_LEGAL_API_KEY_PRIVATE` is missing in any environment, this sets `api_key=None` and prevents `_call_api()` from falling back to `FEC_API_KEY_PRIVATE`. Can you use `settings.FEC_LEGAL_API_KEY_PRIVATE or settings.FEC_API_KEY_PRIVATE`? This will be a nice filter so that we can change if and when we need to.

patphongs (changes requested):
@cnlucas Great work on this, I like having the optionality. 

I found 2 items that need to be addressed.

1) This path needs updating: [`fec/legal/views.py:842`](https://github.com/fecgov/fec-cms/blob/5437cfb023dae9e6df14ef553baf5c0e24b57c34/fec/legal/views.py#L842) still calls `_call_api()` for universal search rulemaking results, so `/legal/search/` with `all` or `rulemakings` will keep using the regular private key. That should route through `_call_legal_api()` too.

patphongs (approved):
@cnlucas Looks good now! It's nice to be able to implement this as it will fallback to the private key if no legal private key is there.

